PR #138, opened 2014-12-12
Support type annotations.

### LINKED ISSUE #136: Type annotations
This would be really easy if `TypeName` was an `abstract class`...

Examples:
```java
@NonNull List<String>
List<@NonNull String>
new @Thing Foo();
new Foo.@Thing Bar();
foo.new @Thing Bar();
<? super @NonNull String>
<T extends @Thing Foo>
```

JakeWharton:
Closes #136.


swankjesse:
Taco builders is happening. I'm closing this for now.

PR #151, opened 2015-01-06
First, rough draft of immutable builders.

swankjesse:
This is currently in its own 'builders' subpackage. I'd like to get
more of this working, and then move it into the main 'javawriter'
package, and remove the code that's currently in that package.

### DIFF
--- a/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
+++ b/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javawriter.builders;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSortedMap;
+import com.squareup.javawriter.ClassName;
+import com.squareup.javawriter.StringLiteral;
+import com.squareup.javawriter.TypeName;
+import com.squareup.javawriter.TypeNames;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Converts a {@link JavaFile} to a string suitable to both human- and javac-consumption. This
+ * honors imports, indentation, and deferred variable names.
+ */
+class CodeWriter {
+  private final String indent = "  ";
+  private final StringBuilder out;
+  private final ImmutableMap<ClassName, String> importedTypes;
+  private final LinkedHashSet<TypeName> emittedTypes = new LinkedHashSet<>();
+  private final List<TypeName> visibleTypes = new ArrayList<>();
+  private int indentLevel;
+
+  public CodeWriter(StringBuilder out, ImmutableMap<ClassName, String> importedTypes) {
+    this.out = checkNotNull(out);
+    this.importedTypes = checkNotNull(importedTypes);
+  }
+
+  public CodeWriter indent() {
+    indentLevel++;
+    return this;
+  }
+
+  public CodeWriter unindent() {
+    checkState(indentLevel > 0);
+    indentLevel--;
+    return this;
+  }
+
+  public CodeWriter pushVisibleType(TypeName typeName) {
+    visibleTypes.add(typeName);
+    return this;
+  }
+
+  public CodeWriter popVisibleType(TypeName typeName) {
+    checkState(visibleTypes.remove(typeName));
+    return this;
+  }
+
+  public CodeWriter emit(String format, Object... args) {
+    return emit(new Snippet(format, args));
+  }
+
+  public CodeWriter emit(Snippet snippet) {
+    int a = 0;
+    for (String part : snippet.formatParts) {
+      switch (part) {
+        case "$L":
+          emitAndIndent(String.valueOf(snippet.args.get(a++)));
+          break;
+
+        case "$S":
+          String arg = String.valueOf(snippet.args.get(a++));
+          emitAndIndent(StringLiteral.forValue(arg).literal());
+          break;
+
+        case "$T":
+          emitType(snippet.args.get(a++));
+          break;
+
+        case "$$":
+          emitAndIndent("$");
+          break;
+
+        default:
+          emitAndIndent(part);
+          break;
+      }
+    }
+    return this;
+  }
+
+  private void emitType(Object arg) {
+    TypeName typeName = toTypeName(arg);
+    emittedTypes.add(typeName);
+
+    String shortName = !visibleTypes.contains(typeName)
+        ? importedTypes.get(typeName)
+        : null;
+
+    emitAndIndent(shortName != null ? shortName : typeName.toString());
+  }
+
+  /** Emits {@code s} with indentation as required. */
+  private void emitAndIndent(String s) {
+    boolean first = true;
+    for (String line : s.split("\n", -1)) {
+      if (!first) out.append('\n');
+      first = false;
+      if (line.isEmpty()) continue; // Don't indent empty lines.
+      emitIndentationIfNecessary();
+      out.append(line);
+    }
+  }
+
+  private void emitIndentationIfNecessary() {
+    // Only emit indentation immediately after a '\n' character.
+    if (out.length() <= 0 || out.charAt(out.length() - 1) != '\n') return;
+
+    for (int j = 0; j < indentLevel; j++) {
+      out.append(indent);
+    }
+  }
+
+  private TypeName toTypeName(Object arg) {
+    if (arg instanceof TypeName) return (TypeName) arg;
+    if (arg instanceof Class<?>) return TypeNames.forClass((Class<?>) arg);
+    throw new IllegalArgumentException("Expected type but was " + arg);
+  }
+
+  /**
+   * Returns the types that should have been imported for this code. If there were any simple name
+   * collisions, that type's first use is imported.
+   */
+  ImmutableMap<ClassName, String> suggestedImports() {
+    // Find the simple names that can be imported, and the classes that they target.
+    Map<String, ClassName> simpleNameToType = new LinkedHashMap<>();
+    for (TypeName typeName : emittedTypes) {
+      if (!(typeName instanceof ClassName)) continue;
+      ClassName className = (ClassName) typeName;
+      if (simpleNameToType.containsKey(className.simpleName())) continue;
+      simpleNameToType.put(className.simpleName(), className);
+    }
+
+    // Invert the map.
+    ImmutableSortedMap.Builder<ClassName, String> typeToSimpleName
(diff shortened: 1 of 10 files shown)


swankjesse:
@gk5885 @JakeWharton @AlecStrong

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -0,0 +1,169 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Converts a {@link JavaFile} to a string suitable to both human- and javac-consumption. This
+ * honors imports, indentation, and deferred variable names.
+ */
+class CodeWriter {
+  private final String indent = "  ";
+  private final StringBuilder out;
```
gk5885: The original choice to code to `Appendable` and `CharSequence` was quite intentional.  Ridiculously enough, we have builds that will actually feel some pain from not being able to write directly to the file objects and having to turn `Name`s into `String`s.

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
```diff
@@ -0,0 +1,120 @@
...
+import com.squareup.javawriter.VoidName;
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.List;
+import javax.lang.model.element.Modifier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/** A generated method declaration. */
+public final class MethodSpec {
```
gk5885: FWIW, not having `toString` on these will almost certainly cause some pain.

src/main/java/com/squareup/javawriter/builders/Snippet.java
```diff
@@ -0,0 +1,72 @@
...
+ *   <li>{@code $L} emits the value with no escaping.
+ *   <li>{@code $S} escape the value as a string, wrap it with double quotes, and emit that.
+ *   <li>{@code $T} emits a type reference. Types will be imported if possible.
+ *   <li>{@code $$} emits a dollar sign.
+ * </ul>
+ */
+final class Snippet {
+  /** A heterogeneous list containing string literals and value placeholders. */
+  final ImmutableList<String> formatParts;
+  final Object[] args;
```
gk5885: Non-private Object array?  What is this?  Android?  ;)

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
swankjesse: Agreed. I think I'll switch 'em all to use AutoValue (in follow-up)

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: What kind of pain?

swankjesse: Ie. you're generating Java source files that don't fit into memory?

src/main/java/com/squareup/javawriter/builders/Snippet.java
swankjesse: Ooh, that's sloppy. Will fix.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -0,0 +1,80 @@
...
+      if (!firstMethod) codeWriter.emit("\n");
+      methodSpec.emit(codeWriter);
+      firstMethod = false;
+    }
+
+    codeWriter.unindent();
+    codeWriter.emit("}\n");
+  }
+
+  public static enum Type {
```
gk5885: Generally -1 on this strategy.  Interfaces can't have constructors, enums are the only ones with values, etc.  A little bit of extra code can make a lot of that code a little bit safer and _nobody_ will ever checkState() before adding a constructor.

JakeWharton: While I agree and love the type safety afforded by the current JW3 code, we also have javac at our backs. We will never be able to prevent 100% of invalid code in the type system so it becomes a balance of how much we put into the API and how much we just defer to javac.

Ultimately I think we should absolutely explore subtypes of this class to map to the different Java types.

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
```diff
@@ -0,0 +1,120 @@
...
+      return this;
+    }
+
+    public Builder addParameter(Class<?> type, String name) {
+      this.parameters.add(new ParameterSpec.Builder().type(type).name(name).build());
+      return this;
+    }
+
+    public Builder modifiers(Modifier... modifiers) {
+      this.modifiers = ImmutableSet.copyOf(modifiers);
```
JakeWharton: This should be additive (via ImmutableSet.Builder). We have instances where the need for a class to be public is determined separately from initial creation. I don't want to have to track the full modifier set myself.

src/main/java/com/squareup/javawriter/builders/Snippet.java
```diff
@@ -0,0 +1,72 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * A deferred format string. Unlike {@link java.text.Format} which uses percent {@code %} to escape
+ * placeholders, this uses {@code $}, and has its own set of permitted placeholders:
+ *
+ * <ul>
+ *   <li>{@code $L} emits the value with no escaping.
+ *   <li>{@code $S} escape the value as a string, wrap it with double quotes, and emit that.
```
JakeWharton: escape --> escapes

src/main/java/com/squareup/javawriter/builders/Snippet.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * A deferred format string. Unlike {@link java.text.Format} which uses percent {@code %} to escape
+ * placeholders, this uses {@code $}, and has its own set of permitted placeholders:
+ *
+ * <ul>
+ *   <li>{@code $L} emits the value with no escaping.
```
JakeWharton: Perhaps say "literal value" an wrap "literal" with `<em>` to try and force the association with the letter. Same for "string" and "type" for the next two.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -0,0 +1,80 @@
...
+package com.squareup.javawriter.builders;
+
+import com.google.common.collect.ImmutableList;
+import com.squareup.javawriter.ClassName;
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/** A generated class, interface or enum declaration. */
```
JakeWharton: nit: oxford comma

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
swankjesse: Fixed and renamed to `addModifiers`.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -0,0 +1,80 @@
...
+
+    codeWriter.unindent();
+    codeWriter.emit("}\n");
+  }
+
+  public static enum Type {
+    CLASS, INTERFACE, ENUM
+  }
+
+  public static class Builder {
```
JakeWharton: nit: final

src/main/java/com/squareup/javawriter/builders/ParameterSpec.java
```diff
@@ -0,0 +1,61 @@
...
+public final class ParameterSpec {
+  public final TypeName type;
+  public final Name name;
+
+  private ParameterSpec(Builder builder) {
+    this.type = checkNotNull(builder.type);
+    this.name = checkNotNull(builder.name);
+  }
+
+  public static class Builder {
```
JakeWharton: nit: final

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
```diff
@@ -0,0 +1,120 @@
...
+    codeWriter.indent();
+    for (Snippet snippet : snippets) {
+      codeWriter.emit(snippet);
+    }
+    codeWriter.unindent();
+
+    codeWriter.emit("}\n");
+  }
+
+  public static class Builder {
```
JakeWharton: nit: final

src/main/java/com/squareup/javawriter/builders/JavaFile.java
```diff
@@ -0,0 +1,70 @@
...
+    codeWriter.emit("package $L;\n", typeSpec.name.packageName());
+    codeWriter.emit("\n");
+    for (ClassName className : imports.keySet()) {
+      codeWriter.emit("import $L;\n", className);
+    }
+    codeWriter.emit("\n");
+    typeSpec.emit(codeWriter);
+  }
+
+  public static class Builder {
```
JakeWharton: nit: fianl

src/main/java/com/squareup/javawriter/builders/FieldSpec.java
```diff
@@ -0,0 +1,58 @@
...
+  public final Name name;
+  public final Snippet initializer;
+
+  private FieldSpec(Builder builder) {
+    this.type = checkNotNull(builder.type);
+    this.name = checkNotNull(builder.name);
+    this.initializer = checkNotNull(builder.initializer);
+  }
+
+  public static class Builder {
```
JakeWharton: nit: fianl

src/main/java/com/squareup/javawriter/builders/Snippet.java
swankjesse: Fixed.

src/main/java/com/squareup/javawriter/builders/Snippet.java
swankjesse: Good idea. Done.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
swankjesse: Fixed.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
swankjesse: Fixed for all.

src/main/java/com/squareup/javawriter/builders/FieldSpec.java
swankjesse: fxdei

JakeWharton:
Here's a few of the hard cases we need to emit in some of our libraries:
```java
  static final Thing.Thang<Foo, Bar> NAME =
      new Thing.Thang<Foo, Bar>() {
        @Override public Thung<? super Bar> call(
            final Thung<? super Foo> thung) {
          return new SimpleThung<Bar>(thung) {
            @Override public void doSomething(Bar bar) {
              /* code snippets */
            }
          };
        }
      };
```
```java
  public Foo(long id, @Ping String one, @Ping String two, @Pong String three,
      @Ping String four) {
    /* code snippets */
  }
```
```java
@JsonAdapter(Foo.class)
private final String thing;
```
```java
@Something(hi = SomeType.FIELD, hey = 12, hello = "goodbye")
public class Foo {}
```

JakeWharton:
Here's a contrived Retrofit service declaration which I'd also like to generate:
```java
interface Service {
  @Headers({
    "Accept: application/json",
    "User-Agent: foobar"
  })
  @POST("/foo/bar")
  Observable<FooBar> fooBar(
    @Body Things<Thing> things,
    @QueryMap(encodeValues = false) Map<String, String> query,
    @Header("Authorization") String authorization
  );
}
```

swankjesse:
Excellent. I'll write tests to generate each of those.

swankjesse:
Rather than going to `master` straight away, I'm going to work in a `builders` branch.
https://github.com/square/javawriter/pull/152

PR #153, opened 2015-01-07
Varargs support for parameters

### LINKED ISSUE #148: JavaWriter 3: Varargs for methods
It seems that JW 3 doesn't support `...`, is this planned?

AlecStrong:
Closes #148

### DIFF
--- a/src/main/java/com/squareup/javawriter/ParameterWriter.java
+++ b/src/main/java/com/squareup/javawriter/ParameterWriter.java
@@ -18,8 +18,14 @@ package com.squareup.javawriter;
 import java.io.IOException;
 
 public final class ParameterWriter extends VariableWriter {
+  private boolean varargs;
+
   ParameterWriter(TypeName type, String name) {
     super(type, name);
+  }
+
+  public void setVarargs(boolean varargs) {
+    this.varargs = varargs;
   }
 
   @Override
@@ -27,6 +33,9 @@ public final class ParameterWriter extends VariableWriter {
     writeAnnotations(appendable, context, ' ');
     writeModifiers(appendable);
     type().write(appendable, context);
+    if (varargs) {
+      appendable.append("...");
+    }
     return appendable.append(' ').append(name());
   }
 }
--- a/src/test/java/com/squareup/javawriter/ParameterWriterTest.java
+++ b/src/test/java/com/squareup/javawriter/ParameterWriterTest.java
@@ -22,7 +22,7 @@ import org.junit.runners.JUnit4;
 import static com.google.common.truth.Truth.assertThat;
 
 /**
- * ParameterWriter only differs from FieldWriter in the format of its annotations.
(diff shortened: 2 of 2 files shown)


src/main/java/com/squareup/javawriter/ParameterWriter.java
```diff
@@ -18,15 +18,24 @@
 import java.io.IOException;
 
 public final class ParameterWriter extends VariableWriter {
+  private boolean isVararg;
+
   ParameterWriter(TypeName type, String name) {
     super(type, name);
   }
 
+  public void setIsVararg(boolean isVararg) {
```
JakeWharton: I think this is just `.varargs()` and sets the value to `true`.

JakeWharton:
LGTM

src/main/java/com/squareup/javawriter/ParameterWriter.java
JakeWharton: or maybe `setVarargs()`?

JakeWharton: Odd for a setter to not take a value, but I find the `boolean` to be useless as well.

JakeWharton: Ugh. How about `setVarargs(boolean varargs)`?

AlecStrong: yea I spent a good five minutes trying to figure out a better nomenclature. I opted not to use setVarargs because it sounded like I was adding varargs to something.

AlecStrong: I could also move the API into MethodWriter but that would mean adding four new functions 
```
methodWriter.addVararg(Class<?>/TypeName/etc...)
```

JakeWharton: In the immutable-builder world this could just be `asVarargs()` on the type (not builder) and return a new instance where it was true...

JakeWharton: I think I still like `setVarargs` best. In normal code you'd never actually set varargs on anything (e.g., reflective method invocation) because it's always just an array. Your setter would be semantic in name (e.g., setColumns). Varargs only exist in code, and we're an API for writing code. Does it still bug you after that?

AlecStrong: Works for me!

swankjesse:
Uh oh, this collides with my intention to kill nearly every class with the `*Writer` suffix. Though I can certainly bring this along.

JakeWharton:
Yeah. Just unblocking ourselves on current JW3 usage. Keep killing.

kenzierocks:
Please do! I'm trying to mass generate lots of classes with varargs.

PR #154, opened 2015-01-08
Support anonymous inner classes in builders.

swankjesse:
Also improve modifiers, fields and annotations.

### DIFF
--- a/src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javawriter.builders;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedMap;
+import com.google.common.collect.Maps;
+import com.squareup.javawriter.ClassName;
+import java.lang.annotation.Annotation;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.SortedMap;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Iterables.getOnlyElement;
+
+/** A generated annotation on a declaration. */
+public final class AnnotationSpec {
+  public final ClassName type;
+  public final ImmutableSortedMap<String, Snippet> members;
+
+  private AnnotationSpec(Builder builder) {
+    this.type = checkNotNull(builder.type);
+    this.members = ImmutableSortedMap.copyOf(builder.members);
+  }
+
+  void emit(CodeWriter codeWriter, boolean inline) {
+    String separator = inline ? "" : "\n";
+    String suffix = inline ? " " : "\n";
+    if (members.isEmpty()) {
+      // @Singleton
+      codeWriter.emit("@$T$L", type, suffix);
+    } else if (members.keySet().equals(ImmutableSet.of("value"))) {
+      // @Named("foo")
+      codeWriter.emit("@$T(");
+      codeWriter.emit(getOnlyElement(members.values()));
+      codeWriter.emit(")$L", suffix);
+    } else {
+      // Inline:
+      //   @Column(name = "updated_at", nullable = false)
+      //
+      // Not inline:
+      //   @Column(
+      //       name = "updated_at",
+      //       nullable = false
+      //   )
+      codeWriter.emit("@$T($L", type, separator);
+      codeWriter.indent();
+      codeWriter.indent();
+      for (Iterator<Map.Entry<String, Snippet>> i = members.entrySet().iterator(); i.hasNext();) {
+        Map.Entry<String, Snippet> entry = i.next();
+        codeWriter.emit("$L = ", entry.getKey());
+        codeWriter.emit(entry.getValue());
+        if (i.hasNext()) codeWriter.emit(",");
+        codeWriter.emit("$L", separator);
+      }
+      codeWriter.unindent();
+      codeWriter.unindent();
+      codeWriter.emit(")$L", suffix);
+    }
+  }
+
+  public static AnnotationSpec of(Class<? extends Annotation> annotation) {
+    return new Builder().type(ClassName.fromClass(annotation)).build();
+  }
+
+  @Override public boolean equals(Object o) {
+    return o instanceof AnnotationSpec
+        && ((AnnotationSpec) o).type.equals(type)
+        && ((AnnotationSpec) o).members.equals(members);
+  }
+
+  @Override public int hashCode() {
+    return type.hashCode() + 37 * members.hashCode();
+  }
+
+  public static final class Builder {
+    private ClassName type;
+    private final SortedMap<String, Snippet> members = Maps.newTreeMap();
+
+    public Builder type(ClassName type) {
+      this.type = type;
+      return this;
+    }
+
+    public Builder addMember(String name, String format, Object... args) {
+      members.put(name, new Snippet(format, args));
+      return this;
+    }
+
+    public AnnotationSpec build() {
+      return new AnnotationSpec(this);
+    }
+  }
+}
--- a/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
+++ b/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
@@ -15,17 +15,24 @@
  */
 package com.squareup.javawriter.builders;
 
+import com.google.common.base.Ascii;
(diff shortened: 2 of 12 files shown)


src/test/java/com/squareup/javawriter/builders/TypeSpecTest.java
```diff
@@ -33,22 +36,96 @@
...
+    ParameterizedTypeName simpleThungOfBar = ParameterizedTypeName.create(simpleThung, bar);
+
+    ParameterSpec thungParameter = new ParameterSpec.Builder()
+        .addModifiers(Modifier.FINAL)
+        .type(thungOfSuperFoo)
+        .name("thung")
+        .build();
+    TypeSpec aSimpleThung = new TypeSpec.Builder()
+        .supertype(simpleThungOfBar)
+        .anonymousTypeArguments("$N", thungParameter)
```
swankjesse: FYI: this is cool. Using $N to refer to a parameter automatically extracts its name, including name collision avoidance magic. So if the parameter was automatically renamed to "thung_1", then this would Do The Right Thing.

src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
```diff
@@ -0,0 +1,103 @@
...
+  private AnnotationSpec(Builder builder) {
+    this.type = checkNotNull(builder.type);
+    this.members = ImmutableSortedMap.copyOf(builder.members);
+  }
+
+  void emit(CodeWriter codeWriter) {
+    if (members.isEmpty()) {
+      // @Singleton
+      codeWriter.emit("@$T\n", type);
+    } else if (members.keySet().equals(ImmutableSortedSet.of("value"))) {
```
JakeWharton: Sorted isn't really useful here

src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
```diff
@@ -0,0 +1,103 @@
...
+      codeWriter.emit("@$T(\n", type);
+      codeWriter.indent();
+      for (Iterator<Map.Entry<String, Snippet>> i = members.entrySet().iterator(); i.hasNext();) {
+        Map.Entry<String, Snippet> entry = i.next();
+        codeWriter.emit("$L = ", entry.getKey());
+        codeWriter.emit(entry.getValue());
+        codeWriter.emit(i.hasNext() ? ",\n" : "\n");
+      }
+      codeWriter.unindent();
+      codeWriter.emit(")\n");
```
JakeWharton: Trailing newline should probably be owned by the parent. Methods want
```java
@Hey
@Buddy
String toString();
```
whereas parameters want
```java
@NotYourBuddy @Guy String name
```

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -68,6 +77,24 @@ public CodeWriter popVisibleType(TypeName typeName) {
     return this;
   }
 
+  public void emitAnnotations(ImmutableList<AnnotationSpec> annotations) {
+    if (annotations.equals(OVERRIDE_ONLY)) {
```
JakeWharton: Despite my love for compactness, this seems like a needless special case. Especially since we often see it push lines past 100 chars.

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -68,6 +77,24 @@ public CodeWriter popVisibleType(TypeName typeName) {
...
+    } else {
+      for (AnnotationSpec annotationSpec : annotations) {
+        annotationSpec.emit(this);
+      }
+    }
+  }
+
+  public void emitModifiers(ImmutableSet<Modifier> modifiers) {
+    // TODO(jwilson): sort in standard order.
+    for (Modifier modifier : modifiers) {
```
JakeWharton:
```java
for (Modifier modifier : EnumSet.copyOf(modifiers))
```
gives you order for free

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -101,15 +132,47 @@ public CodeWriter emit(Snippet snippet) {
...
+    } else if (typeName instanceof WildcardName) {
+      WildcardName wildcardName = (WildcardName) typeName;
+      TypeName extendsBound = wildcardName.extendsBound();
+      TypeName superBound = wildcardName.superBound();
+      if (extendsBound != null) {
+        emit("? extends $T", extendsBound);
+      } else if (superBound != null) {
+        emit("? super $T", superBound);
+      }
+      // TODO(jwilson): special case ? for List<?>.
```
JakeWharton: This is just an `else` on the previous `if`. Both will be null on a boundless wildcard.

src/main/java/com/squareup/javawriter/builders/ParameterSpec.java
```diff
@@ -15,25 +15,59 @@
...
+      return this;
+    }
+
+    public Builder addAnnotation(Class<? extends Annotation> annotation) {
+      this.annotations.add(AnnotationSpec.of(annotation));
+      return this;
+    }
+
+    public Builder addModifiers(Modifier... modifiers) {
+      this.modifiers.addAll(Arrays.asList(modifiers));
```
JakeWharton: `Collections.addAll` saves the wrapper object... although we don't need to go full-on nit picky on allocations like we would something in a library.

src/main/java/com/squareup/javawriter/builders/ParameterSpec.java
```diff
@@ -15,25 +15,59 @@
...
   private ParameterSpec(Builder builder) {
+    this.annotations = ImmutableList.copyOf(builder.annotations);
+    this.modifiers = ImmutableSet.copyOf(builder.modifiers);
     this.type = checkNotNull(builder.type);
     this.name = checkNotNull(builder.name);
   }
 
+  void emit(CodeWriter codeWriter) {
+    codeWriter.emitAnnotations(annotations);
+    codeWriter.emitModifiers(modifiers);
```
JakeWharton: I like the builder approach because `.build()` can throw if `modifiers` has inappropriate modifiers like `private`.

src/test/java/com/squareup/javawriter/builders/TypeSpecTest.java
```diff
@@ -33,22 +36,96 @@
...
+        .name(ClassName.create("com.squareup.tacos", "Taco"))
+        .addField(new FieldSpec.Builder()
+            .addModifiers(Modifier.STATIC, Modifier.FINAL)
+            .type(thingThangOfFooBar)
+            .name("NAME")
+            .initializer("$L", aThingThang)
+            .build())
+        .build();
+
+    // TODO: import Thing, and change references from "Thang" to "Thing.Thang"
```
JakeWharton: I don't think this is strictly necessary. It's one of those things a human writes because it knows `Thing.Thang` carries semantic value whereas a soulless (killing?) machine doesn't need it.

src/test/java/com/squareup/javawriter/builders/TypeSpecTest.java
```diff
@@ -33,22 +36,96 @@
...
+    assertThat(toString(taco)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import com.squareup.tacos.Bar;\n"
+        + "import com.squareup.tacos.Foo;\n"
+        + "import com.squareup.tacos.SimpleThung;\n"
+        + "import com.squareup.tacos.Thing.Thang;\n"
+        + "import com.squareup.tacos.Thung;\n"
```
JakeWharton: @edenman would approve of this package... and I think we just found our next library name.

JakeWharton:
LGTM

src/test/java/com/squareup/javawriter/builders/TypeSpecTest.java
edenman: :shipit:

src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
swankjesse: Agreed. Fixed.

src/main/java/com/squareup/javawriter/builders/AnnotationSpec.java
swankjesse: Good idea. Fixed.

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: Foolishly trying to match your example perfectly! Dropped.

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: Yeah, though that throws if the set is empty.

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: Almost. `List<?>` is shorthand for `List<? extends Object>`.

swankjesse: (Fixed.)

src/main/java/com/squareup/javawriter/builders/ParameterSpec.java
swankjesse: Agreed. I want to add those kinds of strict checks in follow-up.

src/main/java/com/squareup/javawriter/builders/ParameterSpec.java
swankjesse: Nice. Fixed x4.

PR #155, opened 2015-01-09
Constructors and abstract methods.

swankjesse:


src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -55,13 +55,21 @@ public CodeWriter(StringBuilder out, ImmutableMap<ClassName, String> importedTyp
...
   }
 
   public CodeWriter unindent() {
-    checkState(indentLevel > 0);
-    indentLevel--;
+    return unindent(1);
+  }
+
+  public CodeWriter unindent(int levels) {
+    checkState(indentLevel - levels >= 0);
```
JakeWharton: Seems like it should be checkArgument w/message of some sort.

src/main/java/com/squareup/javawriter/builders/MethodSpec.java
```diff
@@ -38,6 +40,11 @@
   public final ImmutableList<Snippet> snippets;
 
   private MethodSpec(Builder builder) {
+    checkArgument(builder.returnType != null ^ builder.name == Name.CONSTRUCTOR,
```
JakeWharton: xor. love it.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -42,6 +43,17 @@
...
     checkArgument(builder.name != null ^ builder.anonymousTypeArguments != null,
         "types must have either a name or anonymous type arguments");
+    boolean isInterface = builder.type == Type.INTERFACE;
+    boolean typeIsAbstract = builder.modifiers.contains(Modifier.ABSTRACT) || isInterface;
+    for (MethodSpec methodSpec : builder.methodSpecs) {
+      checkArgument(typeIsAbstract || !methodSpec.hasModifier(Modifier.ABSTRACT),
+          "non-abstract type %s cannot declare abstract method %s", builder.name, methodSpec.name);
+      checkArgument(!isInterface || methodSpec.hasModifier(Modifier.ABSTRACT),
+          "interface %s cannot declare non-abstract method %s", builder.name, methodSpec.name);
+      checkArgument(!isInterface || methodSpec.hasModifier(Modifier.PUBLIC),
```
JakeWharton: We'll have to go through and add tests which verify these conditions correctly fail.

JakeWharton: Later, that is.

JakeWharton: Need this `public` check for interface fields as well

swankjesse: Fixed. Also static and final for fields.

JakeWharton:
Sorry, LGTM. Was mesmerized by the awesome Retrofit service tests...

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: Fixed. I don't think this needs a message; it's not a public interface.

PR #157, opened 2015-01-10
Type variables, implements and extends.

swankjesse:
```
TypeVariables are tricky because in the declaration we write it as
<T extends CharSequence> but everywhere else it's just <T>.
```

### DIFF
--- a/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
+++ b/src/main/java/com/squareup/javawriter/builders/CodeWriter.java
@@ -25,6 +25,7 @@ import com.squareup.javawriter.ParameterizedTypeName;
 import com.squareup.javawriter.StringLiteral;
 import com.squareup.javawriter.TypeName;
 import com.squareup.javawriter.TypeNames;
+import com.squareup.javawriter.TypeVariableName;
 import com.squareup.javawriter.WildcardName;
 import java.util.ArrayList;
 import java.util.EnumSet;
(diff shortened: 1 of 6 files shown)


src/main/java/com/squareup/javawriter/builders/CodeWriter.java
```diff
@@ -108,6 +109,27 @@ public void emitModifiers(ImmutableSet<Modifier> modifiers) {
     emitModifiers(modifiers, ImmutableSet.<Modifier>of());
   }
 
+  /**
+   * Emit type variables with their qualifiers. This should only be used when declaring type
```
eamonnmcmanus: By qualifiers do you mean bounds?

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -97,13 +108,48 @@ void emit(CodeWriter codeWriter, String enumName) {
...
+      codeWriter.emit("$L $L", Ascii.toLowerCase(type.name()), name.simpleName());
+      codeWriter.emitTypeVariables(typeVariables);
+
+      List<TypeName> extendsTypes;
+      List<TypeName> implementsTypes;
+      if (type == Type.INTERFACE) {
+        extendsTypes = superinterfaces;
+        implementsTypes = ImmutableList.of();
+      } else {
+        extendsTypes = !superclass.equals(ClassName.OBJECT)
```
eamonnmcmanus: Invert this condition?

kenzierocks: Superclass -> OBJECT: Empty list
Superclass -/> OBJECT: List of superclass

I think that looks right.

eamonnmcmanus: The code is right, but I was suggesting that inverting the condition and the two branches of the ?: would make it a bit easier to read.

kenzierocks: Oh, yes. I agree.

JakeWharton:
lgtm :shipit:

src/main/java/com/squareup/javawriter/builders/CodeWriter.java
swankjesse: Yes I do. Fixed.

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
swankjesse: I like to write ternarys as `(expression) ? (positive result) : (negative result)`. I think of having types to print is the positive result, and no types to print is the negative result.

It's annoying 'cause this expression is `(not degenerate)` but overall I think it's consistent.

swankjesse: (But in this case I think you're right and I've flipped it!)

PR #158, opened 2015-01-10
Nested classes and imports.

swankjesse:
This changes TypeSpec to no longer know what its enclosing package,
and potentially also not its enclosing class if there was one. This
makes TypeSpec work the same as FieldSpec and MethodSpec, which also
don't know how they are enclosed. As a model I think it's simpler,
particularly since it was often a lie anyway: what is the fully qualified
name of an anonymous class?

This also improves imports for nested classes.

### DIFF
--- a/src/main/java/com/squareup/javawriter/builders/FieldSpec.java
+++ b/src/main/java/com/squareup/javawriter/builders/FieldSpec.java
@@ -58,6 +58,18 @@ public final class FieldSpec {
     codeWriter.emit(";\n");
   }
 
+  public static FieldSpec of(Class<?> type, String name, Modifier... modifiers) {
+    return of(TypeNames.forClass(type), name, modifiers);
+  }
+
+  public static FieldSpec of(TypeName type, String name, Modifier... modifiers) {
+    return new Builder()
+        .type(type)
+        .name(name)
+        .addModifiers(modifiers)
+        .build();
+  }
+
   public static final class Builder {
     private final List<AnnotationSpec> annotations = new ArrayList<>();
     private final List<Modifier> modifiers = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


src/main/java/com/squareup/javawriter/builders/FieldSpec.java
```diff
@@ -58,6 +58,18 @@ void emit(CodeWriter codeWriter, ImmutableSet<Modifier> implicitModifiers) {
     codeWriter.emit(";\n");
   }
 
+  public static FieldSpec of(Class<?> type, String name, Modifier... modifiers) {
```
swankjesse: Alternately, could have been

```
public static FieldSpec of(Class<?> type, String name, String initializerFormat, Object... initializerArgs) {
```

src/main/java/com/squareup/javawriter/builders/TypeSpec.java
```diff
@@ -160,7 +163,13 @@ void emit(CodeWriter codeWriter, String enumName) {
       if (!firstMember) codeWriter.emit("\n");
       enumConstant.getValue().emit(codeWriter, enumConstant.getKey());
       firstMember = false;
-      codeWriter.emit("$L", i.hasNext() ? ",\n" : ";\n");
+      if (i.hasNext()) {
+        codeWriter.emit("$L", ",\n");
+      } else if (!fieldSpecs.isEmpty() || !methodSpecs.isEmpty() || !typeSpecs.isEmpty()) {
+        codeWriter.emit("$L", ";\n");
+      } else {
+        codeWriter.emit("$L", "\n");
```
JakeWharton: These can just skip the `$L` and emit the comma/semi/newline.

PR #159, opened 2015-01-10
Close remaining gaps between builders and writers.

swankjesse:
Fix bugs in imports (as suggested by the writers test cases).

Add support for array types and intersection types.

Add support for originating elements.

Migrate JavaWriter to emit builder types instead of writer types.

Delete now-obsolete writer types.


kenzierocks:
It is upon us.

src/test/java/com/squareup/javawriter/builders/TypeSpecTest.java
```diff
@@ -688,6 +693,123 @@
...
+        + "      Top internalTop;\n"
+        + "\n"
+        + "      Bottom internalBottom;\n"
+        + "\n"
+        + "      com.squareup.donuts.Top externalTop;\n"
+        + "\n"
+        + "      com.squareup.donuts.Bottom externalBottom;\n"
+        + "    }\n"
+        + "  }\n"
+        + "}\n");
```
JakeWharton: This test is awesome

JakeWharton:
Awesome

PR #160, opened 2015-01-11
Use java.lang.reflect.Type instead of TypeName.

swankjesse:
This saves a ton of code, and means we don't need to convert from
built-in class references like String.class.


JakeWharton:
You got Java 8 problems:
```
/home/travis/build/square/javawriter/src/main/java/com/squareup/javawriter/Types.java:[179,31] <anonymous com.squareup.javawriter.Types$6> is not abstract and does not override abstract method getAnnotatedBounds() in java.lang.reflect.TypeVariable
```

swankjesse:
Yeah, sadly it's impossible to implement TypeVariable in a way that compiles on Java7 and Java8 simultaneously!

JakeWharton:
Can you just specify the method without `@Override`?

src/main/java/com/squareup/javawriter/CodeWriter.java
```diff
@@ -336,9 +352,8 @@ private void emitIndentationIfNecessary() {
     }
   }
 
-  private TypeName toTypeName(Object arg) {
-    if (arg instanceof TypeName) return (TypeName) arg;
-    if (arg instanceof Class<?>) return TypeNames.forClass((Class<?>) arg);
+  private Type toTypeName(Object arg) {
```
JakeWharton: nit: `toType`

src/main/java/com/squareup/javawriter/TypeSpec.java
```diff
@@ -198,25 +195,22 @@ void emit(CodeWriter codeWriter, String enumName) {
...
-    } else {
-      codeWriter.emit("}\n");
+    codeWriter.emit("}");
+    if (enumName == null && anonymousTypeArguments == null) {
+      codeWriter.emit("\n"); // If this type isn't also a value, include a trailing newline.
     }
   }
 
-  public static enum Type {
+  private static enum DeclarationType {
```
JakeWharton: nit: no `static` needed

src/test/java/com/squareup/javawriter/JavaFileTest.java
```diff
@@ -13,8 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.squareup.javawriter.builders;
+package com.squareup.javawriter;
 
+import com.squareup.javawriter.FieldSpec;
+import com.squareup.javawriter.JavaFile;
+import com.squareup.javawriter.TypeSpec;
```
JakeWharton: These shouldn't be needed anymore

src/test/java/com/squareup/javawriter/TypesTest.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.Set;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.TypeParameterElement;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assert_;
```
JakeWharton: While you're in here, can we nuke this uggo method from orbit in favor of the more palatable `assertThat`?

src/main/java/com/squareup/javawriter/CodeWriter.java
swankjesse: Fixed.

src/main/java/com/squareup/javawriter/TypeSpec.java
swankjesse: Fixed.

JakeWharton:
Net code loss. Shorter type creation. Me like. LGTM :shipit:

The only thing I didn't like was `interfaceType()`/`enumType()`.

src/test/java/com/squareup/javawriter/JavaFileTest.java
swankjesse: Fixed.

src/test/java/com/squareup/javawriter/TypesTest.java
swankjesse: Fixed.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: java
 
 jdk:
-  - oraclejdk7
```
kenzierocks: Does this mean a loss of Java 7 compatibility?

JakeWharton: Only in compilation, yes. Not in binary compatibility.

kenzierocks: Okay, awesome.

swankjesse:
The `interfaceType()` / `enumType()` thing is gross. I'm pondering some options.

### A
Minimizing code, we put things as they were before.

```
TypeSpec.Builder enumBuilder = new TypeSpec.Builder()
    .name("Toppings")
    .declarationType(DeclarationType.ENUM);
```

### B
Offer new factory methods to create a specialized builder, but each returns the same type:

```
TypeSpecBuilder enumBuilder = TypeSpec.enumBuilder("Toppings");
TypeSpecBuilder interfaceBuilder = TypeSpec.interfaceBuilder("TortillaEntre");
TypeSpecBuilder classBuilder = TypeSpec.classBuilder("Taco");
TypeSpecBuilder anonymousClassBuilder = TypeSpec.anonymousClassBuilder("$S", "shredded cheese");
```

### C
Maximum type safety. We introduce some new interfaces, all implemented by `ClassSpec.Builder` behind-the-scenes, and make that class private:

```
EnumBuilder enumBuilder = TypeSpec.enumBuilder("Toppings");
InterfaceBuilder interfaceBuilder = TypeSpec.interfaceBuilder("TortillaEntre");
ClassBuilder classBuilder = TypeSpec.classBuilder("Taco");
ClassBuilder anonymousClassBuilder = TypeSpec.anonymousClassBuilder("$S", "shredded cheese");
```
It's safer because `EnumBuilder` is the only one that has `addEnumConstant`. I'm not sure there's more to it than that. If we go this way, we'll need to decide whether we mirror the extra 4 builder interfaces with 4 extra "constructed" interfaces. Plus whether we add specializations for constructors & methods.


...

I'm leaning towards **B**, with a similar approach to constructors and methods: `MethodSpec.constructorBuilder()` and MethodSpec.methodBuilder("toString()")`.

JakeWharton:
I was going to suggest something almost exactly like B, so it gets my vote as well!

swankjesse:
It's decided.

swankjesse:
**B**: https://github.com/square/javawriter/pull/161

.travis.yml
tbroyer: Are you sure? http://www.draconianoverlord.com/2014/04/01/jdk-compatibility.html

swankjesse: I built on Java 8 and ran tests with Java 7, and they pass. Not sure there's much else to do. (Though automating that pair of operations with Travis would be badass.)

eamonnmcmanus:
Maybe I'm misunderstanding, but it seems to me that this change breaks the barrier between the compiler and runtime worlds that both the javax.lang.model API and JavaWriter prior to this change went to some lengths to maintain. Why does javax.lang.model have TypeMirror etc? What reasons apply there that don't apply here?

swankjesse:
Yes, it breaks the barrier.

We get around the corresponding limitations by supporting _both_ `java.lang.Class<?>` and `com.squareup.javawriter.ClassName`. In a few places this could cause us grief; for example checking whether a `Type` is `java.lang.Object` is more difficult.

That said, we don't need create a third model for type names. All we need is something that satisfies our needs: wildcard types, parameterized types, declared classes, arrays, primitives, etc.

Is there something that this approach prevents?

swankjesse:
(I am tempted to possibly add `Types.canonicalize()` that'll convert `Class` instances into `ClassName` instances, which would eliminate potential `equals()` gotchas.)

eamonnmcmanus:
I guess my concern is that it should be possible to reference a class in any context without having to load that class. So I should be able to emit
```java
@MyAnnotation <T extends MyBound> List<MyClass> myMethod(MyClass2<MyClass3> x) {...}
```
without having to load any of the `My*` classes. Looking at the `Types` class, it appears that it should be possible to do that, although I think it may prove troublesome to support that class in a way that works on both Java 7 and Java 8. I'm pretty sure the current version doesn't compile on Java 7. (Maybe that isn't a goal.)

swankjesse:
100% agree with your concern. You can reference classes without loading them, and you can reference classes that you haven't created yet. The test case has plenty of inane examples. Ultimately you'll end up using `ClassName.create("com.example.foo", "MyClass")`.

swankjesse:
Yup, I've deliberately dropped support for compiling on Java 7, due to madness that is Java 8 adding methods to existing interfaces. Tests still pass on Java 7. That said, I'd love to abandon Java 7 support completely. What caveman doesn't want Java 8 and all its lambda glory?

kenzierocks:
This caveman that still wants to cater to users with Java 7, although I'm considering forcing 8 anyways.

eamonnmcmanus:
Thanks for allaying my concerns. :-) I agree that it's a good idea to get rid of the third representation of types, beside the ones the JDK gives us in java.lang.reflect.Type and javax.lang.model.type.TypeMirror. I wonder though if there's a strong reason to prefer the former to the latter? I mean, Class is already a Type which is nice, but as you note with the `Types.canonicalize()` suggestion it isn't the same sort of Type as the one that you get from ClassName.create. On the other hand, the annotation processors that will be clients of this code have TypeMirrors as their input, all of which they are going to have to run through the Types wringer before being able to use them.

swankjesse:
I think we want overloads so you can pass either `Type` or `TypeMirror` & it just works. It doesn't really matter which we use, though I think `Type` requires less code.

.travis.yml
tbroyer: :+1: 

Too bad Travis CI doesn't support Maven Toolchains out of the box (travis-ci/travis-ci#2727), we could have used a reactor build with 2 modules (or maybe even maven-invoker-plugin; note that this is possible by just shipping our own `toolchains.xml`).
I think we can also achieve this with an explicit call to `jdk_switcher use oraclejdk7` before running `mvn test` again in the `script` step of the `.travis.yml`.

cgruber: Ugh. The Java 7/8 compilation thing is going to be a pain inside google
until we have fully switched over to 8.  :/  not sure how we want to handle
that in our fork.  But we need to stay compiling on 7 until that magic
moment.

src/main/java/com/squareup/javawriter/Types.java
```diff
@@ -0,0 +1,351 @@
...
+            && name.equals(((TypeVariable<?>) o).getName());
+      }
+      @Override public int hashCode() {
+        return name.hashCode();
+      }
+      @Override public String toString() {
+        return typeToString(this);
+      }
+
+      // Java 8 requires these methods. We have them to compile, but we don't exercise them.
```
eamonnmcmanus: Regarding Java 7 compatibility, if this is the only place where that comes into play, it wouldn't actually be very hard to return a dynamic proxy here instead of implementing the TypeVariable interface in code. Especially since we are throwing from the new methods anyway. I did something very similar for Guava's Types class in https://github.com/google/guava/commit/f4aa25e74a9466c2f93a2147a7cf9b01850dd41f.

gk5885:
I'm just now getting to catching up with github, but I share @eamonnmcmanus' concerns.  `TypeName` existed for the particular reason that neither the reflection or the mirror model actually suits the problem.

`TypeMirror` just doesn't work because the few mirrors that sprinkle `asElement` on them completely break.  `TypeName` specifically is a string without any resolution and having `asElement` means that you either have to start throwing exceptions or limit types to those that "exist" while the processor runs.  Plus, the API is just way overspecified for what we're trying to do here.

`Type` breaks down because for better or worse too much code assumes that it knows the type hierarchy.  Since there are no visitors or anything on Type, something will eventually do `instanceof Class` and assume that that is how you test for that subset of Type implementation.  This just isn't a "fail gracefully" situation.  E.g. `TypeToken` (and everything else in c.g.common.reflect) is going to fail pretty hard if you pass it a `ClassName`.

Plus, aside from 200-ish lines of really uninteresting value objects, we're at virtually the same amount of code for this change.  I really think that this change is a misstep and a few saved lines isn't worth the awkward fit into the reflection API.

src/main/java/com/squareup/javawriter/ClassName.java
```diff
@@ -103,42 +102,6 @@ public String canonicalName() {
     return fullyQualifiedName;
   }
 
-  public String classFileName() {
```
gk5885: Why did we get rid of these methods?  All of them had callers in ‡.

swankjesse: Simply, I didn't think we needed 'em for JavaWriter. This is a weird method anyway, since it doesn't include the `.class` suffix that might be expected.

What does Dagger use this for? Would it make more sense to define it where it's needed?

src/main/java/com/squareup/javawriter/ClassName.java
```diff
@@ -103,42 +102,6 @@ public String canonicalName() {
...
-            .addAll(enclosingSimpleNames())
-            .add(simpleName())
-            .build(),
-        memberClassName);
-  }
-
-  public ClassName peerNamed(String peerClassName) {
-    checkNotNull(peerClassName);
-    checkArgument(SourceVersion.isIdentifier(peerClassName));
-    checkArgument(Ascii.isUpperCase(peerClassName.charAt(0)));
```
swankjesse: ... these methods are all pretty suspicious anyway. This breaks for `_Remote_Stub`, a foolishly-named JDK class.

swankjesse:
First feedback addressed: https://github.com/square/javawriter/pull/167

swankjesse:
Reusing the reflection API has some advantages...

It's handy when generating code. Passing String.class, int.class, etc is used to good effect in the TypeSpecTest.

We can reuse existing APIs. We really don't need to add our own parameterized type interface. This gets potentially more significant with Java 8 type annotations, where there's also `AnnotatedParameterizedType` for example. The previous API had 11 public classes for the type name APIs; this new API has 2.

Fewer mappings. We still map the Mirror API to the reflect API. But we don't need to also map the reflect API to a third, JavaWriter-specific API.

gk5885:
Right, but no matter how many classes you save, the new `Type` subtypes will never, ever be well-behaving members of the reflection API.  "Reflection allows programmatic access to information about the fields, methods and constructors of loaded classes" and this library quite often refers to types that don't (and might never) exist let alone be loaded.

Plus, I think we're pretty conveniently ignoring some of the places that we're almost certainly just violating contracts.  E.g. `ParametizedType` says that "Instances of classes that implement this interface must implement an equals() method that equates any two instances that share the same generic type declaration and have equal type parameters."  Of course this won't work for any `ParameterizedType` that tries to compare `Class` instances to `ClassName` instances.

If we're truly offended by the extra types, the I would rather see us standardize on the `TypeMirror` API, which at least hand-waves at trying to support types that aren't on the classpath.

eamonnmcmanus:
I agree with Greg. Using java.lang.reflect.Type because it happens to have interfaces that are more or less like what we need is strange. I also think it will encourage people to use reflection when they would be better off using the javax.lang.model API to retrieve the same information. It might be convenient to accept Class<?> in a couple of places but I wouldn't go any further than that.

Using TypeMirror is a decentish alternative to the custom types that there were before. It's a bit strange for TypeVariable since there your input is just as often a TypeParameterElement rather than a TypeVariable. Notice that TypeVariable doesn't even have a method to tell you the name of the variable. You have to use asElement() and get the name from there. That wouldn't be a big concern if you have fake implementations returned from static factory methods (as you currently do for the java.lang.reflect interfaces), because you'd just concoct the appropriate fake Element and put it inside the fake TypeVariable. It's certainly no stranger than throwing UnsupportedOperationException from the methods of java.lang.reflect.TypeVariable where it's not obvious what else to do.

PR #162, opened 2015-01-11
Javadoc.

swankjesse:


JakeWharton:
Awesome.

What do you think about having `addParameter`, `addThrows`, and `addTypeParameter` overloads which take a javadoc string and propagates it to `@param`, `@throws`, and `@param` (for types)? Too clever? Not useful enough? Or just right?

swankjesse:
The @param stuff isn't urgent. We should wait til we have enough callers of the existing methods.

PR #163, opened 2015-01-11
Run tests with OpenJDK 7 to verify runtime compatibility

tbroyer:


kenzierocks:
*watches travis*

JakeWharton:
Looks like you need to skip test compilation as well. Oh, Maven. A better build system would know it already compiled them...

kenzierocks:
Like gradle?

tbroyer:
The problem is the `tools.jar` depends on the `$JAVA_HOME`, so it's different between invocations and Maven then **correctly** recompiles as dependencies have changed.

Invoking `surefire:test` won't run the compile plugin.

PR #164, opened 2015-01-11
Drop name collision avoidance. Tweak ClassName API.

swankjesse:
Name collision avoidance is difficult to do properly at this level
of abstraction. In particular, if we pick a different name for
something, it's awkward to surface that to the caller.

Also tweak the ClassName API. Use a single list for all simple
names, even the class' own simple name. This makes some of the
calling code simpler.


src/main/java/com/squareup/javawriter/ClassName.java
```diff
@@ -130,94 +86,100 @@ public static ClassName fromClass(Class<?> clazz) {
...
 
-  public static ClassName create(
-      String packageName, List<String> enclosingSimpleNames, String simpleName) {
-    return new ClassName(packageName, ImmutableList.copyOf(enclosingSimpleNames),
-        simpleName);
+  /**
+   * Returns a class name created from the given parts. For example, calling this with package name
+   * {@code "java.util"} and simple names {@code "Map"}, {@code "Entry"} yields {@link Map.Entry}.
+   */
+  public static ClassName get(String packageName, String... simpleNames) {
```
tbroyer: Should there be a `checkArgument(simpleNames.length > 0)` to avoid calls like `ClassName.get("java.util.Map.Entry") /* oops! I actually wanted bestGuess! */`? or use a `(String packageName, String simpleName, String... simpleNames)` signature to enforce it at compile-time?

swankjesse: Oooh, very good point. Fixed.

src/main/java/com/squareup/javawriter/ClassName.java
```diff
@@ -130,94 +89,98 @@ public static ClassName fromClass(Class<?> clazz) {
...
-  private static PackageElement getPackage(Element type) {
-    while (type.getKind() != ElementKind.PACKAGE) {
-      type = type.getEnclosingElement();
+    List<String> names = Lists.newArrayList();
+    for (Class<?> c = clazz; c != null; c = c.getEnclosingClass()) {
+      names.add(c.getSimpleName());
     }
-    return (PackageElement) type;
+    names.add(clazz.getPackage().getName());
+    Collections.reverse(names);
```
JakeWharton: Clever

src/main/java/com/squareup/javawriter/JavaWriter.java
```diff
@@ -30,6 +30,7 @@
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+/** Writes generated types to a filesystem using the required directory structure. */
```
JakeWharton: nit: required --> recommended? javac doesn't care about folder structure.

src/main/java/com/squareup/javawriter/ParameterSpec.java
```diff
@@ -49,22 +51,24 @@ void emit(CodeWriter codeWriter) {
...
     private final Type type;
-    private final Name name;
+    private final String name;
 
     private final List<AnnotationSpec> annotations = new ArrayList<>();
     private final List<Modifier> modifiers = new ArrayList<>();
 
-    private Builder(Type type, Name name) {
+    private Builder(Type type, String name) {
+      checkArgument(SourceVersion.isName(name), "not a valid name: %s", name);
```
JakeWharton: I always forget about this API. Such an odd place for it.

src/main/java/com/squareup/javawriter/JavaWriter.java
swankjesse: Fixed to `standard`.

gk5885:
Just so you know, collision avoidance is absolutely critical to implementing ‡ and there must be equivalent functionality before this rewrite could be used there.  Do you have a plan for how we'll be able to handle this?  (And FWIW, surfacing it to the caller seemed pretty straightforward to me…)

PR #165, opened 2015-01-12
Land the builders branch in master

swankjesse:
This is completely incompatible with the previous API.

Still pending: follow up to @gk5885's feedback on the branch.


swankjesse:
Abandoning to deal with the rebase.

PR #166, opened 2015-01-13
Support nested annotations and multiple annotation values.

swankjesse:


src/main/java/com/squareup/javawriter/CodeWriter.java
```diff
@@ -214,6 +215,9 @@ private void emitLiteral(Object o) {
     if (o instanceof TypeSpec) {
       TypeSpec typeSpec = (TypeSpec) o;
       typeSpec.emit(this, null);
+    } else if (o instanceof AnnotationSpec) {
+      AnnotationSpec annotationSpec = (AnnotationSpec) o;
+      annotationSpec.emit(this, true);
```
JakeWharton: Passing `true` for inline is a bit odd—at least as far as consistency goes. I like the output it produces in our test case, but if you had triply(!) nested annotations you would get undesirable behavior.

JakeWharton:
LGTM. One question.

src/main/java/com/squareup/javawriter/CodeWriter.java
swankjesse: Agreed. I think the best behavior here is to compute it magically, possibly by measuring how long the lines are. I'll keep this as-is for now, but I want to do better. (Later).

PR #167, opened 2015-01-13
Address feedback on pull request #160.

swankjesse:


src/main/java/com/squareup/javawriter/CodeWriter.java
```diff
@@ -228,24 +228,19 @@ private CodeWriter emitType(Object arg) {
...
-        if (void.class == classType) return emit("void");
-      } else if (classType.isArray()) {
-        return emit("$T[]", classType.getComponentType());
-      } else {
-        return emitType(ClassName.get(classType));
-      }
-    }
-
-    if (type instanceof ParameterizedType) {
+      if (classType == boolean.class) return emit("boolean");
```
eamonnmcmanus: if (classType.isPrimitive()) return emit(classType.getName());
?

swankjesse: Brilliant.

PR #168, opened 2015-01-13
Fix varargs for builders.

swankjesse:
Closes https://github.com/square/javawriter/issues/148


src/main/java/com/squareup/javawriter/MethodSpec.java
```diff
@@ -75,9 +85,10 @@ void emit(CodeWriter codeWriter, String enclosingName, ImmutableSet<Modifier> im
     }
 
     boolean firstParameter = true;
-    for (ParameterSpec parameterSpec : parameters) {
+    for (Iterator<ParameterSpec> i = parameters.iterator(); i.hasNext(); ) {
```
JakeWharton: Checkstyle doesn't like the whitespace after the last `;` here.

JakeWharton:
LGTM

PR #169, opened 2015-01-13
Top of file comments. Stream to the filesystem.

swankjesse:
Closes https://github.com/square/javawriter/issues/125

### DIFF
--- a/src/main/java/com/squareup/javawriter/CodeWriter.java
+++ b/src/main/java/com/squareup/javawriter/CodeWriter.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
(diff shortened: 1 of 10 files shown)


src/main/java/com/squareup/javawriter/CodeWriter.java
```diff
@@ -365,42 +379,46 @@ private int commonPrefixLength(ImmutableList<String> classNames) {
...
+          }
         }
-      }
 
-      out.append(line);
+        out.append(line);
+        trailingNewline = false;
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
JakeWharton: Why the rethrow? Don't want IO leaking all over?

JakeWharton:
LGTM. Needs rebase.

src/main/java/com/squareup/javawriter/CodeWriter.java
swankjesse: Fixed.

PR #170, opened 2015-01-13
Test that the default package works.

swankjesse:
Closes https://github.com/square/javawriter/issues/80

### DIFF
--- a/src/test/java/com/squareup/javawriter/JavaFileTest.java
+++ b/src/test/java/com/squareup/javawriter/JavaFileTest.java
@@ -16,6 +16,7 @@
 package com.squareup.javawriter;
 
 import java.util.Date;
+import javax.lang.model.element.Modifier;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/squareup/javawriter/JavaFileTest.java
```diff
@@ -73,4 +74,26 @@
...
+                .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                .addParameter(String[].class, "args")
+                .addCode("$T.out.println($S);\n", System.class, "Hello World!")
+                .build())
+            .build())
+        .build()
+        .toString();
+    assertThat(source).isEqualTo(""
+        + "import java.lang.String;\n"
+        + "import java.lang.System;\n"
```
JakeWharton: these shouldn't be imported

JakeWharton: I see that you mention we incorrectly import these in the other issue.

PR #178, opened 2015-01-14
Code blocks.

swankjesse:
Code blocks support old-style JavaWriter control flow APIs, though
they're less strict about balancing.

They can also be embedded as literals.


src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+package com.squareup.javapoet;
+
+import com.google.common.collect.ImmutableList;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * A deferred format string. Unlike {@link java.text.Format} which uses percent {@code %} to escape
```
JakeWharton: Here you say "percent" and then the percent character whereas with the dollar sign we only show the character. We should standardize on one of the two ways.

swankjesse: I think this entire doc needs work. It's describing how it works, but it really wants to describe what it does.

PR #180, opened 2015-01-14
Update CodeBlock documentation.

swankjesse:
Also fix inline indents.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -21,15 +21,23 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
 /**
- * A deferred format string. Unlike {@link java.text.Format} which uses percent {@code %} to escape
- * placeholders, this uses {@code $}, and has its own set of permitted placeholders:
+ * A fragment of a .java file, potentially containing declarations, statements, and documentation.
+ * Code blocks are not necessarily well-formed Java code, and are not validated. This class assumes
+ * javac will check correctness later!
+ *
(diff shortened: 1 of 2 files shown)


JakeWharton:
Loving the linkplain's.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -21,15 +21,23 @@
...
- *   <li>{@code $N} emits a <em>name</em>, using name collision avoidance where necessary.
+ *   <li>{@code $L} emits a <em>literal</em> value with no escaping. Arguments for literals may be
+ *       strings, primitives, {@linkplain TypeSpec type declarations}, {@linkplain AnnotationSpec
+ *       annotations} and even other code blocks.
+ *   <li>{@code $N} emits a <em>name</em>, using name collision avoidance where necessary. Arguments
+ *       for names may be strings, {@linkplain ParameterSpec parameters}, {@linkplain FieldSpec
+ *       fields}, {@linkplain MethodSpec methods}, and {@linkplain TypeSpec types}.
  *   <li>{@code $S} escapes the value as a <em>string</em>, wraps it with double quotes, and emits
- *       that.
- *   <li>{@code $T} emits a <em>type</em> reference. Types will be imported if possible.
```
tbroyer: Documentation for `$T` is now missing.

As a side note, it would be great it `TypeMirror`s were supported for `$T` placeholders, with `CodeWriter#toType` automatically calling `Types.get((TypeMirror) arg)`.
Currently, because a `TypeMirror#toString()` gives you the type's name, the error message is a bit weird: `java.lang.IllegalArgumentException: expected type but was test.OuterType.A`.

tbroyer: #182

PR #182, opened 2015-01-15
Support TypeMirror and Element for $T, and any CharSequence for $N

tbroyer:


tbroyer:
Note: I didn't know where to add tests. I just made sure my ongoing work on Bullet to move to JavaPoet works.

swankjesse:
Nice.

PR #184, opened 2015-01-15
Small API ergonomics.

swankjesse:
Don't require the caller to repeat the type when that's redundant.
addField(FieldSpec.of()) is now just add(FieldSpec.of()). This gives
us the option later of introducing a member type, and interleaving
members with a single add() method.

Also make it possible to write basic control flow without creating
an explicit code block.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -75,6 +75,19 @@ public final class ClassName implements Type, Comparable<ClassName> {
     return names.subList(1, names.size());
   }
 
+  /** Returns a class enclosed by this class. */
+  public ClassName nestedClass(String name) {
+    return new ClassName(new ImmutableList.Builder<String>().addAll(names).add(name).build());
+  }
+
+  /** Returns a class enclosed by the same package or class as this class. */
+  public ClassName peerClass(String name) {
+    return new ClassName(new ImmutableList.Builder<String>()
+        .addAll(names.subList(0, names.size() - 1))
+        .add(name)
+        .build());
+  }
+
   /** Returns the simple name of this class, like {@code "Entry"} for {@link Map.Entry}. */
   public String simpleName() {
     return Iterables.getLast(names);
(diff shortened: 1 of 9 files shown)


diodfr:
I am not sure that it will improve the usability.
When you are new to this project and you use auto completion in your IDE you will be prompt with every 'add' methods.
I think it will be harder to find the good way to add Fields or whatever.
Thanks for this tool.
A new user 😀

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -75,6 +75,19 @@ public ClassName enclosingClassName() {
     return names.subList(1, names.size());
   }
 
+  /** Returns a class enclosed by this class. */
+  public ClassName nestedClass(String name) {
```
JakeWharton: Wtf. You stealing my PR #183?

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -75,6 +75,19 @@ public ClassName enclosingClassName() {
...
+    return new ClassName(new ImmutableList.Builder<String>().addAll(names).add(name).build());
+  }
+
+  /** Returns a class enclosed by the same package or class as this class. */
+  public ClassName peerClass(String name) {
```
JakeWharton: What drove this addition?

tbroyer:
I agree with @diodfr. I find the resulting code less readable (`.add(PUBLIC)` ?!), and I like completing `addField` and then Ctrl+P (in IntelliJ) tells me I need to pass the type, name an optional modifiers, or if I need more (e.g. annotations, javadoc) a `FieldSpec`. With separate `addField(Type,String,Modifier...)` and `add(FieldSpec)` you lose this ability with no real gain in readability. Not worth it if you ask me.

:-1:

JakeWharton:
Yep. Was just about to echo what @tbroyer.

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Derp. I'll rebase!

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Dagger wants it.

swankjesse:
Works for me. I'll try again with just the easier-access CodeBlock methods.

PR #185, opened 2015-01-16
API ergonomics redux.

swankjesse:
Drop a convenience overload for addParameter that was inconsistent;
we don't have similar shortcuts for fields.

Also restore peerNamed() needed by Dagger 2.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -98,10 +98,6 @@ public final class AnnotationSpec {
     codeWriter.emit(whitespace + "}");
   }
 
-  public static AnnotationSpec of(Type annotation) {
-    return builder(annotation).build();
-  }
-
   public static Builder builder(Type type) {
     return new Builder(type);
   }
@@ -126,7 +122,7 @@ public final class AnnotationSpec {
     }
 
     public Builder addMember(String name, String format, Object... args) {
-      members.put(name, CodeBlock.of(format, args));
+      members.put(name, CodeBlock.builder().add(format, args).build());
       return this;
     }
 
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -75,7 +75,7 @@ public final class ClassName implements Type, Comparable<ClassName> {
    * Returns a new {@link ClassName} instance for the specified {@code name} as nested inside this
    * class.
    */
-  public ClassName nestedClassNamed(String name) {
+  public ClassName nestedClass(String name) {
     checkNotNull(name, "name == null");
     return new ClassName(new ImmutableList.Builder<String>()
         .addAll(names)
@@ -85,6 +85,18 @@ public final class ClassName implements Type, Comparable<ClassName> {
 
   public ImmutableList<String> simpleNames() {
     return names.subList(1, names.size());
+  }
+
+  /**
+   * Returns a class that shares the same enclosing package or class. If this class is enclosed by
+   * another class, this is equivalent to {@code enclosingClassName().nestedClass(name)}. Otherwise
+   * it is equivalent to {@code get(packageName(), name)}.
+   */
+  public ClassName peerClass(String name) {
+    return new ClassName(new ImmutableList.Builder<String>()
+        .addAll(names.subList(0, names.size() - 1))
+        .add(name)
+        .build());
   }
 
   /** Returns the simple name of this class, like {@code "Entry"} for {@link Map.Entry}. */
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -57,17 +57,20 @@ public final class CodeBlock {
     this.args = builder.args.build();
   }
 
-  public static CodeBlock of(String format, Object... args) {
-    return new Builder().add(format, args).build();
-  }
-
   public boolean isEmpty() {
     return formatParts.isEmpty();
+  }
+
+  public static Builder builder() {
+    return new Builder();
   }
 
   public static final class Builder {
     final ImmutableList.Builder<String> formatParts = ImmutableList.builder();
     final ImmutableList.Builder<Object> args = ImmutableList.builder();
+
+    private Builder() {
+    }
 
     public Builder add(String format, Object... args) {
       int expectedArgsLength = 0;
@@ -144,7 +147,7 @@ public final class CodeBlock {
       return this;
     }
 
-    public Builder statement(String format, Object... args) {
+    public Builder addStatement(String format, Object... args) {
       return add(format + ";\n", args);
     }
 
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -89,23 +89,22 @@ public final class JavaFile {
     }
   }
 
+  public static Builder builder(String packageName, TypeSpec typeSpec) {
+    return new Builder(packageName, typeSpec);
+  }
+
   public static final class Builder {
-    private CodeBlock.Builder fileComment = new CodeBlock.Builder();
-    private String packageName = "";
-    private TypeSpec typeSpec;
+    private final String packageName;
+    private final TypeSpec typeSpec;
+    private CodeBlock.Builder fileComment = CodeBlock.builder();
 
-    public Builder fileComment(String format, Object... args) {
+    private Builder(String packageName, TypeSpec typeSpec) {
+      this.packageName = checkNotNull(packageName, "packageName == null");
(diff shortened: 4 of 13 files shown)


src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -187,10 +187,6 @@ public Builder addParameter(ParameterSpec parameterSpec) {
       return this;
     }
 
-    public Builder addParameter(Type type, String name, Modifier... modifiers) {
```
tbroyer: Why?

swankjesse: Good question. Foolish consistency? I don't like how there's a bunch of different mechanisms to do the same thing: `ParameterSpec.of`, 
`Parameter.builder`, and this.

I want to clear out everything that I don't immediately need, and then add em back as necessary.

In this case, its tempting to keep this method & drop ParamaterSpec.of, which are mutually redundant.

Like that better?

tbroyer: Ah, so this is why you're duplicating the `CodeBlock.Builder` methods in `MethodSpec.Builder` ;-)

I'd indeed prefer dropping `ParameterSpec.of()` rather than this `addParameter` overload; that'd be consistent with `JavaPoet#add(String,TypeSpec)`.

swankjesse: Fixed. No more `of()` methods!

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,10 +98,6 @@ private void emitAnnotationValue(CodeWriter codeWriter, String whitespace, Strin
     codeWriter.emit(whitespace + "}");
   }
 
-  public static AnnotationSpec of(Type annotation) {
-    return builder(annotation).build();
-  }
-
   public static Builder builder(Type type) {
```
tbroyer: Oh BTW, we're not consistent between static factories (here) and public constructor (e.g. `JavaFile` or `CodeBlock`) for builders.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -75,6 +75,14 @@ public ClassName enclosingClassName() {
     return names.subList(1, names.size());
   }
 
+  /** Returns a class enclosed by the same package or class as this class. */
```
tbroyer: Maybe say that `this is equivalent to {@code className.enclosingClassName().nestedClassNamed(name)}` so it's not to be confused with `ClassName.get(className.packageName(), name)`

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: Good call.

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Yeah, though there isn't necessarily an enclosing class.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Fixed.

swankjesse:
I feel this is still a little bit awkward, but I think it's at least more regular which is a good starting point. It's the result of attempting to port one project over to JavaPoet from JavaWriter 2.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -89,23 +89,22 @@ public String toString() {
...
-    }
+  public static final class Builder {
+    private final String packageName;
+    private final TypeSpec typeSpec;
+    private CodeBlock.Builder fileComment = CodeBlock.builder();
 
-    public Builder packageName(String packageName) {
-      this.packageName = checkNotNull(packageName);
-      return this;
+    public Builder(String packageName, TypeSpec typeSpec) {
```
JakeWharton: `private`

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -89,23 +89,22 @@ public String toString() {
...
+      this.packageName = checkNotNull(packageName, "packageName == null");
+      this.typeSpec = checkNotNull(typeSpec, "typeSpec == null");
     }
 
-    public Builder typeSpec(TypeSpec typeSpec) {
-      this.typeSpec = typeSpec;
+    public Builder fileComment(String format, Object... args) {
```
JakeWharton: This is additive. Prefix with 'add' to match convention?

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -315,15 +317,16 @@ public Builder addEnumConstant(String name, TypeSpec typeSpec) {
       return this;
     }
 
+    public Builder addField(Type type, String name, Modifier... modifiers) {
+      return addField(FieldSpec.builder(type, name, modifiers).build());
+    }
+
+
```
JakeWharton: nit: double newline

JakeWharton: Also if you want to get super-crazy nitty I think everywhere else the overload the takes a `-Spec` comes before these convenience versions.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -144,7 +147,7 @@ public Builder endControlFlow(String controlFlow, Object... args) {
       return this;
     }
 
-    public Builder statement(String format, Object... args) {
+    public Builder addStatement(String format, Object... args) {
```
JakeWharton: This and `add` need Javadoc to clarify their behavior and difference. Future PR.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -87,6 +87,18 @@ public ClassName nestedClassNamed(String name) {
     return names.subList(1, names.size());
   }
 
+  /**
+   * Returns a class that shares the same enclosing package or class. If this class is enclosed by
+   * another class, this is equivalent to {@code enclosingClassName().nestedClassNamed(name)}.
+   * Otherwise it is equivalent to {@code get(packageName(), name)}.
+   */
+  public ClassName peerClass(String name) {
```
JakeWharton: I used `nestedClassNamed` for nesting. Either append 'Named' to this method or remove it from mine.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -413,9 +415,9 @@
         .addTypeVariable(t)
         .addTypeVariable(p)
         .addSuperinterface(Types.parameterizedType(Comparable.class, p))
-        .addField(FieldSpec.of(t, "label"))
-        .addField(FieldSpec.of(p, "x"))
-        .addField(FieldSpec.of(p, "y"))
```
JakeWharton: I'm a sucker for the consistency in builder use (and these helper methods on parent specs), but I can't decide if this is tailoring too much to the test case uses. You can see a counter-example to the current PR's change above where you're now forced to use the builder to create a simple `ParameterSpec`.

Most code that is going to be using JavaPoet will be fraught with conditionals, loops, and helper methods that return objects. You can see this in the PR you made to the you-know-what project to make the switch to JavaPoet using this PR.

I'm fine with removing the helpers. But let's not completely write them off yet as we start to actually play with this more.

JakeWharton:
LGTM

src/main/java/com/squareup/javapoet/JavaFile.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/JavaFile.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/TypeSpec.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/CodeBlock.java
swankjesse: Agreed.

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Fixed using the shorter names: `nestedClass`, `peerClass`

src/test/java/com/squareup/javapoet/TypeSpecTest.java
swankjesse: Yeah, I think we'll want to study the calling code to decide where overloads help.

PR #186, opened 2015-01-18
Use dynamic proxies to implement Type interfaces.

swankjesse:


swankjesse:
This should make everything work on Java 7 completely.

kenzierocks:
This looks really cool!

tbroyer:
> This should make everything work on Java 7 completely.

Update `.travis.yml` to make sure? ;-)

tbroyer:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[246,48] cannot find symbol
  symbol:   class IntersectionType
  location: package javax.lang.model.type
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[168,68] cannot find symbol
  symbol:   class IntersectionType
  location: package javax.lang.model.type
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[157,16] reference to get is ambiguous, both method get(javax.lang.model.type.DeclaredType) in com.squareup.javapoet.Types and method get(IntersectionType) in com.squareup.javapoet.Types match
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[161,16] reference to get is ambiguous, both method get(javax.lang.model.type.PrimitiveType) in com.squareup.javapoet.Types and method get(IntersectionType) in com.squareup.javapoet.Types match
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[165,16] reference to get is ambiguous, both method get(javax.lang.model.type.TypeVariable) in com.squareup.javapoet.Types and method get(IntersectionType) in com.squareup.javapoet.Types match
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[168,7] method does not override or implement a method from a supertype
[ERROR] …/javapoet/src/main/java/com/squareup/javapoet/Types.java:[173,16] reference to get is ambiguous, both method get(javax.lang.model.type.WildcardType) in com.squareup.javapoet.Types and method get(IntersectionType) in com.squareup.javapoet.Types match
[INFO] 7 errors 
```

swankjesse:
Yup. Looks like we still can't build on Java 7 without `javax.lang.model.type.IntersectionType`. But I'm guessing that _running_ on Java 7 should be fine.

tbroyer:
Well, it was fine before too (at least for the part that's unit-tested).

`javax.lang.model.type.IntersectionType` is only used in the `TypeVisitor` in `get(TypeMirror)` and it can be replaced with reflection in `defaultAction` to compile in Java 7.

swankjesse:
Agreed. That's what the code did before I got here; it's my fault for stripping it out.

Is that worthwhile? Does anyone need to build JavaPoet on Java 7 ? I'm expecting that once there's a release in Maven Central the norm will be binary dependencies.

JakeWharton:
Yeah. Agreed.

tbroyer:
Well, the only problem is that building with JDK 8's `bootclasspath` can lead in some cases to code that won't run in a Java 7 VM (mostly when using reflection though IIUC, which users are unlikely to use on JavaPoet classes). In JavaPoet's case, we also run unit-tests with Java 7 so we should be safe, except that unit tests don't exercise everything so there's still a risk…
This has surfaced recently in a Maven Dev Hangout about the need for Maven Toolchains <http://takari.io/2015/01/07/maven-dev-hangout.html> I don't know the details of the project, maybe all they need/want is to run (some) tests in a Java 7 VM to make sure their artifacts are compatible… (like we do, but as part of a `mvn verify` rather than through a separate invocation of Maven)

> Does anyone need to build JavaPoet on Java 7 ?

I think that change was suggested/asked by Googlers due to their build infrastructure.
Cc: @eamonnmcmanus @cgruber @gk5885

JakeWharton:
Yep. We ran into a build problem with the binary artifact in another
project which builds on JDK 7.

src/main/java/com/squareup/javapoet/Types.java
```diff
@@ -364,4 +284,48 @@ static Type arrayComponent(Type type) {
       return null;
     }
   }
+
+  /**
+   * Implement reflective types using a dynamic proxy so we can compile on both Java 7 and Java 8,
```
JakeWharton: nit: compile --> run

JakeWharton: er, maybe not. I didn't read the rest of the sentence before commenting

src/test/java/com/squareup/javapoet/TypesTest.java
```diff
@@ -78,16 +78,35 @@ private TypeMirror getType(Class<?> clazz) {
     assertThat(Types.get(typeVariables.get(0).asType()))
         .isEqualTo(Types.typeVariable("Simple"));
     assertThat(Types.get(typeVariables.get(1).asType()))
-        .isEqualTo(Types.typeVariable("ExtendsClass", Number.class));
+        .isEqualTo(Types.typeVariable("ExtendsClass", ClassName.get(Number.class)));
```
JakeWharton: Are these changes strictly necessary?

JakeWharton:
This change still causes the same failure in the other project when building with JDK 7:
```
error: cannot access IntersectionType
    Type typeClass = com.squareup.javapoet.Types.get(returnType);
```

swankjesse:
We definitely want to support execution on Java 7. Do we also want to support building on Java 7?

tbroyer:
Building on Java 7 would make it 99% safe at runtime too (1% being reflection; and that can be exercised in tests)

tbroyer:
FYI another article about the importance of compiling with your target JDK and not rely on `-target` (at least not without setting the bootclasspath): http://developer-blog.cloudbees.com/2014/12/beware-siren-target-call.html
Even Animal Sniffer didn't catch it.

swankjesse:
Obsoleted by https://github.com/square/javapoet/pull/202

Issue #110, opened 2014-12-06
Interface methods should not emit public, abstract


JakeWharton:
Not quite sure the right approach to take with this. We can either propagate these implementation details downward into `Modifiable` or switch to private fields with public getters (e.g., `Set<Modifier> modifiers()`) which can be overridden to return subsets.

PR #187, opened 2015-01-18
Default modifiers.

swankjesse:
Closes https://github.com/square/javapoet/issues/110

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -19,6 +19,7 @@ import com.google.common.base.Ascii;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
@@ -28,6 +29,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
@@ -39,7 +41,7 @@ import static com.google.common.collect.Iterables.getOnlyElement;
 
 /** A generated class, interface, or enum declaration. */
 public final class TypeSpec {
-  public final DeclarationType declarationType;
+  public final Kind kind;
   public final String name;
   public final CodeBlock anonymousTypeArguments;
   public final CodeBlock javadoc;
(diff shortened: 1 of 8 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -906,6 +906,63 @@
...
+  @Test public void defaultModifiersForMemberInterfacesAndEnums() throws Exception {
+    TypeSpec taco = TypeSpec.classBuilder("Taco")
+        .addType(TypeSpec.classBuilder("Meat")
+            .addModifiers(Modifier.STATIC)
+            .build())
+        .addType(TypeSpec.interfaceBuilder("Tortilla")
+            .addModifiers(Modifier.STATIC)
+            .build())
+        .addType(TypeSpec.enumBuilder("Topping")
+            .addModifiers(Modifier.STATIC)
```
JakeWharton: I think this is a bit unfortunate since you'll _always_ have to specify this for nested enums and interfaces. I guess it's only implicit in emission, not specification.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -225,24 +201,43 @@ void emit(CodeWriter codeWriter, String enumName) throws IOException {
...
+  private enum Kind {
+    CLASS(
+        ImmutableSet.<Modifier>of(),
+        ImmutableSet.<Modifier>of(),
+        ImmutableSet.<Modifier>of(),
+        ImmutableSet.<Modifier>of()),
+
+    INTERFACE(
+        ImmutableSet.of(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL),
+        ImmutableSet.of(Modifier.PUBLIC, Modifier.ABSTRACT),
```
JakeWharton: The abstract here prevents default methods.

tbroyer: Looks like we'd need Java8-specific unit-tests (and use a Maven profile to skip them when running tests in a Java 7 VM)

src/test/java/com/squareup/javapoet/TypeSpecTest.java
swankjesse: Nope! This isn't required.

src/main/java/com/squareup/javapoet/TypeSpec.java
swankjesse: I was hoping to save all the Java 8 features for 3.1.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
JakeWharton: No? I thought the containsAll forced them.

swankjesse: There's 2 kinds of implicit modifiers. We enforce modifiers for everything defined _inside_ an interface. But when you define a nested interface or a nested enum, we just omit the `static`, and that isn't enforced.

swankjesse: Ie. we don't do `containsAll()` checks with `asMemberModifiers`.

JakeWharton: Ah, cool. That's what I get for phone-based reviews.

Issue #112, opened 2014-12-06
Thoughtful ordering of members


JakeWharton:
Right now static methods are emitted below constructors.

Seems like the only logical grouping we should make. There are other one's in practice (public static methods before lesser scoped static methods, public instance methods before lesser scoped instance methods, etc.) but I don't think they should be addressed at this time (if ever).

swankjesse:
For 3.0, I propose this ordering:

static fields
instance fields
constructors
methods
types

Within each group, the caller's declaration order is honored. We can make this fancier as necessary.

PR #188, opened 2015-01-18
Members ordering.

swankjesse:
Closes https://github.com/square/javapoet/issues/112


bsbodden:
Is there a way to turn off member ordering and keep the order of the builder methods?

JakeWharton:
No

Issue #106, opened 2014-12-05
Configurable indent


JakeWharton:
To be set on `JavaWriter`

PR #198, opened 2015-01-22
adding configurable indent

pwittchen:
Hi,

I have added configurable indent. It solves issue #106 . Indent can be configured in `JavaPoet` class (old `JavaWriter` class). Exemplary usage of the configurable indent is presented in the following code snippet:
```java
new JavaPoet(new Indent().level(4));
```

In addition, I've added appropriate Unit Tests verifying behavior of the library with new functionality.
I wanted to make code clean and as abstract as possible.

I hope it would be useful for you.

Regards,
Piotr

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -49,7 +49,7 @@ import static com.google.common.base.Preconditions.checkState;
  * honors imports, indentation, and deferred variable names.
  */
 final class CodeWriter {
-  private final String indent = "  ";
+  private final String indent;
   private final Appendable out;
   private int indentLevel;
 
@@ -68,12 +68,13 @@ final class CodeWriter {
    */
   int statementLine = -1;
 
-  public CodeWriter(Appendable out) {
-    this(out, ImmutableMap.<ClassName, String>of());
+  public CodeWriter(Appendable out, String indent) {
+    this(out, indent, ImmutableMap.<ClassName, String>of());
   }
 
-  public CodeWriter(Appendable out, ImmutableMap<ClassName, String> importedTypes) {
+  public CodeWriter(Appendable out, String indent, ImmutableMap<ClassName, String> importedTypes) {
     this.out = checkNotNull(out);
+    this.indent = checkNotNull(indent);
     this.importedTypes = checkNotNull(importedTypes);
   }
 
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -46,14 +46,14 @@ public final class JavaFile {
     this.typeSpec = builder.typeSpec;
   }
 
-  public void emit(Appendable out) throws IOException {
+  public void emit(Appendable out, String indent) throws IOException {
     // First pass: emit the entire class, just to collect the types we'll need to import.
-    CodeWriter importsCollector = new CodeWriter(NULL_APPENDABLE);
+    CodeWriter importsCollector = new CodeWriter(NULL_APPENDABLE, indent);
     emit(importsCollector);
     ImmutableMap<ClassName, String> suggestedImports = importsCollector.suggestedImports();
 
     // Second pass: write the code, taking advantage of the imports.
-    CodeWriter codeWriter = new CodeWriter(out, suggestedImports);
+    CodeWriter codeWriter = new CodeWriter(out, indent, suggestedImports);
     emit(codeWriter);
   }
 
@@ -84,7 +84,7 @@ public final class JavaFile {
   public String toString() {
     try {
       StringBuilder result = new StringBuilder();
-      emit(result);
+      emit(result, "  ");
       return result.toString();
     } catch (IOException e) {
       throw new AssertionError();
--- a/src/main/java/com/squareup/javapoet/JavaPoet.java
+++ b/src/main/java/com/squareup/javapoet/JavaPoet.java
@@ -33,6 +33,12 @@ import static com.google.common.base.Preconditions.checkArgument;
 /** Writes generated types to a filesystem using the standard directory structure. */
 public final class JavaPoet {
   private final List<JavaFile> javaFiles = new ArrayList<>();
+  private String indent = "  ";
+
+  public JavaPoet setIndent(String indent) {
+    this.indent = indent;
+    return this;
+  }
 
   public JavaPoet add(JavaFile javaFile) {
     javaFiles.add(javaFile);
@@ -59,7 +65,7 @@ public final class JavaPoet {
 
       Path outputPath = outputDirectory.resolve(javaFile.typeSpec.name + ".java");
       try (Writer writer = new OutputStreamWriter(Files.newOutputStream(outputPath))) {
-        javaFile.emit(writer);
+        javaFile.emit(writer, indent);
       }
     }
   }
@@ -76,7 +82,7 @@ public final class JavaPoet {
       JavaFileObject filerSourceFile = filer.createSourceFile(fileName,
           Iterables.toArray(javaFile.typeSpec.originatingElements, Element.class));
       try (Writer writer = filerSourceFile.openWriter()) {
-        javaFile.emit(writer);
+        javaFile.emit(writer, indent);
       } catch (Exception e) {
         try {
           filerSourceFile.delete();
--- a/src/test/java/com/squareup/javapoet/JavaFileTest.java
+++ b/src/test/java/com/squareup/javapoet/JavaFileTest.java
@@ -96,11 +96,11 @@ public final class JavaFileTest {
   @Test public void topOfFileComment() throws Exception {
     String source = JavaFile.builder("com.squareup.tacos",
         TypeSpec.classBuilder("Taco").build())
-        .addFileComment("Generated $L by JavaWriter. DO NOT EDIT!", "2015-01-13")
+        .addFileComment("Generated $L by JavaPoet. DO NOT EDIT!", "2015-01-13")
         .build()
         .toString();
     assertThat(source).isEqualTo(""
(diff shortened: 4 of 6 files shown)


JakeWharton:
CI was broken. Kicking it.

JakeWharton:
This is a very heavyweight API for a relatively lightweight concept. JavaWriter just took a string for the indent.

JakeWharton:
This also doesn't allow using tabs which someone might want.

swankjesse:
Yeah, I think we want to just accept a string: `" "` or `" \t"`, typically.

By the way: what indent are you using?

pwittchen:
You are right. I missed `\t` which by the way I used as indent in some of past projects.
What do you think about such concept of API below?

```java
new JavaPoet(new Indent().space(2));
new JavaPoet(new Indent().tab(1));
// or
new JavaPoet(new Indent().tab()); // 
```

Would it be acceptable?
I can easily modify source code on the branch to add something like that.

I didn't implemented something like:
```
new JavaPoet("  ");
or
new JavaPoet("\t");
```

because in first case, you don't know how many spaces are there just by looking at the source code and you don't know that this argument is used for indent before checking documentation.
I've added this parameter in constructor, because I wanted to to set indent during creating the object.

kenzierocks:
Is there a valid use case for non-space-or-tab indents?

JakeWharton:
@kenzierocks If I hate my coworkers: `" /*space*/"`

@pwittchen Having a dedicated `Indent` class isn't appropriate. At the very least it should be an immutable options class which can control multiple factors of how we output code. I'm certain we can think of more than just indent to configure, but starting with only indent is wise.

swankjesse:
I think this is too much API for something so unimportant. What about just this:

```
JavaPoet javaPoet = new JavaPoet();
javaPoet.setIndent("  ");
```

JakeWharton:
Only downside is that's the only mutable part of the library. That said,
LGTM.

pwittchen:
I've updated code. Now, API looks as @swankjesse proposed:
```java
JavaPoet javaPoet = new JavaPoet();
javaPoet.setIndent("  ");
```

Drawback of this solution is the fact that `indent` variable in `JavaPoet` class is mutable and `setIndent()` method have to be called before `add()` method if we want to have different indentation than default.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -49,7 +49,8 @@
  * honors imports, indentation, and deferred variable names.
  */
 final class CodeWriter {
-  private final String indent = "  ";
+  public static final String DEFAULT_INDENT = "  ";
```
swankjesse: Please make this private.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -256,7 +258,6 @@ private CodeWriter emitType(Object arg) throws IOException {
       if (classType.isPrimitive()) return emit(classType.getName());
       if (classType.isArray()) return emit("$T[]", classType.getComponentType());
       return emitType(ClassName.get(classType));
-
```
swankjesse: Please don't reformat this code.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -28,9 +28,11 @@
     @Override public Appendable append(CharSequence charSequence) {
       return this;
     }
+
```
swankjesse: Please don't reformat this code.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -39,11 +41,13 @@
   public final CodeBlock fileComment;
   public final String packageName;
   public final TypeSpec typeSpec;
+  public final String indent;
```
swankjesse: Indent isn't a property of the `JavaFile`. It's only a property of the `CodeWriter`. Please remove this!

swankjesse:
Sorry about being so picky here. The JavaPoet / JavaFile / CodeWriter code is not structured the way we want it to be in order to make manipulating display-stuff in JavaPoet easy.

And that means the advice I gave about adding accessors to JavaPoet makes little sense.

What I want is accessors on JavaPoet, and then to change `JavaFile.emit()` to take a `String indent` parameter in addition to the `Appendable`.

pwittchen:
I've updated code according to your comments.
I'm only not sure about implementation of `emit(Appendable out)` method where I left the comment, but I don't know how to improve this without adding `indent` field to `JavaFile` class. Now `indent` is local variable of `emit(Appendable out, String indent)` method and `toString()` method uses `emit(Appendable out)` method, so output of `toString()` would be inconsistent with `emit(Appendable out, String indent)` method.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -47,13 +47,17 @@ private JavaFile(Builder builder) {
   }
 
   public void emit(Appendable out) throws IOException {
+    emit(out, "  "); //TODO(pwittchen): this call is made because of toString() and can be improved
```
swankjesse: Just hardcode `toString()` to call the emit overload with `"  "`. That's reasonable.

swankjesse: Just hardcode `toString()` to call the emit overload with `"  "`. That's reasonable.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -61,13 +61,14 @@
...
   private boolean trailingNewline;
 
-  public CodeWriter(Appendable out) {
-    this(out, ImmutableMap.<ClassName, String>of());
+  public CodeWriter(Appendable out, String indent) {
+    this(out, ImmutableMap.<ClassName, String>of(), indent);
   }
 
-  public CodeWriter(Appendable out, ImmutableMap<ClassName, String> importedTypes) {
+  public CodeWriter(Appendable out, ImmutableMap<ClassName, String> importedTypes, String indent) {
```
swankjesse: nit: reorder parameters 3 and 2 ? When there's overloads that make a parameter optional, that parameter should be last.

src/main/java/com/squareup/javapoet/JavaPoet.java
```diff
@@ -33,6 +33,12 @@
 /** Writes generated types to a filesystem using the standard directory structure. */
 public final class JavaPoet {
   private final List<JavaFile> javaFiles = new ArrayList<>();
+  private String indent = "  ";
+
+  public JavaPoet setIndent(String indent) {
+    this.indent = indent;
+    return this;
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -96,11 +96,11 @@
   @Test public void topOfFileComment() throws Exception {
     String source = JavaFile.builder("com.squareup.tacos",
         TypeSpec.classBuilder("Taco").build())
-        .addFileComment("Generated $L by JavaWriter. DO NOT EDIT!", "2015-01-13")
+        .addFileComment("Generated $L by JavaPoet. DO NOT EDIT!", "2015-01-13")
         .build()
         .toString();
     assertThat(source).isEqualTo(""
-        + "// Generated 2015-01-13 by JavaWriter. DO NOT EDIT!\n"
+        + "// Generated 2015-01-13 by JavaPoet. DO NOT EDIT!\n"
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/JavaPoetTest.java
```diff
@@ -193,4 +195,35 @@
...
+        + "\n"
+        + "import java.lang.String;\n"
+        + "import java.lang.System;\n"
+        + "import java.util.Date;\n"
+        + "\n"
+        + "class Test {\n"
+        + "\tDate madeFreshDate;\n"
+        + "\n"
+        + "\tpublic static void main(String[] args) {\n"
+        + "\t\tSystem.out.println(\"Hello World!\");\n"
```
swankjesse: Nice.

swankjesse:
Thanks for your patience. Two last nitpicks & I'm eager to merge this.

swankjesse:
Also: please https://rebaseandsqua.sh
and do our [CLA](https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ).

pwittchen:
I've done the following things:
* performed 2 updates of the code 
 * reordered parameters in constructor of `CodeWriter` class
 * updated `toString()` method in `JavaFile` class and hardcoded indent as you suggested
* merged changes from `square:master` branch with `pwittchen:configurable_indent` branch
* rebased and squashed changes on branch with http://rebaseandsqua.sh website (now there's single commit with changes for this pull request)
* pushed changes
* filled Contributor License Agreement (CLA) form

I hope, now everything is fine :-).

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -68,12 +68,13 @@
...
-  public CodeWriter(Appendable out) {
-    this(out, ImmutableMap.<ClassName, String>of());
+  public CodeWriter(Appendable out, String indent) {
+    this(out, indent, ImmutableMap.<ClassName, String>of());
   }
 
-  public CodeWriter(Appendable out, ImmutableMap<ClassName, String> importedTypes) {
+  public CodeWriter(Appendable out, String indent, ImmutableMap<ClassName, String> importedTypes) {
     this.out = checkNotNull(out);
+    this.indent = checkNotNull(indent);
```
swankjesse: Nice.

swankjesse:
Woohoo!

PR #199, opened 2015-01-24
Double indent wrapped lines.

swankjesse:


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -103,126 +103,136 @@ public static Builder anonymousClassBuilder(String typeArgumentsFormat, Object..
...
-      codeWriter.emit("$L", enumName);
-      if (!anonymousTypeArguments.formatParts.isEmpty()) {
-        codeWriter.emit("(");
+    // Nested classes interrupt wrapped line indentation. Stash the current wrapping state and put
+    // it back afterwards when this type is complete.
+    int previousStatementLine = codeWriter.statementLine;
+    codeWriter.statementLine = -1;
+
+    try {
+      if (enumName != null) {
```
swankjesse: FYI, everything in the `try { }` is unchanged; just indented + 2.

swankjesse:
I'm excited about this. The `@Override` line is indented 6 total when the type its declared inside is on the first line of a statement, and indented 10 total when that type is on a subsequent line. (This is consistent with IntelliJ, but somewhat counter-intuitive.)

swankjesse:
@mattprecious, this one's for you

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1099,4 +1101,93 @@
...
+        + "    return \"Taco(\"\n"
+        + "        + \"beef,\"\n"
+        + "        + \"lettuce,\"\n"
+        + "        + \"cheese\"\n"
+        + "        + \")\";\n"
+        + "  }\n"
+        + "}\n");
+  }
+
+  @Test public void multilineStatementWithAnonymousClass() throws Exception {
```
JakeWharton: Nice test.

PR #200, opened 2015-01-25
Small features for Dagger.

swankjesse:
We had a bug when using Filer to emit to the default package.

Don't wrap null in quotes when emitting it as a string.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -15,7 +15,9 @@
  */
 package com.squareup.javapoet;
 
-import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -49,12 +51,12 @@
...
-  final ImmutableList<String> formatParts;
-  final ImmutableList<Object> args;
+  final List<String> formatParts;
+  final List<Object> args;
 
   private CodeBlock(Builder builder) {
-    this.formatParts = builder.formatParts.build();
-    this.args = builder.args.build();
+    this.formatParts = Collections.unmodifiableList(new ArrayList<>(builder.formatParts));
+    this.args = Collections.unmodifiableList(new ArrayList<>(builder.args));
```
JakeWharton: Aiming to kill Guava? Because that'd be great.

swankjesse: This is to put null into an immutable list.

PR #202, opened 2015-01-25
Support building and running on Java 7.

swankjesse:
This uses dynamic proxies to implement Type interfaces to work
around problems building TypeVariable on Java 8.

This removes IntersectionType, and just uses a list of bound types
when necessary.


swankjesse:
screen shot 2015-01-25 at 11 25 30 am

src/test/java/com/squareup/javapoet/TypesTest.java
```diff
@@ -78,36 +77,34 @@ private TypeMirror getType(Class<?> clazz) {
...
     assertThat(Types.get(typeVariables.get(3).asType()))
         .isEqualTo(Types.typeVariable("ExtendsTypeVariable", Types.typeVariable("Simple")));
-    assertThat(Types.get(typeVariables.get(4).asType()))
-        .isEqualTo(Types.typeVariable("Intersection", Number.class, Runnable.class));
-    assertThat(Types.get(typeVariables.get(5).asType()))
-        .isEqualTo(Types.typeVariable("IntersectionOfInterfaces",
-            Runnable.class, Serializable.class));
+    assertThat(Types.get(typeVariables.get(4).asType())).isEqualTo(
+        Types.typeVariable("Intersection",
+            ClassName.get(Number.class), ClassName.get(Runnable.class)));
```
tbroyer: Why changing from `Number.class` to `ClassName.get(Number.class)`? Does the former cause any issue?

If it's because of, say, equality comparison implementation details, then we should have a comment here, and maybe add a test passing `Number.class` and comparing, say the `toString()`?

swankjesse: Yup, it's because equals changed. Added a comment. The `toString()` is already covered.

.buildscript/deploy_snapshot.sh
```diff
@@ -6,7 +6,7 @@
 # http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/
 
 SLUG="square/javapoet"
-JDK="oraclejdk8"
+JDK="oraclejdk7"
```
JakeWharton: This should still be jdk8 because the Javadocs it creates are so much prettier!

swankjesse: I think we need to build on JDK7 regardless, due to the strange binary-compatibility stuff that @tbroyer mentioned: just the presence of JDK8 signatures during compilation can potentially cause the JDK7 .jar to be malformed.

swankjesse: Relevant link: http://developer-blog.cloudbees.com/2014/12/beware-siren-target-call.html

swankjesse: ... a closer reading tells me that if we can successfully build on JDK7, then a JDK8-built binary targeting JDK7 should work fine. Your call.

JakeWharton: That occurring would be a bug in `javac` I think? For the last year every single release I've made of this library, all of Square's that I touch, and my personal libraries has been made with JDK 8. Yes I do it because I'm vain and want the pretty Javadoc, but I've always relied on the `-target 1.7` to save us.

JakeWharton: Interesting, reading...

swankjesse: I've convinced myself that JDK8 is safe, and flipped this back to JDK8.

src/main/java/com/squareup/javapoet/Types.java
```diff
@@ -266,25 +190,50 @@ private static Type get(DeclaredType t) {
...
-    FluentIterable<TypeMirror> bounds = FluentIterable.from(ImmutableList.of(upperBound));
-    // Try to detect intersection types for Java 7 (Java 8+ has a new TypeKind for that)
-    // Unfortunately, we can't put this logic into Types.get() as this heuristic only really works
-    // in the context of a TypeVariable's upper bound.
+  /**
+   * Returns a list of type mirrors representing the unpacked bounds of {@code typeVariable}. This
+   * is made gnarly by the need to unpack Java 8's new IntersectionType with reflection. We don't
+   * have that type in Java 7, and {@link TypeVariable}'s array of bounds is sufficient anyway.
+   */
+  @SuppressWarnings("unchecked") // Gross things in support of multiple Java 7 and Java 8.
```
JakeWharton: drop "multiple"

PR #203, opened 2015-01-26
Add missing indent.

JakeWharton:


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -68,6 +68,10 @@
    */
   int statementLine = -1;
 
+  CodeWriter(Appendable out) {
```
JakeWharton: 6 classes are looking for this constructor for `toString`.

PR #205, opened 2015-01-27
Don't require any dependencies in JavaPoet.

swankjesse:


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -278,9 +278,9 @@
         + "package com.squareup.tacos;\n"
         + "\n"
         + "@Something(\n"
-        + "    hello = \"goodbye\",\n"
```
swankjesse: This went from being sorted to insertion order. I think this is fine.

gk5885:
To what end?  Who is this change helping?

JakeWharton:
Everyone using a dependency management system (such as a build tool like
maven or gradle). We have problems with Guava (and other dependency)
conflicts all the time because of transitive dependency resolution
semantics.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -0,0 +1,93 @@
...
+      result.append(separator).append(parts.get(i));
+    }
+    return result.toString();
+  }
+
+  public static boolean isLowerCase(char c) {
+    return c >= 'a' && c <= 'z';
+  }
+
+  public static boolean isUpperCase(char c) {
```
JakeWharton: `Character.isUpperCase`

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -0,0 +1,93 @@
...
+    if (parts.isEmpty()) return "";
+    StringBuilder result = new StringBuilder();
+    result.append(parts.get(0));
+    for (int i = 1; i < parts.size(); i++) {
+      result.append(separator).append(parts.get(i));
+    }
+    return result.toString();
+  }
+
+  public static boolean isLowerCase(char c) {
```
JakeWharton: `Character.isLowerCase`

JakeWharton: I guess these aren't ASCII, though.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
JakeWharton: I prefer it :no_mouth: wasn't going to say anything before.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Like Guava, but worse and standalone. This makes it easier to mix JavaPoet with libraries that
+ * bring their own version of Guava.
```
JakeWharton: Accurate, and accurate. We endure the suck so that others (read: me at another point and time) don't have to.

JakeWharton:
Love to hate. Hate to love.

src/main/java/com/squareup/javapoet/Util.java
swankjesse: Fixed.

eamonnmcmanus:
The usual solution is to use something like jarjar or Maven shading rather than reinventing several quite large wheels. Did you consider that?

swankjesse:
Definitely, though jarjar is inelegant. In this case, I'm going with the least bad of two options.

PR #206, opened 2015-01-27
Option to skip java.lang imports.

swankjesse:


src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -112,6 +119,19 @@ public Builder addFileComment(String format, Object... args) {
       return this;
     }
 
+    /**
+     * Call this to omit imports for classes in {@code java.lang}, such as {@code java.lang.String}.
+     *
+     * <p>By default, JavaPoet explicitly imports in {@code java.lang} to defend against naming
```
JakeWharton: "imports types"?

swankjesse: Fixed.

JakeWharton:
Sorry I got distracted!

PR #208, opened 2015-01-27
Character.isUpperCase follow up.

swankjesse:
Being non-ASCII isn't relevant.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -122,7 +122,7 @@ public final class ClassName implements Type, Comparable<ClassName> {
 
     // Add the package name, like "java.util.concurrent", or "" for no package.
     int p = 0;
-    while (p < classNameString.length() && Util.isLowerCase(classNameString.charAt(p))) {
+    while (p < classNameString.length() && Character.isLowerCase(classNameString.codePointAt(p))) {
       p = classNameString.indexOf('.', p) + 1;
       checkArgument(p != 0, "couldn't make a guess for %s", classNameString);
     }
@@ -130,7 +130,7 @@ public final class ClassName implements Type, Comparable<ClassName> {
 
     // Add the class names, like "Map" and "Entry".
     for (String part : classNameString.substring(p).split("\\.", -1)) {
-      checkArgument(!part.isEmpty() && Util.isUpperCase(part.charAt(0)),
+      checkArgument(!part.isEmpty() && Character.isUpperCase(part.codePointAt(0)),
           "couldn't make a guess for %s", classNameString);
       names.add(part);
     }
--- a/src/main/java/com/squareup/javapoet/Util.java
+++ b/src/main/java/com/squareup/javapoet/Util.java
@@ -76,14 +76,6 @@ final class Util {
     return result.toString();
   }
 
-  public static boolean isLowerCase(char c) {
-    return c >= 'a' && c <= 'z';
-  }
-
-  public static boolean isUpperCase(char c) {
-    return c >= 'A' && c <= 'Z';
-  }
-
   public static <T> Set<T> union(Set<T> a, Set<T> b) {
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -122,7 +122,7 @@ public static ClassName bestGuess(String classNameString) {
 
     // Add the package name, like "java.util.concurrent", or "" for no package.
     int p = 0;
-    while (p < classNameString.length() && Util.isLowerCase(classNameString.charAt(p))) {
+    while (p < classNameString.length() && Character.isLowerCase(classNameString.charAt(p))) {
```
eamonnmcmanus: Strictly speaking this should be classNameString.codePointAt(p). There are actually letters in the Supplementary Multilingual Plane, for example [Mathematical Bold Capital A](http://www.fileformat.info/info/unicode/char/1d400/index.htm). It's not very likely you're going to encounter them here, but on the other hand it doesn't cost anything to use the codePoint variants rather than the char ones.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -130,7 +130,7 @@ public static ClassName bestGuess(String classNameString) {
 
     // Add the class names, like "Map" and "Entry".
     for (String part : classNameString.substring(p).split("\\.", -1)) {
-      checkArgument(!part.isEmpty() && Util.isUpperCase(part.charAt(0)),
+      checkArgument(!part.isEmpty() && Character.isUpperCase(part.charAt(0)),
```
eamonnmcmanus: Likewise part.codePointAt(0).

src/main/java/com/squareup/javapoet/ClassName.java
swankjesse: Is that a legal Java identifier?

eamonnmcmanus: Yes. http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8

eamonnmcmanus: As an experiment, in NetBeans I wrote this in a method:
    int \ud835\udc00 = 5;
    int 𝐀 = 5;
It correctly complained that the second 𝐀 was a redefinition of the first (hidden in a surrogate pair). :-)

roman-mazur: In practice you might face with runtime class loading errors if you use
such letters in class names.

eamonnmcmanus: Do you have something specific in mind or is that just a hypothetical possibility? A class loader that couldn't deal with Unicode properly would be buggy.

roman-mazur: A quick example https://ideone.com/68XtFO :)
IMHO class loader can be not always guilty. A filesystem might have troubles with such file names, a bad collation might be set in a database, HTTP header with charset might be not respected...
All this is getting better with time, but still it happens.

JakeWharton: Definitely saw this in Eclipse back in the dark days of when I used Eclipse.

eamonnmcmanus: Fair enough. It's not a reason not to do things properly, though. Just because people _might_ have problems if they use such identifiers doesn't mean that we can just throw them under the bus here.

JakeWharton: Yep, I agree. Just was providing another data point to seeing it not working at one point (a few years ago).

swankjesse: Fixed.

roman-mazur: Definitely.

PR #209, opened 2015-01-28
Don't use java.lang.reflect.Type internally.

swankjesse:
This introduces our own type name hierarchy, inspired by the one
originally in Dagger 2.

This one is called TypeName. Instances are immutable and implement
equals() and hashCode() properly.


src/main/java/com/squareup/javapoet/Types.java
```diff
@@ -1,334 +0,0 @@
...
-  /**
-   * Implement reflective types using a dynamic proxy so we can compile on both Java 7 and Java 8,
-   * even though {@link TypeVariable} gained new methods in Java 8 that won't compile on Java 7.
-   */
-  @SuppressWarnings("unchecked")
-  private static <T extends Type> T newType(
-      final Class<T> type, final Map<String, Object> accessors, final int hashCode) {
-    ClassLoader classLoader = Types.class.getClassLoader();
-    Class[] classes = {type};
-    return (T) Proxy.newProxyInstance(classLoader, classes, new InvocationHandler() {
```
swankjesse: Dead!

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -0,0 +1,185 @@
...
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.ArrayType;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.NoType;
+import javax.lang.model.type.PrimitiveType;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.SimpleTypeVisitor6;
+
+public class TypeName {
```
swankjesse: This needs docs. In particular, how to get type names from mirrors, reflection, and also how to create them programmatically. Also mention that the class is instance-controlled: end-users cannot subclass this, and there is only one TypeName.VOID.

src/main/java/com/squareup/javapoet/FieldSpec.java
```diff
@@ -104,11 +108,15 @@ public Builder addAnnotation(AnnotationSpec annotationSpec) {
       return this;
     }
 
-    public Builder addAnnotation(Type annotation) {
+    public Builder addAnnotation(TypeName annotation) {
       this.annotations.add(AnnotationSpec.builder(annotation).build());
       return this;
     }
 
+    public Builder addAnnotation(Type annotation) {
```
JakeWharton: `MethodSpec` takes a `Class<?>` for this

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -91,11 +95,15 @@ public Builder addAnnotation(AnnotationSpec annotationSpec) {
       return this;
     }
 
-    public Builder addAnnotation(Type annotation) {
+    public Builder addAnnotation(TypeName annotation) {
       this.annotations.add(AnnotationSpec.builder(annotation).build());
       return this;
     }
 
+    public Builder addAnnotation(Type annotation) {
```
JakeWharton: Ditto. Not uniform.

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  CodeWriter emit(CodeWriter out) throws IOException {
+    if (keyword == null) throw new AssertionError();
+    return out.emitAndIndent(keyword);
+  }
+
+  /** Returns a type name equivalent to {@code mirror}. */
+  public static TypeName get(TypeMirror mirror) {
+    return mirror.accept(new SimpleTypeVisitor6<TypeName, Void>() {
```
JakeWharton: `SimpleTypeVisitor7`? :trollface: :shipit:

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -317,32 +316,44 @@ public Builder addAnnotation(AnnotationSpec annotationSpec) {
       return this;
     }
 
-    public Builder addAnnotation(Type annotation) {
+    public Builder addAnnotation(TypeName annotation) {
       return addAnnotation(AnnotationSpec.builder(annotation).build());
     }
 
+    public Builder addAnnotation(Type annotation) {
```
JakeWharton: Another.

Do we ever want to accept `Type`? Or do we just accept `Class<?>` and make you use a `TypeName` if you want something complex. I think that might have been what @gk5885 was saying today.

src/main/java/com/squareup/javapoet/Types.java
JakeWharton: :beers:

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -180,46 +179,62 @@ public Builder addAnnotation(AnnotationSpec annotationSpec) {
...
       return this;
     }
 
-    public Builder addException(Type exception) {
+    public Builder addException(TypeName exception) {
       this.exceptions.add(exception);
       return this;
     }
 
+    public Builder addException(Type exception) {
```
JakeWharton: This can never be anything but a class. Just use `Class<?>`?

src/main/java/com/squareup/javapoet/TypeName.java
swankjesse: Done.

src/main/java/com/squareup/javapoet/MethodSpec.java
swankjesse: Yup.
```
  @SuppressWarnings("unchecked")
  private static <T extends Throwable> void sneakyThrow2(Throwable t) throws T {
    throw (T) t;
  }
```

src/main/java/com/squareup/javapoet/FieldSpec.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/TypeName.java
swankjesse: Fixed!

src/main/java/com/squareup/javapoet/TypeSpec.java
swankjesse: I think we want `Class<?>` / `ClassName` for annotations. I've changed 'em all over to that.

src/main/java/com/squareup/javapoet/ParameterSpec.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/MethodSpec.java
JakeWharton: Ha. Nice example.

PR #213, opened 2015-01-31
Don't allow double field initialization.

ben-manes:
The FieldSpec initializer may be called 0 or 1 times. Previousy,
it could be called multiple times to create a multi-line statement.
As the method name initializer implies a single statement per call,
a CodeBlock may be provided for complex initialization scenarios.


swankjesse:
Thanks!

PR #216, opened 2015-02-02
Add @RunWith boilerplate to tests

cgruber:
Add @RunWith boilerplate to tests - lets alternative test runners in other build systems pick the correct runner.

### DIFF
--- a/src/test/java/com/squareup/javapoet/ClassNameTest.java
+++ b/src/test/java/com/squareup/javapoet/ClassNameTest.java
@@ -21,11 +21,14 @@ import javax.lang.model.element.TypeElement;
 import javax.lang.model.util.Elements;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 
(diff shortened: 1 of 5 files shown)


cgruber:
Note - in theory, we have this in our fork, so this is totally an offer, but we can live with it if you don't want to take this on-board.  But in the interest of minimizing differences, I thought I'd throw this over the wall.

swankjesse:
I've never understood this. Isn't this the default?

cgruber:
It's the default in a standard JUnit4 runner via something like maven.  We
have custom runners which do funky things like identity types of tests and
restrict "short" tests from exercising the filesystem/network (you
know...make them real unit tests). etc.  It is possible we could default
things to JUnit4, or be more smart about detecting Junit4isms... but at
this point, it just runs with JUnit3 semantics (public void methods with
testFoo() naming conventions).

swankjesse:
Got it. Everyone else upgraded to JUnit 4 in the 1990's, but it sounds like everything is more difficult at Google Scale.

PR #218, opened 2015-02-05
Fix WildcardTypeName.get(WildcardType)

eamonnmcmanus:
It would always fail with NulllPointerException since either the extends bound or the super bound or both will be null.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypesTest.java
+++ b/src/test/java/com/squareup/javapoet/TypesTest.java
@@ -26,6 +26,9 @@ import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.type.WildcardType;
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/TypesTest.java
```diff
@@ -157,6 +160,30 @@ private TypeMirror getMirror(Class<?> clazz) {
     assertThat(type.toString()).isEqualTo("? super java.lang.String");
   }
 
+  @Test public void wildcardMirrorNoBounds() throws Exception {
+    WildcardType wildcard = compilation.getTypes().getWildcardType(null, null);
```
swankjesse: That's allowed? Ew!

swankjesse:
Thanks!

swankjesse:
Does this happen with javac-created wildcards are only those created by `Types`?

eamonnmcmanus:
I saw this in an annotation processor when it passed the TypeMirror for java.util.Map<java.lang.String,?> through to TypeName.get. So yes, it does happen with javac-created wildcards.

PR #220, opened 2015-02-09
Box and unbox APIs.

swankjesse:


swankjesse:
These are slightly asymmetric. String.class has no corresponding primitive type. Boxing it returns itself, but unboxing it throws.

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -87,6 +97,45 @@ public boolean isPrimitive() {
     return keyword != null && this != VOID;
   }
 
+  /**
+   * Returns a boxed type like {@code java.lang.Integer} if this is a primitive type like {@code
+   * int} or void. Returns this type if boxing doesn't apply.
```
JakeWharton: I would change this to:

> Returns a boxed type if this is a primitive type (like `Integer` for `int`) or void. Returns this type if boxing doesn't apply.

The two "like" fragments and the part about void reads funny to me. For instance, if I could interpret this as returning `Integer` for `void` depending on how I take the final "or".

swankjesse: Good call. I didn't like this language either.

JakeWharton:
LGTM!

PR #224, opened 2015-02-11
Change CodeBlock.nextControlFlow to be consistent with beginControlFlow.

eamonnmcmanus:
It should include a space before {. Bizarrely, a test checks the current behaviour, but I don't see how it can be desirable.


swankjesse:
Yup, that's preferable. Thanks!

PR #225, opened 2015-02-11
Fix nextControlFlow not accepting arguments

dhleong:
Included a test using try-catch

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -1546,4 +1546,31 @@ public final class TypeSpecTest {
         + "}\n"
     );
   }
+
+  @Test public void tryCatch() {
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1546,4 +1546,31 @@ private String toString(TypeSpec typeSpec) {
...
+      )
+      .build();
+    assertThat(toString(taco)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "class Taco {\n"
+        + "  void addTopping(Topping topping) {\n"
+        + "    try {\n"
+        + "      /* do something tricky with the topping */\n"
+        + "    } catch (IllegalToppingException e) {\n"
```
JakeWharton: :+1: :+1: :+1:

swankjesse:
CLA is good! Thanks!

PR #226, opened 2015-02-11
Switch to non-snapshot dependencies.

### LINKED ISSUE #223: Build instructions
I know this should be easy enough to figure out, but your README should include some basic build instructions. I did git clone the project, and when I run mvn install or mvn clean compile, etc.. I keep getting told artifact for google.truth can't be found. We have an internal maven repository... I am no expert in maven, do you need to specify an external repository perhaps in the pom file? It seems to try to hit our internal repository only, not sure how to get around that.

JakeWharton:
Closes #223.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,8 +22,8 @@
 
     <java.version>1.7</java.version>
     <junit.version>4.12</junit.version>
-    <truth.version>1.0-SNAPSHOT</truth.version>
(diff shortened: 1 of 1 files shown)


swankjesse:
Nice.

justjacksonn:
Thanks.. fixed my build issue as well.

PR #229, opened 2015-02-16
MethodSpec factory for overriding an ExecutableElement.

### LINKED ISSUE #228: Factory for MethodSpec.Builder for overriding ExecutableElement
I'm working in a place where I need to override methods. It would be nice to get a method builder of the same name, return type, params, throws declarations, and already having `@Override`on it.

Maybe `MethodSpec.methodBuilderOverriding(ExecutableElement)`?

JakeWharton:
Anyone know a good way to test this without writing a novel of mocking? I filed https://github.com/google/compile-testing/issues/65 which seems like it would be a good approach.

Closes #228.

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -19,13 +19,19 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import javax.lang.model.SourceVersion;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.element.TypeParameterElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.type.TypeVariable;
 
 import static com.squareup.javapoet.Util.checkArgument;
 import static com.squareup.javapoet.Util.checkNotNull;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Also, two questions:
 * Do we copy method annotations?
 * Do we copy param annotations?

Copying type annotations seems like an obvious yes but we don't support that yet (#136).

swankjesse:
Make sure you don't get two `@Override` annotations if one is already present.

On copying over annotations, we should do what IntelliJ does: copy 'em. Though we need API to remove annotations too.

On testing, can you do what TypesTest does to get Elements for the current file?

JakeWharton:
PTAL.

Copying annotations requires non-trivial infrastructure around being able to create `AnnotationSpec`'s from `AnnotationMirrors` which I'll do in a follow-up.

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -150,6 +156,52 @@ public static Builder constructorBuilder() {
     return new Builder(CONSTRUCTOR);
   }
 
+  /**
+   * Create a builder which overrides the specified {@code method}. This will copy its visibility
```
swankjesse: nit: drop _the specified_. Using code font is good enough!

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -150,6 +156,52 @@ public static Builder constructorBuilder() {
...
+    if (modifiers.contains(Modifier.PRIVATE)
+        || modifiers.contains(Modifier.FINAL)
+        || modifiers.contains(Modifier.STATIC)) {
+      throw new IllegalArgumentException("cannot override method with modifiers: " + modifiers);
+    }
+
+    String methodName = method.getSimpleName().toString();
+    MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(methodName);
+
+    // TODO copy method annotations.
```
swankjesse: add an `@Ignore` failing test for these?

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -150,6 +156,52 @@ public static Builder constructorBuilder() {
...
+    // TODO check to ensure we're not duplicating override annotation.
+    methodBuilder.addAnnotation(Override.class);
+
+    modifiers = new LinkedHashSet<>(modifiers); // Local copy so we can remove.
+    modifiers.remove(Modifier.ABSTRACT);
+    methodBuilder.addModifiers(modifiers.toArray(new Modifier[modifiers.size()]));
```
swankjesse: We probably want an overload of `addModifiers()` that takes a `Collection` !

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -150,6 +156,52 @@ public static Builder constructorBuilder() {
...
+
+    for (TypeParameterElement typeParameterElement : method.getTypeParameters()) {
```
swankjesse: Off topic: I can't think of an overriding method with type parameters.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -15,20 +15,39 @@
...
+  }
 
   @Test public void nullAnnotationsAddition() {
     try {
       MethodSpec.methodBuilder("doSomething").addAnnotations(null);
       fail();
     } catch (IllegalArgumentException expected) {
-      assertThat(expected.getMessage())
-          .isEqualTo("annotationSpecs == null");
+      assertThat(expected).hasMessage("annotationSpecs == null");
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -15,20 +15,39 @@
...
+import java.util.List;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.TypeElement;
+import org.junit.Rule;
 import org.junit.Test;
 
+import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.common.truth.Truth.assertThat;
+import static javax.lang.model.util.ElementFilter.methodsIn;
```
swankjesse: neat.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -57,8 +74,55 @@
...
+
+  @Test public void overrideEverything() {
+    TypeElement classElement = getElement(Everything.class);
+    ExecutableElement methodElement = getOnlyElement(methodsIn(classElement.getEnclosedElements()));
+
+    MethodSpec method = MethodSpec.overriding(methodElement).build();
+    assertThat(method.toString()).isEqualTo(""
+        + "@java.lang.Override\n"
+        + "protected <T extends java.lang.Runnable & java.io.Closeable> "
+        + "java.lang.Runnable everything(java.lang.String arg0, java.util.List<? extends T> arg1) "
```
swankjesse: arg0, arg1 is very sad!

swankjesse:
LGTM

src/test/java/com/squareup/javapoet/MethodSpecTest.java
JakeWharton: ew, yeah. wtf. that shouldn't happen.

src/main/java/com/squareup/javapoet/MethodSpec.java
JakeWharton: ha, yeah. not sure why I didn't just add one. will do.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
JakeWharton: I'm pretty sure this is just due to our use of compile-testing which references back into the elements of the test code (cc/ @gk5885). In an annotation processor, you'd have the real names.

src/main/java/com/squareup/javapoet/MethodSpec.java
JakeWharton: done

PR #233, opened 2015-02-24
Creates annotation type from TypeSpec.

### LINKED ISSUE #232: Create an annotation type
I want to generate a new java annotation type, something like :

```
TypeSpec.annotationBuilder("MyAnnotation")
            .addMethod(
                        MethodSpec.methodBuilder("test")
                        .defaultValue("0")
                        .returns(int.class).
            );
```

It should generate :
```
public @interface MyAnnotation {
    public int test() default 0;
}
```

GerardPaligot:
Adds a new Kind in TypeSpec to represent an annotation and adds an initializer in MethodSpec used to initialize methods of annotation type.

Closes #232

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -51,6 +51,7 @@ public final class MethodSpec {
   public final boolean varargs;
   public final List<TypeName> exceptions;
   public final CodeBlock code;
+  public final CodeBlock defaultValue;
 
   private MethodSpec(Builder builder) {
     CodeBlock code = builder.code.build();
@@ -68,6 +69,7 @@ public final class MethodSpec {
     this.parameters = Util.immutableList(builder.parameters);
     this.varargs = builder.varargs;
     this.exceptions = Util.immutableList(builder.exceptions);
+    this.defaultValue = builder.defaultValue;
     this.code = code;
   }
 
@@ -102,6 +104,12 @@ public final class MethodSpec {
     }
 
     codeWriter.emit(")");
+
+    if (defaultValue != null && !defaultValue.isEmpty()) {
+      codeWriter.emit(" default ");
+      codeWriter.emit(defaultValue);
+    }
+
     if (!exceptions.isEmpty()) {
       codeWriter.emit(" throws");
       boolean firstException = true;
@@ -214,6 +222,7 @@ public final class MethodSpec {
     private final List<TypeName> exceptions = new ArrayList<>();
     private final CodeBlock.Builder code = CodeBlock.builder();
     private boolean varargs;
+    private CodeBlock defaultValue;
 
     private Builder(String name) {
       checkArgument(name.equals(CONSTRUCTOR) || SourceVersion.isName(name),
@@ -325,6 +334,16 @@ public final class MethodSpec {
 
     public Builder addCode(CodeBlock codeBlock) {
       code.add(codeBlock);
+      return this;
+    }
+
+    public Builder defaultValue(String format, Object... args) {
+      return defaultValue(CodeBlock.builder().add(format, args).build());
+    }
+
+    public Builder defaultValue(CodeBlock codeBlock) {
+      checkState(this.defaultValue == null, "defaultValue was already set");
+      this.defaultValue = checkNotNull(codeBlock, "codeBlock == null");
       return this;
     }
 
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -98,6 +98,10 @@ public final class TypeSpec {
         .build());
   }
 
+  public static Builder annotationBuilder(String name) {
+    return new Builder(Kind.ANNOTATION, checkNotNull(name, "name == null"), null);
+  }
+
   void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifiers)
       throws IOException {
     // Nested classes interrupt wrapped line indentation. Stash the current wrapping state and put
@@ -126,7 +130,11 @@ public final class TypeSpec {
         codeWriter.emitJavadoc(javadoc);
         codeWriter.emitAnnotations(annotations, false);
         codeWriter.emitModifiers(modifiers, Util.union(implicitModifiers, kind.asMemberModifiers));
-        codeWriter.emit("$L $L", kind.name().toLowerCase(Locale.US), name);
+        if (kind == Kind.ANNOTATION) {
+          codeWriter.emit("$L $L", "@interface", name);
+        } else {
+          codeWriter.emit("$L $L", kind.name().toLowerCase(Locale.US), name);
+        }
         codeWriter.emitTypeVariables(typeVariables);
 
         List<TypeName> extendsTypes;
@@ -262,7 +270,13 @@ public final class TypeSpec {
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
-        Collections.singleton(Modifier.STATIC));
+        Collections.singleton(Modifier.STATIC)),
+
+    ANNOTATION(
+        Util.immutableSet(Arrays.asList(Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)),
+        Util.immutableSet(Arrays.asList(Modifier.PUBLIC, Modifier.ABSTRACT)),
+        Util.immutableSet(Arrays.asList(Modifier.PUBLIC, Modifier.STATIC)),
+        Util.immutableSet(Arrays.asList(Modifier.STATIC)));
 
     private final Set<Modifier> implicitFieldModifiers;
     private final Set<Modifier> implicitMethodModifiers;
@@ -424,6 +438,9 @@ public final class TypeSpec {
       checkArgument(methodSpec.modifiers.containsAll(kind.implicitMethodModifiers),
           "%s %s.%s requires modifiers %s", kind, name, methodSpec.name,
           kind.implicitMethodModifiers);
+      if (kind != Kind.ANNOTATION) {
+        checkState(methodSpec.defaultValue == null, "default method != null");
+      }
       methodSpecs.add(methodSpec);
       return this;
     }
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -650,6 +650,59 @@ public final class TypeSpecTest {
         + "}\n");
   }
 
+  @Test public void annotation() throws Exception {
+    TypeSpec annotation = TypeSpec.annotationBuilder("MyAnnotation")
+        .addModifiers(Modifier.PUBLIC)
+        .addMethod(MethodSpec.methodBuilder("test")
+                .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                .defaultValue("$L", 0)
+                .returns(int.class)
+                .build())
+        .build();
+
+    assertThat(toString(annotation)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "public @interface MyAnnotation {\n"
+        + "  int test() default 0;\n"
+        + "}\n"
+    );
+  }
+
+  @Test public void innerAnnotationInAnnotationDeclaration() throws Exception {
+    TypeSpec bar = TypeSpec.annotationBuilder("Bar")
+        .addMethod(MethodSpec.methodBuilder("value")
+            .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+            .defaultValue("@$T", Deprecated.class)
+            .returns(Deprecated.class)
+            .build())
+        .build();
+
+    assertThat(toString(bar)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import java.lang.Deprecated;\n"
+        + "\n"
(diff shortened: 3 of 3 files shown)


src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -328,6 +339,16 @@ public Builder addCode(CodeBlock codeBlock) {
       return this;
     }
 
+    public Builder initializer(String format, Object... args) {
```
swankjesse: rename to defaultValue

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -328,6 +339,16 @@ public Builder addCode(CodeBlock codeBlock) {
...
+      return initializer(CodeBlock.builder().add(format, args).build());
+    }
+
+    public Builder initializer(CodeBlock codeBlock) {
```
swankjesse: rename to defaultValue

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -51,6 +51,7 @@
   public final boolean varargs;
   public final List<TypeName> exceptions;
   public final CodeBlock code;
+  public final CodeBlock initializer;
```
swankjesse: rename to defaultValue

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -68,6 +69,9 @@ private MethodSpec(Builder builder) {
     this.parameters = Util.immutableList(builder.parameters);
     this.varargs = builder.varargs;
     this.exceptions = Util.immutableList(builder.exceptions);
+    this.initializer = (builder.initializer == null)
```
swankjesse: please leave this as `null` if there's no initializer. (We should do similarly for field initializers also; I think it's a mistake I made to treat an empty code block as the same as no code block!)

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -214,10 +224,11 @@ public static Builder overriding(ExecutableElement method) {
     private final List<TypeName> exceptions = new ArrayList<>();
     private final CodeBlock.Builder code = CodeBlock.builder();
     private boolean varargs;
+    private CodeBlock initializer;
 
     private Builder(String name) {
       checkArgument(name.equals(CONSTRUCTOR) || SourceVersion.isName(name),
-          "not a valid name: %s", name);
+              "not a valid name: %s", name);
```
swankjesse: nit: we're always +4 on wrapped lines

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -98,6 +98,10 @@ public static Builder anonymousClassBuilder(String typeArgumentsFormat, Object..
         .build());
   }
 
+  public static Builder annotationBuilder(String name) {
+    return new Builder(Kind.ANNOTATION, checkNotNull(name, "name == null"), null);
```
swankjesse: nice

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -126,7 +130,11 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         codeWriter.emitJavadoc(javadoc);
         codeWriter.emitAnnotations(annotations, false);
         codeWriter.emitModifiers(modifiers, Util.union(implicitModifiers, kind.asMemberModifiers));
-        codeWriter.emit("$L $L", kind.name().toLowerCase(Locale.US), name);
+        if (isAnnotationDeclaration()) {
```
swankjesse: Wanna just add a field on `kind` like `keyword` that has values `"class"`, `"interface"`, `"enum"` and `"@interface"` ? Then we don't need this special case here. The name().toLowerCase() trick is gross when it doesn't work universally!

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -234,6 +242,10 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
     }
   }
 
+  public boolean isAnnotationDeclaration() {
```
swankjesse: move this to be under `hasModifier` ? Might as well put these methods all together.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -262,7 +274,13 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
-        Collections.singleton(Modifier.STATIC));
+        Collections.singleton(Modifier.STATIC)),
+
+    ANNOTATION(
```
swankjesse: I think it's a bug in our code that `Kind` is a private interface, since we expose it in the public `Kind` field. Do you mind making `Kind` public, and then we don't need `isAnnotationDeclaration()` ? The caller can just check the kind!

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -98,6 +98,10 @@ public static Builder anonymousClassBuilder(String typeArgumentsFormat, Object..
         .build());
   }
 
+  public static Builder annotationBuilder(String name) {
```
swankjesse: Can you add a checkState() in `addMethod` to make sure that we reject methods with default values unless the current typespec is an annotation? It's an easy check to add and makes the code a little easier to reason about.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -262,7 +274,13 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
         Collections.<Modifier>emptySet(),
-        Collections.singleton(Modifier.STATIC));
+        Collections.singleton(Modifier.STATIC)),
+
+    ANNOTATION(
+        Collections.<Modifier>emptySet(),
```
swankjesse: This should be the same as for regular interfaces. Annotation fields work just like interface fields with respect to default modifiers.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -262,7 +274,13 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
...
+        Util.immutableSet(Arrays.asList(Modifier.PUBLIC, Modifier.ABSTRACT)),
+        Collections.<Modifier>emptySet(),
```
swankjesse: should be same as for interfaces

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -262,7 +274,13 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
...
+        Collections.<Modifier>emptySet());
```
swankjesse: should be same as for interfaces

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -650,6 +650,25 @@ private TypeElement getElement(Class<?> clazz) {
         + "}\n");
   }
 
+  @Test public void annotation() throws Exception {
```
swankjesse: Nice!

swankjesse:
Really Awesome. 

I've given you a lot of feedback, much of it is me asking you to fix problems with the code that were there before you got here. If you'd rather not, let me know & I'll merge this and do my own follow up!

Also I need you to complete Square's CLA so we can redistribute your work.
https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ

src/main/java/com/squareup/javapoet/MethodSpec.java
GerardPaligot: I wanted to remain consistent with your initializer in FieldSpec but no problem, I can modify it!

src/main/java/com/squareup/javapoet/TypeSpec.java
GerardPaligot: I wanted use `kind.name().toLowerCase()` too but we can't have a field which begins with an '@' in an enumeration.

src/main/java/com/squareup/javapoet/TypeSpec.java
GerardPaligot: The responsibility of `hasModifier` isn't the same. `hasModifier` checks `modifiers` field of `TypeSpec`. `isAnnotationDeclaration` checks `kind` field of `TypeSpec`. But we can have a new method like :

```java
public boolean isKind(Kind kind) {
  return this.kind == kind;
}
```

What do you think of it?

src/main/java/com/squareup/javapoet/TypeSpec.java
GerardPaligot: I think it isn't necessary to expose `Kind` in public. `Kind` and `isAnnotationDeclaration()` are both in `TypeSpec`.

GerardPaligot:
Thanks for your feedback. I'm waiting your answers for some questions in response of yours and I'll push my branch.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -650,6 +650,25 @@ private TypeElement getElement(Class<?> clazz) {
...
+                .initializer("$L", 0)
+                .returns(int.class)
+                .build())
+        .build();
+
+    assertThat(toString(annotation)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "public @interface MyAnnotation {\n"
+        + "  int test() default 0;\n"
```
tbroyer: To validate the API, could you add a test that generates a default value being an annotation?

```
@interface Foo {
  int value();
}
@interface Bar
  Foo value() default @Foo(0)
}
```

(it might be something we need to test when emitting annotations too: are we able to emit `@Bar(@Foo(42)) int someField`?)

JakeWharton: Both good ideas :+1:

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -80,6 +80,10 @@ public boolean hasModifier(Modifier modifier) {
     return modifiers.contains(modifier);
   }
 
+  public boolean isKind(Kind kind) {
```
swankjesse: I think we can drop this, since `kind` is a public field.

```
if (type.kind == Kind.ANNOTATION)
```

src/main/java/com/squareup/javapoet/TypeSpec.java
swankjesse: Yeah, I'm sorry. I meant making `Kind` the type public:

```
public enum Kind {
```

It's dumb that we have a public field that's member is a private class.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -424,6 +442,9 @@ public Builder addMethod(MethodSpec methodSpec) {
       checkArgument(methodSpec.modifiers.containsAll(kind.implicitMethodModifiers),
           "%s %s.%s requires modifiers %s", kind, name, methodSpec.name,
           kind.implicitMethodModifiers);
+      if (kind != Kind.ANNOTATION) {
+        checkState(methodSpec.defaultValue == null, "default method != null");
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -650,6 +650,59 @@ private TypeElement getElement(Class<?> clazz) {
...
+            .build())
+        .build();
+
+    assertThat(toString(bar)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import java.lang.Deprecated;\n"
+        + "\n"
+        + "@interface Bar {\n"
+        + "  Deprecated value() default @Deprecated;\n"
```
swankjesse: Fun.

src/main/java/com/squareup/javapoet/TypeSpec.java
swankjesse: We might as well just make it public. This API doesn't try very hard to encapsulate things.

swankjesse: (and that's okay)

swankjesse:
Just one small fix for making the `Kind` type public, so we can nuke the corresponding check method. (Right now nobody could call that, because they can't get a reference to the private `Kind` type anyway).

Then I'll merge.

Thanks!

GerardPaligot:
Well, it's done.

swankjesse:
Thank you come again!

src/test/java/com/squareup/javapoet/TypeSpecTest.java
tbroyer: Note: we should use `AnnotationSpec` in this test; `("@$T", Deprecated.class)` is a hack. Now maybe using the `toString` of `AnnotationSpec` would be enough (I must say it's already too long since I looked closely at JavaPoet's code)

PR #242, opened 2015-03-05
Add toBuilder() methods

### LINKED ISSUE #240: Converting Specs to Builders
Right now, you have to recreate the Builder from the spec manually, which is not easy for more complex things like `MethodSpec` and requires static utility functions for maintaining. A simple fix would be to add a new `*Spec.builder()` instance method that converts that Spec instance to the respective builder type.

kenzierocks:
Closes #240

### DIFF
--- a/src/main/java/com/squareup/javapoet/FieldSpec.java
+++ b/src/main/java/com/squareup/javapoet/FieldSpec.java
@@ -89,6 +89,15 @@ public final class FieldSpec {
     return builder(TypeName.get(type), name, modifiers);
   }
 
+  public Builder toBuilder() {
+    Builder builder = new Builder(type, name);
+    builder.javadoc.add(javadoc);
+    builder.annotations.addAll(annotations);
+    builder.modifiers.addAll(modifiers);
+    builder.initializer = initializer.isEmpty() ? null : initializer;
+    return builder;
+  }
+
   public static final class Builder {
     private final TypeName type;
     private final String name;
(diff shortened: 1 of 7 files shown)


cgruber:
Do we really want this annotationSpec.toBuilder() pattern, or do we want AnnotationSpec.builder(annotationSpec);?

kenzierocks:
I'm up for either, although I think the first might be better.

src/main/java/com/squareup/javapoet/FieldSpec.java
```diff
@@ -89,6 +89,15 @@ public static Builder builder(Type type, String name, Modifier... modifiers) {
     return builder(TypeName.get(type), name, modifiers);
   }
 
+  public Builder toBuilder() {
+    Builder builder = new Builder(type, name);
+    builder.javadoc.add(javadoc);
+    builder.annotations.addAll(annotations);
+    builder.modifiers.addAll(modifiers);
+    builder.initializer = initializer;
```
swankjesse: nit: check if this is empty, and if it is pass `null`? Otherwise you can't add an initializer on the copied instance. (I'm looking at the constructor and trying to figure out what the opposite assignment is!)

swankjesse:
Code all looks good. Thanks! I'm unsure about whether it should be `toBuilder()`, `buildUpon()`, or a static `builder(Foo)` class. For now, let's take this as-is and if we rename it later, that's cool.

Can you fix the one field bug and rebase & squash? Then I'll merge this.

kenzierocks:
Alright, done.

cgruber:
So it's barn-door-post-horse time, but I think this P/R would have been the
best time to decide between an instance method copy method, vs. a static
builder built around a copy, since it's not a rename, but a more
substantial refactor of flow to change later.  /shrug. That said, it's
done, so... er... yeah. :)

PR #243, opened 2015-03-06
Rename toBuilder() to builder().

swankjesse:
I had a small insight: the only place where the static builder()
method conflicted with the non-static builder() method was in
CodeBlock. That class is composable: it's just as easy to create
a new code block, and call add(), as it is to convert an existing
code block into a builder. So we don't need the toBuilder() method
there anyway.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -99,7 +99,7 @@ public final class AnnotationSpec {
     return builder(ClassName.get(type));
   }
 
-  public Builder toBuilder() {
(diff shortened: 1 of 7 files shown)


swankjesse:
FYI @kenzierocks

kenzierocks:
Yea, that works. Good thinking!

eamonnmcmanus:
I think this kind of goes against convention. The convention I've seen most often is that builders are made with a static builder() method. The method to convert an object back to a builder is called toBuilder(). I admit to bias here, because that's the convention we endorse with AutoValue, though we don't require it.

swankjesse:
I like being consistent with AutoValue. I'll withdraw this.

PR #244, opened 2015-03-07
Add support for Java 8 modifier changes

kenzierocks:
Adds support for most Java 8 modifier changes.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -22,12 +22,14 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
@@ -35,6 +37,8 @@ import javax.lang.model.element.Modifier;
 import static com.squareup.javapoet.Util.checkArgument;
 import static com.squareup.javapoet.Util.checkNotNull;
 import static com.squareup.javapoet.Util.checkState;
+import static com.squareup.javapoet.Util.hasDefaultModifier;
+import static com.squareup.javapoet.Util.requireExactlyOneOf;
 
 /** A generated class, interface, or enum declaration. */
 public final class TypeSpec {
@@ -429,9 +433,14 @@ public final class TypeSpec {
     }
 
     public Builder addField(FieldSpec fieldSpec) {
-      checkArgument(fieldSpec.modifiers.containsAll(kind.implicitFieldModifiers),
-          "%s %s.%s requires modifiers %s", kind, name, fieldSpec.name,
-          kind.implicitFieldModifiers);
+      checkState(kind != Kind.ANNOTATION, "%s %s cannot have fields", kind, name);
+      if (kind == Kind.INTERFACE) {
+        requireExactlyOneOf(EnumSet.of(Modifier.PUBLIC, Modifier.PRIVATE),
+            fieldSpec.modifiers);
+        Set<Modifier> check = EnumSet.of(Modifier.STATIC, Modifier.FINAL);
+        checkState(fieldSpec.modifiers.containsAll(check), "%s %s.%s requires modifiers %s",
+            kind, name, fieldSpec.name, check);
+      }
       fieldSpecs.add(fieldSpec);
       return this;
     }
@@ -453,11 +462,25 @@ public final class TypeSpec {
     }
 
     public Builder addMethod(MethodSpec methodSpec) {
-      checkArgument(methodSpec.modifiers.containsAll(kind.implicitMethodModifiers),
-          "%s %s.%s requires modifiers %s", kind, name, methodSpec.name,
-          kind.implicitMethodModifiers);
+      if (kind == Kind.INTERFACE) {
+        Set<Modifier> mods = EnumSet.of(Modifier.ABSTRACT, Modifier.STATIC);
+        if (Util.DEFAULT != null) {
+          mods.add(Util.DEFAULT);
+        }
+        requireExactlyOneOf(mods, methodSpec.modifiers);
+        requireExactlyOneOf(EnumSet.of(Modifier.PUBLIC, Modifier.PRIVATE),
+            methodSpec.modifiers);
+      } else if (kind == Kind.ANNOTATION) {
+        checkState(methodSpec.modifiers.containsAll(kind.implicitMethodModifiers),
+            "%s %s.%s cannot have modifiers", kind, name, methodSpec.name);
+      }
       if (kind != Kind.ANNOTATION) {
-        checkState(methodSpec.defaultValue == null, "default method != null");
+        checkState(methodSpec.defaultValue == null, "%s %s.%s cannot have a default value",
+            kind, name, methodSpec.name);
+      }
+      if (kind != Kind.INTERFACE) {
+        checkState(!hasDefaultModifier(methodSpec.modifiers), "%s %s.%s cannot be default",
+            kind, name, methodSpec.name);
       }
       methodSpecs.add(methodSpec);
       return this;
@@ -474,7 +497,7 @@ public final class TypeSpec {
     public Builder addType(TypeSpec typeSpec) {
       checkArgument(typeSpec.modifiers.containsAll(kind.implicitTypeModifiers),
           "%s %s.%s requires modifiers %s", kind, name, typeSpec.name,
-          kind.implicitFieldModifiers);
+          kind.implicitTypeModifiers);
       typeSpecs.add(typeSpec);
       return this;
     }
--- a/src/main/java/com/squareup/javapoet/Util.java
+++ b/src/main/java/com/squareup/javapoet/Util.java
@@ -24,12 +24,24 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.lang.model.element.Modifier;
+
 /**
  * Like Guava, but worse and standalone. This makes it easier to mix JavaPoet with libraries that
  * bring their own version of Guava.
  */
 final class Util {
   private Util() {
+  }
+
+  public static final Modifier DEFAULT;
+  static {
+    Modifier def = null;
+    try {
+      def = Modifier.valueOf("DEFAULT");
+    } catch (IllegalArgumentException ignored) {
+    }
+    DEFAULT = def;
   }
 
   public static <K, V> Map<K, List<V>> immutableMultimap(Map<K, List<V>> multimap) {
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -82,4 +84,13 @@ public static String join(String separator, List<String> parts) {
     result.addAll(b);
     return result;
   }
+
+  public static boolean hasDefaultModifier(Collection<Modifier> modifiers) {
```
JakeWharton: Should not be public

JakeWharton: Actually I guess doesn't matter, this class isn't public

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -55,7 +56,11 @@
 
   private MethodSpec(Builder builder) {
     CodeBlock code = builder.code.build();
-    checkArgument(code.isEmpty() || !builder.modifiers.contains(Modifier.ABSTRACT),
+    checkArgument(!(builder.modifiers.contains(Modifier.ABSTRACT)
+        && hasDefaultModifier(builder.modifiers)),
```
swankjesse: Can be just `builder.modifiers.contains(Modifier.DEFAULT)`

swankjesse:
For tests, copy a method from TypeSpecTest and modify it until it does what you need! We want to accept `default` on interface methods, but not elsewhere.

src/main/java/com/squareup/javapoet/MethodSpec.java
JakeWharton: That won't compile on JDK 7, unfortunately

swankjesse: Ahhh, got it. Yup.

kenzierocks:
Looks like I wrote the wrong checkState, I'll add tests and attempt to fix that.

kenzierocks:
@swankjesse Is there any preferred order for test methods, or can I insert anywhere?

swankjesse:
@kenzierocks related tests are usually together, but the order isn't too important. Probably best to put these with the other tests that check for modifiers.

kenzierocks:
And for the Java 8 compile restriction?

swankjesse:
One interesting/bad consequence of the `DEFAULT` constant is that testing this will cause our tests to require Java 8. @cgruber is this a problem?

JakeWharton:
Hmm... what about doing an `assumeTrue(isJava8());` and then `Modifer.valueOf("DEFAULT")`?

kenzierocks:
Can do.

kenzierocks:
Alright, that should be it.

kenzierocks:
Okay, found a new problem. `default` methods aren't `abstract`, so that's no longer implicit in interfaces because it would cause a [`checkArgument`](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/TypeSpec.java#L456) to fail. However, when there's no `default`, `abstract` is implicit, but there's no clear way to indicate this. Is there a preferred way that this should be handled?

swankjesse:
@kenzierocks I think you should require either `default` or `abstract`, but not both.

kenzierocks:
Yes, but then it emits an `abstract` modifier in cases where it's implicit. I could change that `checkArgument` I linked above.

JakeWharton:
Don't forget `static`. Also, while we're in here, we should support `private` methods on interfaces as well (Java 9).

kenzierocks:
Oh, that sounds cool. I think I'll rework the implicit structure a bit to ease this in the future in case even more implicitness is added/removed.

kenzierocks:
I changed it so that implicit modifiers are not required, which is the most minimalist change I can think of that won't break compatibility but will allow this new rule. A side effect of this is allowing the above changes to also work. I just need to add a test that checks for `static` on interfaces working.

swankjesse:
@kenzierocks I'd like to continue requiring implicit modifiers. When Java 9 comes and interface methods can be public or private, I want the public to be in the callsite so we don't have to guess!

kenzierocks:
Then I'll need to think of some way to allow the implicit modifiers through, or scrap the whole system and just always print modifiers explicitly.

kenzierocks:
Updated the `hasDefaultModifier` method to cache the value of `DEFAULT`.

kenzierocks:
I'm not finding the specification for `private` on interface methods, would you mind showing me?

JakeWharton:
https://bugs.openjdk.java.net/browse/JDK-8071453

kenzierocks:
Okay, tell me what y'all think of the new system.

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -55,7 +56,11 @@
...
+        "abstract method %s cannot be default", builder.name);
+    checkArgument(code.isEmpty() || !builder.modifiers.contains(Modifier.ABSTRACT)
+        || hasDefaultModifier(builder.modifiers),
```
swankjesse: Oh wow, is `abstract default` allowed?

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -32,6 +38,72 @@
...
   }
 
+  private static final Modifier DEFAULT;
+  /**
+   * Used to detect conflicting modifiers, maps from a modifier to the
+   * conflicting modifier list.
+   */
+  private static final Map<Modifier, Set<Modifier>> CONFLICTING_MODIFIER_MAP;
+  static {
+    Modifier def = null;
```
swankjesse: neat

src/main/java/com/squareup/javapoet/MethodSpec.java
kenzierocks: No, see checkArg above. Couldn't get the whole thing in a boolean statement, if someone has better logic than me feel free to offer up one.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -32,6 +38,72 @@
...
+    try {
+      def = Modifier.valueOf("DEFAULT");
+    } catch (IllegalArgumentException expected) {
```
swankjesse: nit: expected –> ignored 

(expected: this should happen 100% of the time. ignored: I don't care if this happens)

kenzierocks: Right.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -32,6 +38,72 @@
...
+    temp.putAll(generateCombinations(Modifier.ABSTRACT, Modifier.NATIVE));
+    temp.putAll(generateCombinations(Modifier.ABSTRACT, Modifier.SYNCHRONIZED));
+    temp.putAll(generateCombinations(Modifier.NATIVE, Modifier.STRICTFP));
+
+    EnumMap<Modifier, Set<Modifier>> result = new EnumMap<>(Modifier.class);
+    for (Map.Entry<Modifier, Set<Modifier>> entry : temp.entrySet()) {
+      if (entry.getValue().isEmpty()) continue;
+      result.put(entry.getKey(), immutableSet(entry.getValue()));
+    }
+    CONFLICTING_MODIFIER_MAP = Collections.unmodifiableMap(result);
```
swankjesse: I think this code is actually harder to follow than just statically listing them all. (At the risk that you're asymmetric, but I don't think that'll be a problem in practice.)

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -32,6 +38,72 @@
...
+  }
+
+  private static Map<Modifier, Set<Modifier>> generateCombinations(Modifier... source) {
+    Set<Modifier> sourceNonNull = new HashSet<>(Arrays.asList(source));
```
swankjesse: nit: never, never, never use HashSet. Always use LinkedHashSet. The extra determinism is worth it.

src/main/java/com/squareup/javapoet/Util.java
kenzierocks: That might be. I'll use JavaPoet to write out the static list :)

src/main/java/com/squareup/javapoet/Util.java
kenzierocks: I always use `Hash*`, will fix.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -82,4 +154,23 @@ public static String join(String separator, List<String> parts) {
     result.addAll(b);
     return result;
   }
+
+  public static Set<Modifier> getMissingImplicitModifiers(Set<Modifier> modifiers,
```
swankjesse: Hmmmm I don't know if I follow this. It's kind of hard to reason about what's happening here, with a whole bunch of different sets. Wouldn't we be better off at the call sites to have a method like this:

```
  if (kind == INTERFACE) {
    requireSubsetOf(modifiers, Modifier.PUBLIC, Modifier.ABSTRACT, Util.DEFAULT);
    requireExactlyOne(modifiers, Modifier.ABSTRACT, Util.DEFAULT);
  }
```

As is, it's actually awkward to recognize what the rules are.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -54,6 +55,13 @@ private TypeElement getElement(Class<?> clazz) {
     return compilation.getElements().getTypeElement(clazz.getCanonicalName());
   }
 
+  private boolean isJava8() {
+    String version = System.getProperty("java.version");
+    int pos = version.indexOf('.');
+    pos = version.indexOf('.', pos+1);
+    return Double.valueOf(version.substring (0, pos)).compareTo(Double.valueOf(1.8)) >= 0;
```
swankjesse: I think you can probably just do string comparison. The right thing will happen:
```
  return System.getProperty("java.version").compareTo("1.8") >= 0;
```

src/main/java/com/squareup/javapoet/Util.java
kenzierocks: Yes I can do that.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
swankjesse: (any situation where this fails, it's also going to throw when you call `Double.valueOf()`)

kenzierocks: What if they decide to do `1.10`? I think that leaves us with the wrong answer.

swankjesse: Good point! Similarly, do we run into problems like 1.9-beta?

swankjesse: The way I prefer to do this sort of thing is lame, it's just:

```
try {
  Modifier.valueOf("DEFAULT")
  return true;
} catch (Exception e) {
  return false;
}
```
That way you're testing for the feature you want, not the version it arrived in. This also works very nicely for Android.

src/main/java/com/squareup/javapoet/Util.java
swankjesse: Clever!

src/main/java/com/squareup/javapoet/Util.java
JakeWharton: :+1: Was going to say something similar (and if you look at the mercurial commit where private methods were added you see they do similar things).

swankjesse:
I love how this is coming together. Thanks for your patience on my strict and pedantic reviews. I think it's worth it.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
kenzierocks: Yes, I think that would work much better.

kenzierocks:
Oh, I don't mind. The only pain I have is my formatter is off by a little bit so I'm trying not to hit format.

src/main/java/com/squareup/javapoet/Util.java
kenzierocks: Can you explain what `requireSubsetOf` should be doing? It's not exactly clear from the name, as that would contain some invalid combos as currently specified.

swankjesse:
```
Set<Modifier> s = new LinkedHashSet<Modifer>(PUBLIC, STATIC);
requireSubsetOf(s, PUBLIC, STATIC, FINAL); // Good, s is a subset of the varargs
requireSubsetOf(s, STATIC, FINAL); // Bad, s is not a subset of the varargs
```

kenzierocks: But the subsets also contain the empty set...should I ignore the empty set?

swankjesse: yeah, that's fine. The `requireSubsetOf()` is intended to break if an unsupported modifier is used. The `requireExactlyOne` call makes sure a necessary modifier is present.

kenzierocks:
Okay, this should be good.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -429,9 +433,16 @@ public Builder addFields(Collection<FieldSpec> fieldSpecs) {
...
 
     public Builder addField(FieldSpec fieldSpec) {
-      checkArgument(fieldSpec.modifiers.containsAll(kind.implicitFieldModifiers),
-          "%s %s.%s requires modifiers %s", kind, name, fieldSpec.name,
-          kind.implicitFieldModifiers);
+      checkState(kind != Kind.ANNOTATION, "%s %s cannot have fields", kind, name);
+      if (kind == Kind.INTERFACE) {
+        requireExactlyOneOf(EnumSet.of(Modifier.PUBLIC, Modifier.PRIVATE),
+            fieldSpec.modifiers);
+        if (kind == Kind.INTERFACE) {
```
JakeWharton: This `if` can be removed. You already have it above.

kenzierocks: That was leftover from when the above `if` had `|| kind == Kind.ANNOTATION`. Fixing.

kenzierocks:
Fixed problem, let me know if there's anything else I need to modify.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -429,9 +433,14 @@ public Builder addFields(Collection<FieldSpec> fieldSpecs) {
     }
 
     public Builder addField(FieldSpec fieldSpec) {
-      checkArgument(fieldSpec.modifiers.containsAll(kind.implicitFieldModifiers),
-          "%s %s.%s requires modifiers %s", kind, name, fieldSpec.name,
-          kind.implicitFieldModifiers);
+      checkState(kind != Kind.ANNOTATION, "%s %s cannot have fields", kind, name);
```
swankjesse: I love this

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -82,4 +94,63 @@ public static String join(String separator, List<String> parts) {
...
+  /**
+   * Finds a subset of {@code set} in {@code modifiers}. Ignores empty subsets,
+   * and if {@code set} is empty then {@code true} is always returned.
+   *
+   * @param modifiers
+   *          - The input modifiers
+   * @param set
+   *          - The set of modifiers to make subsets of
+   */
+  public static void requireSubsetOf(Set<Modifier> set, Set<Modifier> modifiers) {
```
swankjesse: Unused.

swankjesse: (Though neat_

swankjesse:
This works! I'm going to merge it as-is and send my own follow-up to you where I think we can slightly shrink the code required. (Fun java tricks!)

swankjesse:
Thanks for the fix, and for your patience!

swankjesse:
https://github.com/square/javapoet/pull/251

PR #257, opened 2015-03-25
Added support for arbitrary blocks of code (like static {}) in the TypeSpec

DanielSerdyukov:


confile:
+1

swankjesse:
(Sorry I missed this PR.)

confile:
@swankjesse It would be great if you could merge this request. Thanks!

swankjesse:
Much sorry for sitting on this for nearly forever. Now that I'm finally back, you have my attention.

I think we want to implement these as `<clinit>` MethodSpec instances, instead of as code blocks that float within the type spec. The problem with code blocks is that structurally they're a lie: there's no such thing as a code block immediately within a type. Instead it's a method that the compiler calls `<clinit>`.

Since this is vague & involves some hidden details of our API, I'm happy to do the work to make this work. But if you'd like to venture forward and bring this over the finish line, I invite you to. You started this, after all. But I want the design to be static class initializer methods inside of TypeSpecs, not code blocks inside of TypeSpecs.

tbroyer:
@swankjesse There's not only [static initializers](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-StaticInitializer), but also [instance initializers](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-InstanceInitializer) (sometimes used in so-called [double-brace initialization](http://c2.com/cgi/wiki?DoubleBraceInitialization)) Would you use `<init>` for an instance initializer? (it's already used to denote constructors, so we have a problem)
Note: the great thing with instance initializers is that they're automatically [copied by the compiler into each constructor](https://docs.oracle.com/javase/tutorial/java/javaOO/initial.html) so you don't have to repeat yourself.

There can also be several static and instance initializer blocks per class declaration, and they can be interspersed with field declarations, and order matters! I could live with the limitation that you just cannot emit those kind of things with JavaPoet and need to move your field initializers to initializer blocks to avoid ordering issues (given that JavaPoet emits fields before methods). Or maybe initializer blocks should be implemented as `FieldSpec` rather than `MethodSpec`? (using the same `<init>` and `<clinit>` special names)

E.g.

```java
static final ImmutableList<String> FIRST;
static {
   ImmutableList.Builder<String> builder = ImmutableList.builder();
   for (int i = 0; i < Integer.getInteger("someSystemProperty", 0); i++) {
     builder.add("i" + i);
   }
   FIRST = builder.build();
}
static final int SECOND = FIRST.size() / Double.getDouble("someOtherSystemProperty");
```

This is obviously not something you'd like to do in practice, just a contrived example where order matters. If the `FIRST` field were a mutable list, it could even be initialized at field declaration, then populated in the static initializer and the size used in the initialization of the `SECOND` field; moving the static initializer after the fields wouldn't lead to a compilation error, but the `SECOND` field value would be wrong (`FIRST.size()` always being `0`, as the static initializer would run afterwards)

swankjesse:
@tbroyer good points. Let's leave instance initializers out of JavaPoet. For static initializers, we'll allow just one, and declare it between static fields and instance fields?

tbroyer:
Works for me. Instance initializers are rare enough that it shouldn't hurt too much (and it's easy to duplicate the code into each constructor with `CodeBlock`), and static initializers could be composed similarly from several `CodeBlock` if needed. Those limitations should be clearly documented though.

dev-troy:
@swankjesse I have corrected the code according to the comments

swankjesse:
LGTM

Two last tasks:

 * http://rebaseandsqua.sh the change
 * complete Square's [contributors agreement](https://docs.google.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ)

DanielSerdyukov:
@swankjesse Done!

swankjesse:
CLA is good. I'm going to squash manually.

PR #259, opened 2015-03-31
Fix import error.

jbeder:
When a type in the current package has the same name as a type in another package, javapoet should not attempt to import the other package's type; rather, it should just refer to it by its fully qualified name.


swankjesse:
Yikes. Good find. One implementation question: would it work to add the same-package type to the 
`importedTypes` map, only to skip writing the import statement when it's time to do that? Or does that trigger another imports edge case?

(CLA is good.)

jbeder:
I think that'd be tricky because you might see the classes in the other order. E.g., if you were generating package "foo" and you had types foo.X and bar.X, if you saw bar.X first, it would mark bar.X as an importableType; and then later you saw foo.X, so you'd have to "unmark" it.

Something like that probably could work, but this solution seemed simpler.

swankjesse:
Works for me. Thanks for the fix!

PR #262, opened 2015-04-08
Fix Util error message

kenzierocks:
Array printing without `Arrays.toString` isn't pretty.


swankjesse:
Checkstyle is unhappy. Switch your tabs to spaces and squash?

kenzierocks:
My new Eclipse install isn't fully setup, must not have changed that setting...

kenzierocks:
Alright, that should be good.

PR #268, opened 2015-05-04
AnnotationSpec.get and MethodSpec.overriding

sormuras:
Added static get method to AnnotationSpec using an AnnotationMirror as source.

Next, I implemented some of the TODOs in MethodSpec.overriding - and changed the signature to include an ExecutableType. That allows the actual generic type parameters retrieval for the return type and the signature.
See the interface extending parameterized interfaces test case: ExtendsOthers in MethodSpecTest.

PS: This time, checkstyle and unit tests are green: https://travis-ci.org/sormuras/javapoet

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -25,6 +25,14 @@ import java.util.Map;
 
 import static com.squareup.javapoet.Util.checkNotNull;
 
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.SimpleAnnotationValueVisitor7;
+
 /** A generated annotation on a declaration. */
 public final class AnnotationSpec {
   public final TypeName type;
@@ -90,6 +98,18 @@ public final class AnnotationSpec {
     codeWriter.emit(whitespace + "}");
   }
 
+  public static AnnotationSpec get(AnnotationMirror annotation) {
+    TypeElement element = (TypeElement) annotation.getAnnotationType().asElement();
+    AnnotationSpec.Builder builder = AnnotationSpec.builder(ClassName.get(element));
+    Visitor visitor = new Visitor(builder);
+    for (ExecutableElement executableElement : annotation.getElementValues().keySet()) {
+      String name = executableElement.getSimpleName().toString();
+      AnnotationValue value = annotation.getElementValues().get(executableElement);
+      value.accept(visitor, new Entry(name, value));
+    }
+    return builder.build();
+  }
+
   public static Builder builder(ClassName type) {
     checkNotNull(type, "type == null");
     return new Builder(type);
@@ -152,4 +172,50 @@ public final class AnnotationSpec {
       return new AnnotationSpec(this);
     }
   }
+
+  /**
+   * Tuple of name and value.
+   */
+  private static class Entry {
+    final String name;
+    final AnnotationValue value;
+
+    Entry(String name, AnnotationValue value) {
+      this.name = name;
+      this.value = value;
+    }
+  }
+
+  /**
+   * Annotation value visitor adding members to the given builder instance.
+   */
+  private static class Visitor extends SimpleAnnotationValueVisitor7<Builder, Entry> {
+    final Builder builder;
+
+    Visitor(Builder builder) {
+      super(builder);
+      this.builder = builder;
+    }
+
+    @Override
+    protected Builder defaultAction(Object o, Entry entry) {
+      return builder.addMember(entry.name, "$L", entry.value);
+    }
+
+    @Override
+    public Builder visitAnnotation(AnnotationMirror a, Entry entry) {
+      return builder.addMember(entry.name, "$L", get(a));
+    }
+
+    @Override
+    public Builder visitEnumConstant(VariableElement c, Entry entry) {
+      return builder.addMember(entry.name, "$T.$L", c.asType(), c.getSimpleName());
+    }
+
+    @Override
+    public Builder visitType(TypeMirror t, Entry entry) {
+      return builder.addMember(entry.name, "$T.class", t);
+    }
+  }
+
 }
--- a/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import javax.lang.model.element.TypeElement;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import com.google.testing.compile.CompilationRule;
+
+public final class AnnotationSpecTest {
+
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface AnnotationA {
+  }
+
+  @Inherited
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface AnnotationB {
+  }
+
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface AnnotationC {
+    String value();
+  }
+
+  static enum Breakfast {
+    WAFFLES, PANCAKES
+  }
+
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface HasDefaultsAnnotation {
+
+    byte a() default 5;
+
+    short b() default 6;
+
+    int c() default 7;
+
+    long d() default 8;
+
+    float e() default 9.0f;
+
+    double f() default 10.0;
+
+    char g() default 'k';
+
+    boolean h() default true;
+
+    Breakfast i() default Breakfast.WAFFLES;
+
+    AnnotationA j() default @AnnotationA();
+
+    String k() default "maple";
+
+    Class<? extends Annotation> l() default AnnotationB.class;
+
+    int[] m() default {1, 2, 3};
+
+    Breakfast[] n() default {Breakfast.WAFFLES, Breakfast.PANCAKES};
+
+    Breakfast o();
+
+    int p();
+
+    AnnotationC q() default @AnnotationC("foo");
+  }
+
+  @HasDefaultsAnnotation(
+      o = Breakfast.PANCAKES,
+      p = 1701,
+      f = 11.1,
+      m = {9, 8, 1},
+      l = Override.class,
+      j = @AnnotationA,
+      q = @AnnotationC("bar"))
+  public class IsAnnotated {
+    // empty
+  }
+
+  @Rule
+  public final CompilationRule compilation = new CompilationRule();
+
+  @Test
+  public void testHasDefaultAnnotation() {
+    String name = IsAnnotated.class.getCanonicalName();
+    TypeElement element = compilation.getElements().getTypeElement(name);
+    AnnotationSpec annotation = AnnotationSpec.get(element.getAnnotationMirrors().get(0));
+    assertThat(annotation.toString()).isEqualTo(
+        "@com.squareup.javapoet.AnnotationSpecTest.HasDefaultsAnnotation("
+            + "o = com.squareup.javapoet.AnnotationSpecTest.Breakfast.PANCAKES"
+            + ", p = 1701"
+            + ", f = 11.1"
+            + ", m = {9, 8, 1}"
+            + ", l = java.lang.Override.class"
+            + ", j = @com.squareup.javapoet.AnnotationSpecTest.AnnotationA"
+            + ", q = @com.squareup.javapoet.AnnotationSpecTest.AnnotationC(\"bar\")"
+            + ")");
+  }
+
+  @Test
+  public void testHasDefaultAnnotationWithImport() {
+    String name = IsAnnotated.class.getCanonicalName();
+    TypeElement element = compilation.getElements().getTypeElement(name);
+    AnnotationSpec annotation = AnnotationSpec.get(element.getAnnotationMirrors().get(0));
+    TypeSpec.Builder typeBuilder = TypeSpec.classBuilder(IsAnnotated.class.getSimpleName());
+    typeBuilder.addAnnotation(annotation);
+    JavaFile file = JavaFile.builder("com.squareup.javapoet", typeBuilder.build()).build();
+    assertThat(file.toString()).isEqualTo(
+        "package com.squareup.javapoet;\n"
+            + "\n"
+            + "import java.lang.Override;\n"
+            + "\n"
+            + "@AnnotationSpecTest.HasDefaultsAnnotation(\n"
+            + "    o = AnnotationSpecTest.Breakfast.PANCAKES,\n"
+            + "    p = 1701,\n"
+            + "    f = 11.1,\n"
+            + "    m = {9, 8, 1},\n"
+            + "    l = Override.class,\n"
+            + "    j = @AnnotationSpecTest.AnnotationA,\n"
+            + "    q = @AnnotationSpecTest.AnnotationC(\"bar\")\n"
+            + ")\n"
+            + "class IsAnnotated {\n"
+            + "}\n"
+    );
+  }
+}


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +104,17 @@ public static Builder builder(Class<?> type) {
     return builder(ClassName.get(type));
   }
 
+  public static AnnotationSpec get(AnnotationMirror annotation) {
+    AnnotationSpec.Builder builder = AnnotationSpec.builder(
+        ClassName.get((TypeElement) annotation.getAnnotationType().asElement()));
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry
+        : annotation.getElementValues().entrySet()) {
+      String name = entry.getKey().getSimpleName().toString();
+      builder.addMember(name, CodeBlock.builder().add(entry.getValue().toString()).build());
```
swankjesse: Does this work for all annotation value types? Would be useful to use Compile-Testing to exercise this.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +104,17 @@ public static Builder builder(Class<?> type) {
     return builder(ClassName.get(type));
   }
 
+  public static AnnotationSpec get(AnnotationMirror annotation) {
```
swankjesse: nit: do you mind moving this above the `builder()` methods? Awkward to separate those from the `Builder` itself

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -182,9 +185,13 @@ public static Builder overriding(ExecutableElement method) {
     String methodName = method.getSimpleName().toString();
     MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(methodName);
 
-    // TODO copy method annotations.
-    // TODO check to ensure we're not duplicating override annotation.
     methodBuilder.addAnnotation(Override.class);
+    for (AnnotationMirror mirror : method.getAnnotationMirrors()) {
+      if (((TypeElement) mirror.getAnnotationType().asElement()).getQualifiedName()
+          .equals(Override.class.getName()))
```
swankjesse: nit: our code style wants braces `{...}` on call multiple-line statements

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -195,18 +202,29 @@ public static Builder overriding(ExecutableElement method) {
...
+         psb.addAnnotation(AnnotationSpec.get(mirror));
+      }
+      methodBuilder.addParameter(psb.build());
     }
 
     for (TypeMirror thrownType : method.getThrownTypes()) {
       methodBuilder.addException(TypeName.get(thrownType));
     }
 
+    if (method.isVarArgs()) {
```
swankjesse: Nice.

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -195,18 +202,29 @@ public static Builder overriding(ExecutableElement method) {
...
-      // TODO copy parameter annotations.
-      methodBuilder.addParameter(TypeName.get(parameter.asType()),
-          parameter.getSimpleName().toString());
+    List<? extends VariableElement> parameters = method.getParameters();
+    List<? extends TypeMirror> types = executableType.getParameterTypes();
+    for (int index = 0; index < parameters.size(); index++) {
+      TypeName type = TypeName.get(types.get(index));
+      String name = parameters.get(index).getSimpleName().toString();
+      ParameterSpec.Builder psb = ParameterSpec.builder(type, name, parameters.get(index)
+          .getModifiers().toArray(new Modifier[0]));
```
swankjesse: extract the value of `parameters.get()` ? You're calling it 3 times

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -94,60 +100,82 @@ private TypeElement getElement(Class<?> clazz) {
...
+  interface ExtendsOthers extends Callable<Integer>, Comparable<Long> {
+    // empty on purpose
+  }
+
   @Test public void overrideEverything() {
     TypeElement classElement = getElement(Everything.class);
     ExecutableElement methodElement = getOnlyElement(methodsIn(classElement.getEnclosedElements()));
-
-    MethodSpec method = MethodSpec.overriding(methodElement).build();
+    ExecutableType type = (ExecutableType) compilation.getTypes().asMemberOf((DeclaredType) classElement.asType(), methodElement);
```
swankjesse: nit: we wrap at 100 columns

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -94,60 +100,82 @@ private TypeElement getElement(Class<?> clazz) {
...
-        + "}");
+        + "public java.lang.String toString() {\n"
+        + "}\n");
+  }
+
+  @Test public void overrideExtendsOthersWorksWithActualTypeParameters() {
+    TypeElement classElement = getElement(ExtendsOthers.class);
+    List<ExecutableElement> methods = methodsIn(compilation.getElements().getAllMembers(classElement));
+    Iterator<ExecutableElement> iter = methods.iterator();
+    while (iter.hasNext()) {
```
swankjesse: Definitely not a fan of while loops in tests. Can you write a test for `call` and a separate test for `compareTo` ?

As is this will pass if the iterator has no elements, or if the methods aren't found

swankjesse:
Dig. I think the thing this still needs is code to unpack an AnnotationMirror that has parameters. I'm guessing that isn't going to be rendered properly.

src/main/java/com/squareup/javapoet/MethodSpec.java
sormuras: Tried to insert them where needed. Missed those. But why didn't the Travis report the checkstyle problem?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Sure. Will move the code to the builder construct. Copied the get()-ty way from ClassName ... but you never have those static factory methods when there's a builder available. Correct?

sormuras:
Didn't tested an AnnotationMirror with parameters... ehm, what kind of parameters? Type variables?

src/test/java/com/squareup/javapoet/MethodSpecTest.java
sormuras: Good point. Will split the loop in distinct tests.

src/main/java/com/squareup/javapoet/MethodSpec.java
sormuras: 2 times, only. One is `types.get()`. Will introduce a local variable, anyway.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
sormuras: Need to install checkstyle plugin... ;-)

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: It's a mix.

src/main/java/com/squareup/javapoet/MethodSpec.java
swankjesse: checkstyle isn't smart enough.

swankjesse:
Sorry, annotations that have attributes. An example is from libcore's annotation tests:
https://android.googlesource.com/platform/libcore/+/master/luni/src/test/java/libcore/java/lang/reflect/AnnotationsTest.java

```
    static enum Breakfast { WAFFLES, PANCAKES }
    @Retention(RetentionPolicy.RUNTIME)
    public @interface HasDefaultsAnnotation {
        byte a() default 5;
        short b() default 6;
        int c() default 7;
        long d() default 8;
        float e() default 9.0f;
        double f() default 10.0;
        char g() default 'k';
        boolean h() default true;
        Breakfast i() default Breakfast.WAFFLES;
        AnnotationA j() default @AnnotationA();
        String k() default "maple";
        Class l() default AnnotationB.class;
        int[] m() default { 1, 2, 3 };
        Breakfast[] n() default { Breakfast.WAFFLES, Breakfast.PANCAKES };
        Breakfast o();
        int p();
    }
```

sormuras:
Btw, what do you think of the extra parameter in MethodSpec.overriding()?

I tried to avoid adding it, but then there's no way(?) to correctly expand the actual type parameters used. See return value of asMemberOf() in http://docs.oracle.com/javase/7/docs/api/javax/lang/model/util/Types.html#asMemberOf(javax.lang.model.type.DeclaredType,%20javax.lang.model.element.Element).

The use of that method could be hidden by implementing a convenient wrapper that cares about casting as well
```java
public static Builder overriding(ExecutableElement method, Types types, TypeMirror containing) {
  return overriding(method, (ExecutableType) types.asMemberOf((DeclaredType) containing, method));
}
```

sormuras:
Yeah, that's nice annotation, that @ HasDefaultsAnnotation. Will incorporate that in the AnnotationSpecTest.java.

Just to sort things out, with parameters/attributes you mean "default values", right? If that's the case, only adding an extra parameter, namely Elements, will do the trick: http://docs.oracle.com/javase/7/docs/api/javax/lang/model/util/Elements.html#getElementValuesWithDefaults(javax.lang.model.element.AnnotationMirror)

swankjesse:
I'm mostly concerned with things like `@Named("foo")` copying over parameter values correctly. This includes things like string escaping, adding `.class` to type names, and using `$T` so that referenced types are imported. Plus annotations can contain other annotations, so it's a lot of work.

sormuras:
Testing the static get() yields no syntactical problem. See https://github.com/sormuras/javapoet/commit/d54d90569a8c65ca8e9b8913c8536954a1f1772e
```java
  @Test
  public void testHasDefaultAnnotation() {
    String name = IsAnnotated.class.getCanonicalName();
    TypeElement element = compilation.getElements().getTypeElement(name);
    AnnotationSpec annotation = AnnotationSpec.get(element.getAnnotationMirrors().get(0));
    assertThat(annotation.toString()).isEqualTo(
        "@com.squareup.javapoet.AnnotationSpecTest.HasDefaultsAnnotation("
            + "o = com.squareup.javapoet.AnnotationSpecTest.Breakfast.PANCAKES"
            + ", p = 1701"
            + ", f = 11.1"
            + ", l = java.lang.Override.class"
            + ", j = @com.squareup.javapoet.AnnotationSpecTest.AnnotationA"
            + ", q = @com.squareup.javapoet.AnnotationSpecTest.AnnotationC(\"bar\")"
            + ")");
  }
```
But `$T` and import support is another story...

swankjesse:
If you'd like, I'm okay with fully-qualified type names for the first pass. We can save the `$T` thing for a possible follow-up.

sormuras:
...only 20% away from working `$T` using `AnnotationValueVisitor`. Just the annotation in an annotation and the list/array of annotation values are missing.

sormuras:
99% ... what is missing, Jesse?

I didn't move the code to the builder, yet. First making it work, then beautiful. This time.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +107,60 @@ public static Builder builder(Class<?> type) {
...
+        ClassName.get((TypeElement) annotation.getAnnotationType().asElement()));
+    AnnotationSpecBuildingVisitor visitor = new AnnotationSpecBuildingVisitor(builder);
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry
+        : annotation.getElementValues().entrySet()) {
+      entry.getValue().accept(visitor, entry);
+    }
+    return builder.build();
+  }
+
+  public static class AnnotationSpecBuildingVisitor
```
swankjesse: Definitely shouldn't be public

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +107,60 @@ public static Builder builder(Class<?> type) {
...
+    public AnnotationSpecBuildingVisitor(Builder builder) {
+      super(builder);
+      this.builder = builder;
+    }
+
+    @Override
+    protected Builder defaultAction(Object o, Map.Entry<? extends ExecutableElement,
+                                                        ? extends AnnotationValue> entry) {
+      String name = entry.getKey().getSimpleName().toString();
+      return builder.addMember(name, CodeBlock.builder().add(entry.getValue().toString()).build());
```
swankjesse: Alternately: 
```
  builder.addMember(name, "$S", entry.getValue());
```

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +107,60 @@ public static Builder builder(Class<?> type) {
...
+      String name = entry.getKey().getSimpleName().toString();
+      return builder.addMember(name, "$L", get(a));
+    }
+
+    @Override
+    public Builder visitEnumConstant(VariableElement c,
+                                     Map.Entry<? extends ExecutableElement,
+                                               ? extends AnnotationValue> entry) {
+      String name = entry.getKey().getSimpleName().toString();
+      return builder.addMember(name, "$T.$L", c.asType(), c.getSimpleName());
```
swankjesse: neat

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +107,60 @@ public static Builder builder(Class<?> type) {
...
+    }
+
+    @Override
+    public Builder visitType(TypeMirror t,
+                             Map.Entry<? extends ExecutableElement,
+                                       ? extends AnnotationValue> entry) {
+      String name = entry.getKey().getSimpleName().toString();
+      return builder.addMember(name, "$T.class", t);
```
swankjesse: neat

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -99,6 +107,60 @@ public static Builder builder(Class<?> type) {
...
+
+    @Override
+    public Builder visitEnumConstant(VariableElement c,
+                                     Map.Entry<? extends ExecutableElement,
+                                               ? extends AnnotationValue> entry) {
+      String name = entry.getKey().getSimpleName().toString();
+      return builder.addMember(name, "$T.$L", c.asType(), c.getSimpleName());
+    }
+
+    @Override
```
swankjesse: nit: we try to pack our signature lines onto fewer lines, not more. This one would be:

```
    @Override public Builder visitType(TypeMirror t,
        Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry) {
```

src/main/java/com/squareup/javapoet/MethodSpec.java
swankjesse: this specific if still needs `{` and `}`

src/main/java/com/squareup/javapoet/MethodSpec.java
swankjesse: please do

src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
```diff
@@ -0,0 +1,133 @@
+package com.squareup.javapoet;
```
swankjesse: Please add Square copyright

src/test/java/com/squareup/javapoet/MethodSpecTest.java
swankjesse: Please do

swankjesse:
Structurally this is awesome. Style wise I have some nitpicks. Please address 'em, and then use `http://rebaseandsqua.sh`

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: One more question. Locally, I changed the signature to
```java
public static Builder builder(AnnotationMirror annotation) {
  Builder builder = builder(ClassName.get(...
  ...
  return builder;
}
```
D'accord?

sormuras: Reading your first comment once more - you didn't voted against the get():AnnotationSpec signature, right? Just the location within the file was the issue.

swankjesse: I like a single method `get()` that returns an `AnnotationSpec`.

sormuras:
I'll rebase by hand and split the PR into two parts:
 * AnnotationSpec.get() + Test
 * MethodSpec.implementing() implementation

sormuras:
Stripped down to only AnnotationSpec[Test] changes, rebased and squashed. Ready to merge?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -90,6 +98,18 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
     codeWriter.emit(whitespace + "}");
   }
 
+  public static AnnotationSpec get(AnnotationMirror annotation) {
```
swankjesse: Nice.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -152,4 +172,50 @@ public AnnotationSpec build() {
...
+    Entry(String name, AnnotationValue value) {
+      this.name = name;
+      this.value = value;
+    }
+  }
+
+  /**
+   * Annotation value visitor adding members to the given builder instance.
+   */
+  private static class Visitor extends SimpleAnnotationValueVisitor7<Builder, Entry> {
```
swankjesse: This is great.

swankjesse:
Two last small steps:
 * Please run http://rebaseandsqua.sh to keep our git history tidy.
 * Please complete Square's [contributor agreement](https://github.com/square/javapoet/blob/master/CONTRIBUTING.md) so we can redistribute your code.

swankjesse:
Looking forward to this!

sormuras:
Hm, I tried to rebase and squash inside Eclipse via EGit. What went wrong? I don't have a git/curl-enabled shell available at the moment.
I could revert all my history on this fork and re-apply only those two file. Will that do?

Concerning the contributor agreement, isn't it sufficient, if I say: use these contributions free-ly. Free as in beer. Or does does the agreement just say this ... in more words?

sormuras:
Dooh. I didn't close it! It's been closed by forced push... gonna re-open it and get the one and only commit into here.

kojilin:
Hi, how about treating array like below?
```
@Override
public AnnotationSpec.Builder visitArray(List<? extends AnnotationValue> vals, Entry entry) {
  Visitor visitor = new Visitor(builder);
  vals.forEach(val -> val.accept(visitor, new Entry(entry.name, val)));
  return builder;
}
```
I think it will make something convenient, especially my use case is extract codeblocks(Array) from AnnotationSpec then create new AnnotationSpec with more element.

sormuras:
@kojilin Lambdas are not supported by JavaPoets' Java 7 dependency. An old Java-5-style foreach-loop does the trick here.
Your proposal applied yields in extra new lines:
```java
package com.squareup.javapoet;

import java.lang.Override;

@AnnotationSpecTest.HasDefaultsAnnotation(
    o = AnnotationSpecTest.Breakfast.PANCAKES,
    p = 1701,
    f = 11.1,
    m = {
        9,
        8,
        1
    },
    l = Override.class,
    j = @AnnotationSpecTest.AnnotationA,
    q = @AnnotationSpecTest.AnnotationC("bar")
)
class IsAnnotated {
}
```
 instead of just `m = {9, 8, 1},`. Is your request just a format optimization or are there other impacts (malformed syntax-related) by letting a visitor visit all array values?

swankjesse:
@sormuras the agreement confirms that you own the code, and you're giving permission to Square to redistribute it.

kojilin:
Sorry I forget to remove lambda. I want to do something like below.
```
AnnotationSpec annotationSpec = annotationSpec.get(annotationMirror);
....
AnnotationSpec.Builder replace = AnnotationSpec.builder((ClassName) annotationSpec.type);
Map<String, List<CodeBlock>> members = annotationSpec.members;
for(String key:members.keySet()){
  List<CodeBlock>codeBlocks=members.get(key);
  for(CodeBlock codeBlock:codeBlocks){
    replace.addMember(key,codeBlock);
  }
}
replace.addMember("value","$S","Hoge");
```
I think it's easier to achieve this if can treat it like array?

sormuras:
@swankjesse Confirmed and agreed.

@kojilin ...may be, you can rewrite it like this:
```java
AnnotationSpec annotationSpec = annotationSpec.get(annotationMirror);
...
AnnotationSpec.Builder replace = annotationSpec.toBuilder();
replace.addMember("value","$S","Hoge");
```

swankjesse:
Thanks!

swankjesse:
I just had a thought though . . . do imports do the right thing if there's a class inside of an array?

sormuras:
*Glad contributing back to the poetry.*

As I am writing more test cases, I'll one for classes inside of an array.

Locally, I added the `visitArray` @kojilin proposed. That should also care about type import inside arrays - only the emitted java code format is slightly un-beautiful. See me first reply concerning *extra lines*.

PR #272, opened 2015-05-08
Poetry, or javax.lang.model-related utilities.

sormuras:
The name `Poetry` is an abstract working title. Is `JavaxModelUtil` better?

The purpose is to collect non-trivial parsing/transforming methods like `MethodSpec.overriding()`, that usually need processing utilities like Elements and Types from [javax.lang.model.util](http://docs.oracle.com/javase/7/docs/api/javax/lang/model/util/package-summary.html) in one place. 
In Poetry, you can also re-use common utility methods like `ClassName.isClassOrInterface(Element)` or `ClassName.getPackage(Element)`.
`AnnotationSpec.get(AnnotationMirror)` and all related private classes are a good candidates for moving, too. Actually, even all trivial methods like `ClassName.get(TypeElement)` could move to Poetry. That way, you keep the JavaPoet mode classes clean & simple & free from static factories.

Instead of writing:
```java
MethodSpec ms = MethodSpec.overriding(method, containing, type, elements);
```
you get:
```java
MethodSpec ms = new Poetry(type, elements).overriding(method, containing);
```
The complete implementation of `MethodSpec.overriding()` is included.

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -22,16 +22,10 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import javax.lang.model.SourceVersion;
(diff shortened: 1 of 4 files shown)


sormuras:
Closing this PR, needs more work and style.

Issue #267, opened 2015-05-03
JavaFile.toJavaFileObject()


jgustie:
I have been working on some tests using [Compile Testing](https://github.com/google/compile-testing) which will accept a `javax.tools.JavaFileObject`; it occurred to me that it might be nice if there were a `JavaFile.toJavaFileObject()` method that could be used to pass generated code into the `java.tools.JavaCompiler`.

This would simplify using JavaPoet for generating the code being input to the compilation process (which otherwise might be done using Compile Testing's `JavaFileObjects.forSourceLines(...)`). It may also simplify isolating the tests from code formatting changes in JavaPoet: an annotation processor using JavaPoet could be validated using Compile Testing's `generatesSources(...)` method with input also backed by JavaPoet.

The obvious work around is just using something like `JavaFileObjects.forSourceLines(javaFile.packageName + '.' + javaFile.typeSpec.name, Splitter.on('\n').split(javaFile.toString()))` in the tests; but that assumes you have Compile Testing (in theory this does not need to be restricted to test environments). Looking at `JavaFileObjects`, just subclassing `javax.tools.SimpleJavaFileObject` to wrap a `JavaFile` would be relatively straight forward (I can try to get a pull request in if there is any interest).

swankjesse:
Not sure how actionable this is.

If you're using Compile Testing, the workaround is pretty good, right? Maybe that project could add a method like `JavaFileObjects.forSource(String, String)`.

If you aren't using Compile Testing, I don't know why you'd want to get a `JavaFileObject` instead of going through `Filer`. What's the use case?

jgustie:
I suppose the `Splitter` part of the workaround isn't necessary, the `forSourceLines` method in question takes a varargs.

A `toJavaFileObject()` would be kind of like the opposite of what you currently get from a `Filer`: the `JavaFile.writeTo(Filer)` method uses the `Writer` of the `JavaFileObject` as opposed to the `Reader`/character content that you would need to pass a `JavaFile` _into_ the compiler:

```
JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
DiagnosticCollector<JavaFileObject> diagnosticCollector = new DiagnosticCollector<>();
JavaFileManager fileManager = compiler.getStandardFileManager(diagnosticCollector, Locale.getDefault(), StandardCharsets.UTF_8);
CompilationTask task = compiler.getTask(null, 
    fileManager, diagnosticCollector,
    Collections.emptySet(), Collections.emptySet(),
[...]

PR #276, opened 2015-05-13
JavaFile.toJavaFileObject

jgustie:
As discussed in #267, adds a helper to simplify passing a `JavaFile` to a `javax.tools.JavaCompiler`.

### DIFF
--- a/src/test/java/com/squareup/javapoet/FileReadingTest.java
+++ b/src/test/java/com/squareup/javapoet/FileReadingTest.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.Locale;
+import java.util.concurrent.Callable;
+
+import javax.lang.model.element.Modifier;
+import javax.tools.DiagnosticCollector;
+import javax.tools.JavaCompiler;
+import javax.tools.JavaCompiler.CompilationTask;
+import javax.tools.JavaFileObject;
+import javax.tools.JavaFileObject.Kind;
+import javax.tools.StandardJavaFileManager;
+import javax.tools.StandardLocation;
+import javax.tools.ToolProvider;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import com.google.common.io.ByteStreams;
+
+@RunWith(JUnit4.class)
+public class FileReadingTest {
+  
+  // Used for storing compilation output.
+  @Rule public final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test public void javaFileObjectUri() {
+    TypeSpec type = TypeSpec.classBuilder("Test").build();
+    assertThat(JavaFile.builder("", type).build().toJavaFileObject().toUri())
+        .isEqualTo(URI.create("Test.java"));
+    assertThat(JavaFile.builder("foo", type).build().toJavaFileObject().toUri())
+        .isEqualTo(URI.create("foo/Test.java"));
+    assertThat(JavaFile.builder("com.example", type).build().toJavaFileObject().toUri())
+        .isEqualTo(URI.create("com/example/Test.java"));
+  }
+  
+  @Test public void javaFileObjectKind() {
+    JavaFile javaFile = JavaFile.builder("", TypeSpec.classBuilder("Test").build()).build();
+    assertThat(javaFile.toJavaFileObject().getKind()).isEqualTo(Kind.SOURCE);
+  }
+  
+  @Test public void javaFileObjectCharacterContent() throws IOException {
+    TypeSpec type = TypeSpec.classBuilder("Test")
+        .addJavadoc("Testing, 1, 2, 3!")
+        .addMethod(MethodSpec.methodBuilder("fooBar").build())
+        .build();
+    JavaFile javaFile = JavaFile.builder("foo", type).build();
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
+package com.squareup.javapoet;
```
swankjesse: nit: Square copyright statement please. (copy and paste from another file.)

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import com.google.common.io.ByteStreams;
+
+@RunWith(JUnit4.class)
+public class FileReadingTest {
+  
+  // Used for storing compilation output.
+  @Rule public final TemporaryFolder tmp = new TemporaryFolder();
```
swankjesse: nit: rename to temporaryFolder?

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Test public void javaFileObjectUri() {
+    TypeSpec type = TypeSpec.classBuilder("Test").build();
+    assertThat(JavaFile.builder("", type).build().toJavaFileObject().toUri())
+      .isEqualTo(URI.create("Test.java"));
+    assertThat(JavaFile.builder("foo", type).build().toJavaFileObject().toUri())
+      .isEqualTo(URI.create("foo/Test.java"));
```
swankjesse: style nit: +4 on wrapped lines.

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+            .returns(String.class)
+            .addModifiers(Modifier.PUBLIC)
+            .addStatement("return $S", value)
+            .build())
+        .build();
+    JavaFile javaFile = JavaFile.builder("foo", type).build();
+
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    DiagnosticCollector<JavaFileObject> diagnosticCollector = new DiagnosticCollector<>();
+    StandardJavaFileManager fileManager = compiler.getStandardFileManager(diagnosticCollector, Locale.getDefault(), StandardCharsets.UTF_8);
```
swankjesse: nit: wrap at 100 columns

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+        .addFileComment("\u00A9 Copyright character gets you everytime")
+        .build();
+    byte[] bytes = ByteStreams.toByteArray(javaFile.toJavaFileObject().openInputStream());
+    
+    // Be explicit with the default charset to assert the implicit behavior in the code
+    assertThat(bytes).isEqualTo(javaFile.toString().getBytes(Charset.defaultCharset()));
+  }
+  
+  @Test public void compileJavaFile() throws Exception {
+    final String value = "Hello World!";
```
swankjesse: sweet.

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+        fileManager,
+        diagnosticCollector,
+        Collections.<String>emptySet(),
+        Collections.<String>emptySet(),
+        Collections.singleton(javaFile.toJavaFileObject()));
+    
+    assertThat(task.call()).isTrue();
+    assertThat(diagnosticCollector.getDiagnostics()).isEmpty();
+
+    ClassLoader loader = new URLClassLoader(new URL[] { classOutput.toURI().toURL() });
```
swankjesse: interesting!

src/test/java/com/squareup/javapoet/FileReadingTest.java
```diff
@@ -0,0 +1,111 @@
...
+    Callable<?> test = Class.forName("foo.Test", true, loader).asSubclass(Callable.class).newInstance();
+    assertThat(Callable.class.getMethod("call").invoke(test)).isEqualTo(value);
```
swankjesse: This is insanely cool

swankjesse:
I'm impressed and delighted. I had no idea that this was possible, and now I want to do it everywhere. Wow. Almost makes me want to have a method on TypeSpec that returns a `Class<?>` that does the compiler and everything. Not yet!

I've got a bunch of style nitpicks. Nothing that's a big deal.

I saw that you signed the CLA.

Next steps: if you dont' mind addressing the nitpicks, please do, then squash down to one commit (http://rebaseandsqua.sh) and I'll merge it. If you want me to fix arbitrary style stuff, I can do it also.

Thanks!

jgustie:
Squashed some nits, let me know if there is anything else.

The `compileJavaTest` is largely what I wanted to accomplish: getting a `JavaFile` right into the compiler. I agree that there are other ways you can take this though: you could imagine a `JavaFileManager` implementation similar to the [Compile Testing](https://github.com/google/compile-testing) `InMemoryJavaFileManager` that by-passes the need to go to disk and compiles JavaPoet output directly into a live class loader; or you could have a no-op file manager and just capture the output from the diagnostic collector and provide real-time compilation errors for a `JavaFile` (take it a step further and use something like [Error Prone](http://errorprone.info/) for additional analysis); you could also rig something like `JavaFile.process(Processor)` to run other annotation processors over generated code by passing the processors into the `CompilationTask` (might be useful for [AutoService](https://github.com/google/auto/tree/master/service)).

Seed planted ;)

sormuras:
I'd very happy to see this merged. Then, I could remove a lot of internal test code coping with in-memory compiltation.

+1

swankjesse:
Fun. I'm guessing it's nearly time to cut a release?

jgustie:
I've been thinking (as I sometimes do): should `JavaFile` just implement `JavaFileObject` itself? It might be less awkward then a minimal anonymous wrapper class hiding behind a long method name.

I don't know how you feel about mucking up your public API like that, but conceptually it kind of makes sense. I have the changes done up in my local clone on this branch: I can push it or put them on a different branch if it is something you are interested in. Technically I made `JavaFile` extend `SimpleJavaFileObject` which is something I might be less inclined to do in a pure Java 8 environment (default implementations mean you do not strictly need that buffer between `JavaFileObject`), but it makes the change pretty minimal. The biggest downside is the impact to what JavaPoet exposes (which is why I would completely understand if you want to leave things the way they are).

swankjesse:
I don't want to implement `JavaFileObject` directly. Especially since there's an entire type hierarchy there that's a different model than our `JavaFile`.

jgustie:
Fair enough. I did have one other minor clean up that I discovered after the fact: `StandardJavaFileManager.getClassLoader(StandardLocation.CLASS_OUTPUT)` builds a more appropriate `URLClassLoader` then what I had used (actually it is exactly the same, it just obtained more appropriately). I also noticed my Eclipse import cleaner is setup differently so the `FileReadingTest` has static imports first with package breaks in the rest of the import statements. Would you like those two clean ups on this PR?

swankjesse:
Yes please!

jgustie:
I guess since this has already been merge that branch needs a new PR. #267 can be closed out at this point as well.

PR #279, opened 2015-05-15
AnnotationSpec fixes

sormuras:
AnnotationSpec.toBuilder() did return an immutable collection.
AnnotationSpec.get(AnnotationMirror) visitor now visits arrays as well.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -121,7 +121,9 @@ public final class AnnotationSpec {
 
   public Builder toBuilder() {
     Builder builder = new Builder(type);
-    builder.members.putAll(members);
+    for (Map.Entry<String, List<CodeBlock>> entry : members.entrySet()) {
+      builder.members.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+    }
     return builder;
   }
 
@@ -215,6 +217,14 @@ public final class AnnotationSpec {
     @Override
     public Builder visitType(TypeMirror t, Entry entry) {
       return builder.addMember(entry.name, "$T.class", t);
+    }
+
+    @Override
+    public Builder visitArray(List<? extends AnnotationValue> values, Entry entry) {
+      for (AnnotationValue value : values) {
+        value.accept(this, new Entry(entry.name, value));
+      }
+      return builder;
     }
   }
 
--- a/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
@@ -85,6 +85,9 @@ public final class AnnotationSpecTest {
     int p();
 
     AnnotationC q() default @AnnotationC("foo");
+
+    Class<? extends Number>[] r() default {Byte.class, Short.class, Integer.class, Long.class};
+
   }
 
   @HasDefaultsAnnotation(
@@ -94,7 +97,8 @@ public final class AnnotationSpecTest {
       m = {9, 8, 1},
       l = Override.class,
       j = @AnnotationA,
-      q = @AnnotationC("bar"))
+      q = @AnnotationC("bar"),
+      r = {Float.class, Double.class})
   public class IsAnnotated {
     // empty
   }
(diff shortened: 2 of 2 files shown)


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -122,6 +122,9 @@ public static Builder builder(Class<?> type) {
   public Builder toBuilder() {
     Builder builder = new Builder(type);
     builder.members.putAll(members);
+    for (Map.Entry<String, List<CodeBlock>> entry : builder.members.entrySet()) {
```
tbroyer: I think I'd have rather looped over the `members`' entry set to `put()` each entry into `builder.members` instead of using `putAll` and then replacing the entries' values.

tbroyer: Maybe add a `mutableMultimap()` method in `Util` to do just that? (reverse of `immutableMultimap()` as used in `build()`)

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -216,6 +219,15 @@ public Builder visitEnumConstant(VariableElement c, Entry entry) {
     public Builder visitType(TypeMirror t, Entry entry) {
       return builder.addMember(entry.name, "$T.class", t);
     }
+
+    @Override
+    public Builder visitArray(List<? extends AnnotationValue> values, Entry entry) {
+      Visitor visitor = new Visitor(builder);
```
tbroyer: Why not just pass `this`?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Sure. You're very welcome to change that.

Btw. all other toBuilder() method in *Spec classes should work - as there're no nested immutable collections. I think. Will write some unit tests asserting that.

src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
```diff
@@ -130,19 +135,96 @@ public void testHasDefaultAnnotationWithImport() {
...
+            + "import java.lang.Double;\n"
+            + "import java.lang.Float;\n"
             + "import java.lang.Override;\n"
             + "\n"
             + "@AnnotationSpecTest.HasDefaultsAnnotation(\n"
             + "    o = AnnotationSpecTest.Breakfast.PANCAKES,\n"
             + "    p = 1701,\n"
             + "    f = 11.1,\n"
-            + "    m = {9, 8, 1},\n"
+            + "    m = {\n        9,\n        8,\n        1\n    },\n"
```
tbroyer: Would be more readable with one logical line per physical line (and only having `\n` at end of lines)

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: True. Why not. Using `this` here. Tests stay green, less memory used.

src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
sormuras: Good point - my intention was not to mess around with more "empty" new lines.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Regarding the `mutableMultimap()`: then, one has to either remove the `final` modifier from Builder#members or pass the target map as a second parameter to the utility method. Both are not so nice solutions, imho. What do you think? Do it inline as is, and iterate over members' entry set.

sormuras: ...or it resolves to a code block like:
```java
  public Builder toBuilder() {
    Builder builder = new Builder(type);
    builder.members.putAll(Util.mutableMultimap(members));
    return builder;
  }
```
Here, there's a map populated twice. I'm still favoring the inline method as initially proposed.

swankjesse:
Thanks!

Issue #273, opened 2015-05-10
How do I convert AnnotationMirror to AnnoationSpec? and convert ExecutableElement to MethodSpec?


jiqimaogou:

sormuras:
Have a look  at [AnnotationSpec.java#L101](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/AnnotationSpec.java#L101) in current master HEAD and [MethodSpec.java#L172](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/MethodSpec.java#L172) for an example of the second one - or here [Poetry](https://github.com/sormuras/javapoet/blob/poetry/src/main/java/com/squareup/javapoet/Poetry.java).

jiqimaogou:
thank you, but if I just want to copy not overriding the method, (overriding exception @override annotation).

sormuras:
Then you just copy the fields 1:1 from the source ExecuteElement/ExecutableType into the MethodSpec.Builder instance.

swankjesse:
Would it be sufficient to offer a method to remove the offending `@Override` annotation?

sormuras:
What about `MethodSpec.get(...)` implementation that just copies all attributes from ExecutableElement et al and returns a Builder. And convenient `MethodSpec.getOverriding(...)` that adds the `@Override` annotation if needed?

swankjesse:
Like this?

```
public static Builder MethodSpec.methodBuilder(String name)
public static Builder MethodSpec.methodBuilder(ExecutableElement copyFrom)
public static Builder MethodSpec.overriding(ExecutableElement method)
```

sormuras:
Ja-in.

With `MethodSpec.overriding(ExecutableElement method)` only calling `MethodSpec.methodBuilder(ExecutableElement copyFrom)` and add the `@Override` annotation. And then, add a `MethodSpec.implementing(ExecutableElement abstractMethod)` as well. ;-)

All those not-so-trivial methods yield for an own class, don't they? Especially, when it comes to helper functions around `javax.lang.model` utilities like `getPackageOfElement()` and others. Add some references to Types/Elements and you'll end up with large amount of code in JavaPoet core classes, that does parsing work.

[...]

PR #280, opened 2015-05-16
MethodSpec.varargs(boolean) added

sormuras:
First, the PR contains a minor update of the varargs()-method in the MethodSpec.Builder:
 * added method varargs(boolean varargs) which allows a) setting the value to false and b) cleaner code in API usage, i.e. `builder.varargs(method.isVarArgs())` instead of
```java
if (method.isVarArgs()) {
  builder.varargs();
}
```
 * second, the `boolean chainable` is introduced. It indicates a assignable type relation between the method container and the method return type. Usually, it means, the method returns `this`. That relation is difficult to reflect after the MethodSpec is built. It is easier to obtain, when the annotation process is running and providing Types utilities:
```java
boolean chainable = processingEnv.getTypeUtils().isAssignable(declaringType, method.getReturnType());
```
Knowing chainability(?) when creating return statements is crucial, when you're creating facade implementations of methods. See [StashGenerator.java#L172](https://github.com/sormuras/stash/blob/master/src/stash/de/sormuras/stash/StashGenerator.java#L172) as an example.

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -323,7 +323,11 @@ public final class MethodSpec {
     }
 
     public Builder varargs() {
-      this.varargs = true;
+      return varargs(true);
+    }
+
+    public Builder varargs(boolean varargs) {
+      this.varargs = varargs;
       return this;
     }
 


swankjesse:
On the chainable thing: should we just let you attach an arbitrary object to your `MethodSpec`, for the application to interpret? Might want to track chainable in one context, but something different elsewhere: how much work the method does, whether it has side effects, etc.

Which leads me to wonder if this is the right place for this state. I agree it's superconvenient for your use case, but I don't anticipate it being useful generally.

sormuras:
Like the good old Delphi [tag](http://stackoverflow.com/questions/1690764/what-is-the-purpose-of-the-tag-property-of-delphi-vcl-components)?

swankjesse:
Yep. (Though I'm not fortunate enough to have done any Delphi!)

sormuras:
What other solutions are available? Because the general purpose Object-tag is a kind of no go pattern. 
 * Extend MethodSpec and it's Builder ... lot of API change incomming.
 * Maintain a custom map like `Map<MethodSpec, Context>`, where `Context` collects all things of interest.
 * Have a synthetic annotation like `@Chainable` which can be attached to the MethodSpec via its Builder.

swankjesse:
Just curious: why no-go? As far as I know Android has it also!

sormuras:
I have a constant fear that someone will override my tag before I can handle its value. It just feels quick and dirty. And not type-safe.
An analog solution but far better implementation is [ClassValue](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassValue.html) for every class object in Java.  To my use-case here, it seems a bit of an overkill.

I think, I'll revoke this PR and stick with the custom annotation - as I only this flag for now.

sormuras:
Here are more tag-gy concerns: http://stackoverflow.com/a/18184505/1431016

swankjesse:
Makes sense! I still dig the varargs part of this change, if you wanna resend that!

sormuras:
Sure. And working on the MethodSpec.overriding 'poetry' as well - will add comments here: https://github.com/square/javapoet/issues/273

sormuras:
Stripped the chainable part - should have been two PRs in the first place.

PR #281, opened 2015-05-17
Proposing JavaxLangModelPoet for parsing javax.lang.model elements and types.

sormuras:
Follow up from https://github.com/square/javapoet/issues/273

Implementation outline:
 * Package-private JavaxLangModelPoet w/ test created.
 * Moved/finished `overriding` implementation and all related tests to this poet and its test class.
 * Enhanced signature to `MethodSpec.overriding(Elements, Types, ExecutableElement, DeclaredType)` - all necessary ensuring generic type resolution at processing time.

Two (or more) issues:
 * needed a way to interact with (read clear) the MethodSpec.Builder annotation and modifier lists. Instead of removing the `private` from the field declaration, I introduced package-private accessor methods: `MethodSpec.Builder#annotations()` and `~#modifiers()`. See [JavaxLangModelPoet.java#L130](https://github.com/sormuras/javapoet/blob/javaxlangmodelpoet/src/main/java/com/squareup/javapoet/JavaxLangModelPoet.java#L130)
 * the builder()-method does *not* copy the `@Override` annotation from `HasAnnotation#toString` in the unit test - I think, that's okay, because `@Override` uses `RetentionPolicy.SOURCE` and the unit test utilizes Googles `CompilationRule` that analyzes the annotation at runtime. I guess, when the test is running inside a real processing environment, the `@Override` carries over as expected. Need to verify my brave assumption, though.

What do you think?

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaxLangModelPoet.java
+++ b/src/main/java/com/squareup/javapoet/JavaxLangModelPoet.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import static com.squareup.javapoet.Util.checkNotNull;
+
+import java.util.LinkedHashSet;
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/JavaxLangModelPoet.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Create a method spec builder that copied 1:1 from {@code method} and is viewed as being a
+   * member of the specified {@code containing} class or interface.
+   *
+   * This will copy its modifiers, annotations, type parameters, return type, name, parameters and
+   * throws declarations.
+   */
+  public MethodSpec.Builder builder(ExecutableElement method, DeclaredType containing) {
```
sormuras: After the merge of https://github.com/square/javapoet/pull/287 - what to do with this feature of creating an 1:1 copy from an `ExecutableElement`?

swankjesse: Yeah, I think if we were ever to do this, it would just be a factory method on MethodSpec.builder().

PR #282, opened 2015-05-17
Added source plugin to pom.xml

sormuras:
Creates a `javapoet-{version}-sources.jar` next to `javapoet-{version}.jar`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,21 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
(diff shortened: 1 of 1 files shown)


JakeWharton:
Why is this useful? You can invoke `sources:jar` for this. We also publish sources with the snapshot to Sonatype and the releases (of course) include source jars.

sormuras:
Nevermind. Thought, for calling goal `source:jar`, the plugin has to be configured.

PR #284, opened 2015-05-19
Assuming/conditional builder method

sormuras:
Allows inline conditional operation, paired with lambdas, you get:

    builder.assume(true, b -> b.addMember("flagPresent", "1"));

instead of:

    if (true) {
      builder.addMember("flagPresent", "1");
    }

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -148,7 +148,7 @@ public final class AnnotationSpec {
     }
   }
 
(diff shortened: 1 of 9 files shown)


kenzierocks:
FYI you have some checkstyle problems.

sormuras:
One space char in two javadoc blocks ... learning: never document.(-:

swankjesse:
I don't like this idea. Method chaining is good because it shortens repetitive, linear code. But shortening control flow is quite weird. I don't think this is more maintainable or more readable.

PR #290, opened 2015-06-11
Minor changes to TypeName.list

shalecraig:
- Initialize the ArrayList to some size.
- ~~Use array index lookup, instead of an interator.~~

I can squash the two separate commits, if you want.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -245,7 +245,7 @@ public class TypeName {
 
   /** Converts an array of types to a list of type names. */
   static List<TypeName> list(Type[] types) {
-    List<TypeName> result = new ArrayList<>();
+    List<TypeName> result = new ArrayList<>(types.length);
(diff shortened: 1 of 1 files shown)


src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -245,9 +245,10 @@ public static TypeName get(Type type) {
 
   /** Converts an array of types to a list of type names. */
   static List<TypeName> list(Type[] types) {
-    List<TypeName> result = new ArrayList<>();
-    for (Type type : types) {
-      result.add(get(type));
+    List<TypeName> result = new ArrayList<>(types.length);
+    //noinspection ForLoopReplaceableByForEach
+    for (int i = 0; i < types.length; i++) {
```
JakeWharton: This is bytecode equivalent to using the for-each syntax on an array.

shalecraig: cool, didn't know. Will revert

Issue #238, opened 2015-02-26
Argument index


svenjacobs:
`beginControlFlow()`, `addStatement()` etc. should support `String.format()` like argument indices so that reoccurring arguments don't have to be added multiple times. For example:

```java
.beginControlFlow("if (%1$L == null || %1$L == %2$S)", varName, someValue)
```

eamonnmcmanus:
I've also felt the need for this, though I think a syntax like `$1L` might be better than `%1$L`.

swankjesse:
Agreed. `$1L` looks good. And that's what we'll do unless anyone has other ideas!

svenjacobs:
Great, I'm fine with that! :)

sormuras:
No other idea. Would be a great feature. +1 for `$1L`

swankjesse:
(Pull requests welcome.)

PR #293, opened 2015-06-16
Added the ability to use indexes in format arguments

APaloski:
In response to issue #238 I added in the ability to use indexes in the format strings.

I used the $-index-format syntax discussed in that thread and made the index start at one to be consistent with the way that java formatters act.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+
 import javax.lang.model.element.Element;
 import javax.lang.model.type.TypeMirror;
 
@@ -106,39 +107,71 @@ public final class CodeBlock {
         if (format.charAt(p) != '$') {
           nextP = format.indexOf('$', p + 1);
           if (nextP == -1) nextP = format.length();
+          formatParts.add(format.substring(p, nextP));
         } else {
           checkState(p + 1 < format.length(), "dangling $ in format string %s", format);
-          switch (format.charAt(p + 1)) {
+          /* Rules:
+           * Format must be in the form $[Numeric Index]{N, L, S, T}
+           * OR
+           * in the form ${$, >, <, [, ]}
+           * Numeric index is ONE BASED (for consistency with java formatting)
+           * Numeric indexing must refer to a format argument within the args array
+           * Only progress iteration over the argument iterator if we do not hit an indexed entry
+           */
+          int countOfIndexCharacters = 0;
+          while (isSimpleDigit(format.charAt(p + countOfIndexCharacters + 1))) {
+            countOfIndexCharacters++;
+            checkArgument(format.length() > p + countOfIndexCharacters + 1,
+                "Dangling format characters '%s' in format string '%s'",
+                format.substring(p), format);
+          }
+
+          int argsIndex = -1;
+          if (countOfIndexCharacters != 0) {
+            argsIndex = Integer.parseInt(format.substring(p + 1, p + 1 + countOfIndexCharacters));
+            checkArgument(argsIndex <= args.length,
+                "Argument index %s in '%s' is larger than number of parameters",
+                argsIndex, format);
+            checkArgument(argsIndex > 0,
+                "Argument index %s in '%s' is less than one, the minimum format index",
+                argsIndex, format);
+          }
+          switch (format.charAt(p + countOfIndexCharacters + 1)) {
             case 'N':
-              this.args.add(argToName(i.next()));
+              this.args.add(argToName((argsIndex == -1) ? i.next() : args[argsIndex - 1]));
               break;
             case 'L':
-              this.args.add(argToLiteral(i.next()));
+              this.args.add(argToLiteral((argsIndex == -1) ? i.next() : args[argsIndex - 1]));
               break;
             case 'S':
-              this.args.add(argToString(i.next()));
+              this.args.add(argToString((argsIndex == -1) ? i.next() : args[argsIndex - 1]));
               break;
             case 'T':
-              this.args.add(argToType(i.next()));
+              this.args.add(argToType((argsIndex == -1) ? i.next() : args[argsIndex - 1]));
               break;
             case '$':
             case '>':
             case '<':
             case '[':
             case ']':
+              checkState(argsIndex == -1, "$$, $>, $<, $[ and $] may not have an index");
               break;
             default:
               throw new IllegalArgumentException("invalid format string: " + format);
           }
 
-          nextP = p + 2;
+          nextP = p + countOfIndexCharacters + 2;
+          formatParts.add(String.format("%c%c", format.charAt(p), format.charAt(nextP - 1)));
         }
-
-        formatParts.add(format.substring(p, nextP));
       }
-
-      checkArgument(!i.hasNext(), "unexpected args for %s: %s", format, Arrays.asList(args));
       return this;
+    }
+
+    /**
+     * A version of {@link Character#isDigit(char)} that only accepts '0'-'9'.
+     */
+    private boolean isSimpleDigit(char toCheck) {
+      return toCheck >= '0' && toCheck <= '9';
     }
 
     private String argToName(Object o) {
--- a/src/test/java/com/squareup/javapoet/CodeBlockTest.java
+++ b/src/test/java/com/squareup/javapoet/CodeBlockTest.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.fail;
+
+public class CodeBlockTest {
+  
+  @Test
+  public void indentCannotBeIndexed() {
+    try {
+      CodeBlock.builder().add("$1>", "taco").build();
+      fail();
+    } catch (IllegalStateException exp) {
+      assertThat(exp).hasMessage("$$, $>, $<, $[ and $] may not have an index");
+    }
+  }
+  
+  @Test
+  public void deindentCannotBeIndexed() {
+    try {
+      CodeBlock.builder().add("$1<", "taco").build();
+      fail();
+    } catch (IllegalStateException exp) {
+      assertThat(exp).hasMessage("$$, $>, $<, $[ and $] may not have an index");
+    }
+  }
+  
+  @Test
+  public void dollarSignEscapeCannotBeIndexed() {
+    try {
+      CodeBlock.builder().add("$1$", "taco").build();
+      fail();
+    } catch (IllegalStateException exp) {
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
...
+          /* Rules:
+           * Format must be in the form $[Numeric Index]{N, L, S, T}
+           * OR
+           * in the form ${$, >, <, [, ]}
+           * Numeric index is ONE BASED (for consistency with java formatting)
+           * Numeric indexing must refer to a format argument within the args array
+           * Only progress iteration over the argument iterator if we do not hit an indexed entry
+           */
+          int countOfIndexCharacters = 0;
+          while (Character.isDigit(format.charAt(p + countOfIndexCharacters + 1))) {
```
swankjesse: We probably want a much simpler, much dumber set of index characters. `c >= '0' && c <= '9'`. Character.isDigit supports lots of "digit" like characters!

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
...
+            countOfIndexCharacters++;
+          }
+          int argsIndex = -1;
+          if (countOfIndexCharacters != 0) {
+            argsIndex = Integer.parseInt(format.substring(p + 1, p + 1 + countOfIndexCharacters));
+            checkArgument(!(argsIndex > args.length),
+                "Argument index %s in '%s' is larger than number of parameters",
```
swankjesse: Awesome

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
...
+           * Only progress iteration over the argument iterator if we do not hit an indexed entry
+           */
+          int countOfIndexCharacters = 0;
+          while (Character.isDigit(format.charAt(p + countOfIndexCharacters + 1))) {
+            countOfIndexCharacters++;
+          }
+          int argsIndex = -1;
+          if (countOfIndexCharacters != 0) {
+            argsIndex = Integer.parseInt(format.substring(p + 1, p + 1 + countOfIndexCharacters));
+            checkArgument(!(argsIndex > args.length),
```
swankjesse: Maybe simplify to `checkArgument(argsIndex < args.length, ...`

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
...
             case 'T':
-              this.args.add(argToType(i.next()));
+              this.args.add(argToType((argsIndex == -1) ? i.next() : args[argsIndex - 1]));
               break;
             case '$':
             case '>':
             case '<':
             case '[':
             case ']':
+              checkState(argsIndex == -1, "$$, $>, $<, $[ and $] may not have an index");
```
swankjesse: Sweet.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
         if (format.charAt(p) != '$') {
           nextP = format.indexOf('$', p + 1);
           if (nextP == -1) nextP = format.length();
+          formatParts.add(format.substring(p, nextP));
         } else {
           checkState(p + 1 < format.length(), "dangling $ in format string %s", format);
-          switch (format.charAt(p + 1)) {
+          /* Rules:
```
swankjesse: Nice doc.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,38 +107,58 @@ public Builder add(String format, Object... args) {
...
             case ']':
+              checkState(argsIndex == -1, "$$, $>, $<, $[ and $] may not have an index");
               break;
             default:
               throw new IllegalArgumentException("invalid format string: " + format);
           }
 
-          nextP = p + 2;
+          nextP = p + countOfIndexCharacters + 2;
+          formatParts.add(countOfIndexCharacters == 0
```
swankjesse: Does it do the right thing automatically without the ternary? I think it might just work

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -0,0 +1,128 @@
+package com.squareup.javapoet;
```
swankjesse: please copy & paste a square copyright header from another file!

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      CodeBlock.builder().add("$T $2T", String.class).build();
+      fail();
+    } catch (IllegalArgumentException exp) {
+      assertThat(exp).hasMessage("Argument index 2 in '$T $2T' is larger than number of parameters");
+    }
+  }
+  
+  @Test
+  public void indexTooLow() {
```
swankjesse: what about `0` ?

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -0,0 +1,128 @@
...
+      fail();
+    } catch (IllegalArgumentException exp) {
+      assertThat(exp).hasMessage("invalid format string: $T $-1T");
+    }
+  }
+  
+  @Test
+  public void indexingDoesNotIncreaseNaturalIndex() {
+    CodeBlock block = CodeBlock.builder().add("$L $L $2L $L", 1, 2, 3).build();
+    assertThat(block.toString()).isEqualTo("1 2 2 3");
```
swankjesse: great test

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+  
+  @Test
+  public void indexingCanBeInterleved() {
+    CodeBlock block = CodeBlock.builder().add("$L $3L $L $2L $L $1L", 1, 2, 3).build();
+    assertThat(block.toString()).isEqualTo("1 3 2 2 3 1");
+  }
+  
+  @Test
+  public void sameIndexCanBeUsedWithDifferentFormats() {
```
swankjesse: neato

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1043,6 +1043,32 @@ public void interfaceDefaultMethods() throws Exception {
         + "  }\n"
         + "}\n");
   }
+  
+  @Test public void indexedElseIf() throws Exception {
+    TypeSpec taco = TypeSpec.classBuilder("Taco")
+        .addMethod(MethodSpec.methodBuilder("choices")
+            .beginControlFlow("if ($1L != null || $1L == $2L)", "mTaco", "mOtherTaco")
```
swankjesse: mind dropping the `m` prefix here? We don't want to endorse that code style, not even in generated code!

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1043,6 +1043,32 @@ public void interfaceDefaultMethods() throws Exception {
...
+        .build();
+    assertThat(toString(taco)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import java.lang.System;\n"
+        + "\n"
+        + "class Taco {\n"
+        + "  void choices() {\n"
+        + "    if (mTaco != null || mTaco == mOtherTaco) {\n"
+        + "      System.out.println(\"only one taco? NOO!\");\n"
```
swankjesse: This is hilarious.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1818,15 +1844,6 @@ private String toString(TypeSpec typeSpec) {
     }
   }
 
-  @Test public void tooManyArguments() {
```
swankjesse: Interesting. Yup, that makes sense.

swankjesse:
This is wonderful, thank you!

Please complete our contributors paperwork so we can redistribute your code:

APaloski:
I added in fixes to the comments you left above and completed the contributors form. I hope all looks well.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,41 +107,69 @@ public Builder add(String format, Object... args) {
         if (format.charAt(p) != '$') {
           nextP = format.indexOf('$', p + 1);
           if (nextP == -1) nextP = format.length();
+          formatParts.add(format.substring(p, nextP));
         } else {
-          checkState(p + 1 < format.length(), "dangling $ in format string %s", format);
-          switch (format.charAt(p + 1)) {
+          checkState(p <= format.length(), "dangling $ in format string %s", format);
```
swankjesse: why this change ? I think this is a bug

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,41 +107,69 @@ public Builder add(String format, Object... args) {
...
+          /* Rules:
+           * Format must be in the form $[Numeric Index]{N, L, S, T}
+           * OR
+           * in the form ${$, >, <, [, ]}
+           * Numeric index is ONE BASED (for consistency with java formatting)
+           * Numeric indexing must refer to a format argument within the args array
+           * Only progress iteration over the argument iterator if we do not hit an indexed entry
+           */
+          int countOfIndexCharacters = 0;
+          while (isSimpleDigit(format.charAt(p + countOfIndexCharacters + 1))) {
```
swankjesse: does this crash on malformed format strings like "$" ?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -106,41 +107,69 @@ public Builder add(String format, Object... args) {
...
+            countOfIndexCharacters++;
+          }
+          int argsIndex = -1;
+          if (countOfIndexCharacters != 0) {
+            argsIndex = Integer.parseInt(format.substring(p + 1, p + 1 + countOfIndexCharacters));
+            checkArgument(argsIndex <= args.length,
+                "Argument index %s in '%s' is larger than number of parameters",
+                argsIndex, format);
+            checkArgument(argsIndex > 0,
```
swankjesse: I think you can drop this case; we never accept the `-` character anywhere.

swankjesse: ... nope I'm wrong. `0`.

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.fail;
+
+public class CodeBlockTest {
+  
+  @Test
+  public void indentCannotBeIndexed() {
```
swankjesse: Please add a test that uses "$" all by itself, and another one that tests "$1" ? I think both of these will crash in the wrong way

swankjesse: (they should throw exceptions with nice messages)

swankjesse:
I checked your CLA. Looks good.

Mind adding two more short test cases that might catch a problem in the code? Also please use http://rebaseandsqua.sh to get only a single commit for this change.

And then I'll merge this. Thanks!

src/main/java/com/squareup/javapoet/CodeBlock.java
APaloski: My mistake, changed the wrong line based upon previous comments and didn't catch it. Fixed in next check in.

svenjacobs:
Just when I was about to get started on #238 I see that the work has already been done. That's great :+1: :) Looking forward for this being merged.

swankjesse:
My apologies for sitting on this for 17 days. I missed the notification on your updated push.

LGTM, thanks!

PR #301, opened 2015-07-06
Fix bug when using javapoet with Eclipse compiler

marcosb:
For the following code:

public interface Interface1 {}
public interface Interface 2 {}
public <T extends Interface1 & Interface2> T myMethod(T input) {}

a code generator (javax.annotation.processing.Processor) in eclipse will have a TypeVariable for the return type T instead of a DeclaredType.

This results in a stack overflow exception, because the return value of Collections.singletonList(upperBound) ends up returning a list containing, in essence, the input parameter.


marcosb:
Woops, had a copy-paste error on this one; re-submitting.

PR #305, opened 2015-07-10
Workaround for Eclipse JDT/APT bug around intersection types

tbroyer:
When in presence of a TypeVariable with an intersection type as the upper
bound, Eclipse JDT/APT returns a TypeVariable equivalent to the declaring
TypeVariable when calling getUpperBound. The only way to retrieve the
actual bounds is to go look at the TypeParameterElement represented by
the TypeVariable.

See #303

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypesEclipseTest.java
+++ b/src/test/java/com/squareup/javapoet/TypesEclipseTest.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright (C) 2014 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import static com.google.common.base.Charsets.*;
+import static com.google.common.base.Preconditions.*;
+
+import java.util.Locale;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+
+import javax.annotation.processing.AbstractProcessor;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.Processor;
+import javax.annotation.processing.RoundEnvironment;
+import javax.lang.model.SourceVersion;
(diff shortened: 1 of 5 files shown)


tbroyer:
I'm absolutely not fond (read: I find it ugly) of the copied `CompilationRule` (and it cannot even use the `InMemoryJavaFileManager` as that's package-private in compile-testing) but that was the easiest way to do it.

src/test/java/com/squareup/javapoet/TypesEclipseTest.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @throws IllegalStateException if this method is invoked outside the execution of the rule.
+     */
+    public Types getTypes() {
+      checkState(elements != null, "Not running within the rule");
+      return types;
+    }
+
+    static private boolean compile(Iterable<? extends Processor> processors) {
+      JavaCompiler compiler = new EclipseCompiler();
```
eamonnmcmanus: Would I be right in thinking that the need for this copied `@Rule` would go away if the original `CompilationRule` had an optional constructor parameter that was a `JavaCompiler` or a `Supplier<JavaCompiler>`? That seems as if it would be a very useful thing for cases like this. What does @gk5885 think?

tbroyer: That's right, that's the only required change.

gk5885: Occasionally I try to take a stab at making compile testing work with `ecj`, but end up running into walls where it makes assumptions about the implementation of the filer and whatnot and ends up crashing.  I haven't tried in particular with `CompilationRule` though.  It might work, but I wouldn't be remotely surprised if it didn't either.

tbroyer: Ah right, I also had to change from compiling dummy sources to processing a class instead, because ECJ requires "all JavaFileObjects be Files" as you note in google/compile-testing#31.

So maybe `CompilationRule` could just use the `JavaCompiler` directly taking into account those limitations, rather than calling `Compilation.compile()`. AFAICT, it doesn't really need an `InMemoryJavaFileManager` either, or processing the diagnostics, so it would be really simple (well, just like the rule in this PR)

tbroyer:
I'll revisit this PR once google/compile-testing#71 is merged. I might revisit it earlier though wrt the simplification discussed in #303 to unconditionally use `asElement().getBounds()`.

swankjesse:
Wow, that's involved! Should I merge as-is and you can follow up?

tbroyer:
I just added a commit with the simplification; I also added a `TypeVariableName.get(TypeParameterElement)` method.

It's good to merge IMO, and we can follow up with a simplification to the tests when compile-testing makes it possible.

swankjesse:
Works for me. Merged. Thanks @tbroyer !

PR #310, opened 2015-07-21
Create better messages for auto-indexing

### LINKED ISSUE #309: Indexed and positional CodeBlock replacement blocks error
..is awful. I don't have an example right now, but both too few and too many arguments create misleading exceptions based on list iteration.

kenzierocks:
Fixes #309.

I'm not really sure about checking for unused arguments, it seems a little useless. If it is preferred, I'll write a test for it.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -1848,7 +1848,17 @@ public final class TypeSpecTest {
     try {
       CodeBlock.builder().add("$S");
       fail();
-    } catch (NoSuchElementException expected) {
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Not enough parameters were given; expected 1, got 0");
+    }
+  }
+
+  @Test public void unusedArguments() {
+    try {
+      CodeBlock.builder().add("", "foo");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Too many parameters were given; expected 0, got 1");
     }
   }
 
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
-import java.util.Arrays;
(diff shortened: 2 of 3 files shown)


JakeWharton:
Unused arguments are an error.

JakeWharton:
Also mixing indexed and positional should be flat out forbidden as it gets crazy fast.

kenzierocks:
Ok, I'll whip up a test then.

kenzierocks:
Ok, done.

JakeWharton:
Wow you're fast! Thanks.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1848,7 +1848,17 @@ private String toString(TypeSpec typeSpec) {
...
+      assertThat(expected).hasMessage("Not enough parameters were provided; expected at least 1, got 0");
+    }
+  }
+
+  @Test public void unusedArguments() {
+    try {
+      CodeBlock.builder().add("", "foo");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Too many parameters were given; parameters [foo] were unused");
```
JakeWharton: I think this message is a bit much. Can we make them match?
> "Expected 1, got 0"
> "Expected 0, got 1"

kenzierocks:
@JakeWharton I'll remove support for that then, and probably change how the too many arguments detector works.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
kenzierocks: Yes, I'll do that with the removal of index with positional.

kenzierocks:
Alright, that should be good. Tell me if there's anything you want changed :)

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1848,7 +1848,17 @@ private String toString(TypeSpec typeSpec) {
     try {
       CodeBlock.builder().add("$S");
       fail();
-    } catch (NoSuchElementException expected) {
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("Not enough parameters were given; expected 1, got 0");
+    }
+  }
+
+  @Test public void unusedArguments() {
```
swankjesse: Nice.

swankjesse:
Neato.

PR #312, opened 2015-07-22
Shorten parameter validation.

swankjesse:
Pass the same tests with less code.


swankjesse:
@kenzierocks

kenzierocks:
I think there probably needs to be a test for mixing indexed and positioned arguments, since we don't allow it. Otherwise, looks good.

JakeWharton:
LGTM. +1 for test for mixed as follow up. Maybe I'll do it.

PR #313, opened 2015-07-22
NameAllocator.

swankjesse:


src/main/java/com/squareup/javapoet/NameAllocator.java
```diff
@@ -0,0 +1,131 @@
...
+ *   }
+ *   nameAllocator.newName("sb", "string builder");
+ * }</pre>
+ *
+ * Pass a unique tag object to each allocation. The tag scopes the name, and can be used to look up
+ * the allocated name later. Typically the tag is the object that is being named. In the above
+ * example we use {@code property} for the user-supplied property names, and {@code "string
+ * builder"} for our constant string builder.
+ *
+ * <p>One we've allocated names we can use them when generating code: <pre>   {@code
```
tbroyer: typo: Once

rharter:
👍🏼

src/main/java/com/squareup/javapoet/NameAllocator.java
```diff
@@ -0,0 +1,131 @@
...
+ *
+ *   MethodSpec.Builder builder = MethodSpec.methodBuilder("toString")
+ *       .addAnnotation(Override.class)
+ *       .addModifiers(Modifier.PUBLIC)
+ *       .returns(String.class);
+ *
+ *   builder.addStatement("$1T $2N = new $1T()",
+ *   StringBuilder.class, nameAllocator.get("string builder"));
```
eamonnmcmanus: This example code would be easier to read if this line were indented.

src/main/java/com/squareup/javapoet/NameAllocator.java
```diff
@@ -0,0 +1,131 @@
...
+ *     return sb_.toString();
+ *   }
+ * }</pre>
+ *
+ * The underscore is appended to {@code sb} to avoid conflicting with the user-supplied {@code sb}
+ * property. Underscores are also prefixed for names that start with a digit, and used to replace
+ * name-unsafe characters like space or dash.
+ */
+public final class NameAllocator {
+  public static final Set<String> JAVA_KEYWORDS = Util.immutableSet(Arrays.asList(
```
eamonnmcmanus: There's no way you could use SourceVersion.isKeyword to avoid having to declare this set?

src/main/java/com/squareup/javapoet/NameAllocator.java
```diff
@@ -0,0 +1,131 @@
...
+      throw new IllegalArgumentException("tag " + tag + " cannot be used for both '" + replaced
+          + "' and '" + suggestion + "'");
+    }
+
+    return suggestion;
+  }
+
+  public static String toJavaIdentifier(String suggestion) {
+    StringBuilder result = new StringBuilder();
+    for (int i = 0, codePoint; i < suggestion.length(); i += Character.charCount(codePoint)) {
```
eamonnmcmanus: I think this would be clearer as a `while` loop, or a `for` with an empty third part. Certainly the `codePoint` declaration without an initializer brought me up short.

src/main/java/com/squareup/javapoet/NameAllocator.java
swankjesse: Ooooh, good idea.

tbroyer: I thought the same, and then checked the javadoc: it depends on the Java version you'd execute the code in, which is not necessarily the same as the one you'll compile the generated code with/for.
In practice, the list probably hasn't changed in years, and shouldn't change in the future (for backwards compatibility), but still…
YMMV

src/main/java/com/squareup/javapoet/NameAllocator.java
swankjesse: Can do.

src/main/java/com/squareup/javapoet/NameAllocator.java
eamonnmcmanus: How would it be better to have a fixed list of keywords? Surely if SourceVersion.isKeyword could be out of sync with the "correct" list, a fixed list could be too?

kenzierocks: @tbroyer There's no reason they couldn't change it, as the JVM doesn't know about keywords and accepts identifiers with keyword names just fine. Adding to the list doesn't break bytecode compatibility.

eamonnmcmanus: The list did change in Java 1.4, with the addition of the `assert` keyword, and again in Java 5.0, with the addition of `enum`. It hasn't changed since. Both of those additions caused considerable pain and it's clear that a significant effort has been made since then to avoid adding new keywords, including for example defining the syntax for modules to special-case the `module` identifier without making it a reserved word.

I think the right thing to do is to use `SourceVersion.isKeyword`, which has been constant for as long as it has existed, and if by some misfortune a new keyword `thing` is indeed added to Java at some point, the code here should be changed to say `SourceVersion.isKeyword(s) || s.equals("thing")`.

tbroyer: SGTM

src/main/java/com/squareup/javapoet/NameAllocator.java
swankjesse: Fixed!

swankjesse:
New patch set!

eamonnmcmanus:
LGTM

Issue #253, opened 2015-03-18
How do I add an annotation to an enum value


inder123:
For Json2Java project, I need to define Gson SerializedName annotation on Enum values. How do I do that? I only see addEnumConstant() method

swankjesse:
Not currently possible. I think what we probably want is to change it so you can use this overload:

 ```
    public Builder addEnumConstant(String name, TypeSpec typeSpec) {
```

And the typeSpec there can carry annotations that get attached to the enum. The one catch is that right now we forbid annotations on anonymous types; we could just relax that now that we have a use case for it.

Wanna send a pull request? You can copy+paste `TypeSpecTest#enumWithSubclassing` as a starting point!

inder123:
I ran into a bunch of other issues trying to use JavaPoet. 
For example, to define a field, you require me to specify a Type. I don't have a type, only a name.

JakeWharton:
We require a `Type` or `TypeName`, the latter of which doesn't require the actual type to be available yet but serves as a reference to everything needed to refer to it.

kenzierocks:
This project does need much more documenting, I might PR a bunch of javadocs.

swankjesse:
@kenzierocks please don't send Javadocs that don't add value.
https://publicobject.com/2007/08/stop-writing-redundant-javadoc.html

What was missing? I tried to make our README comprehensive.

kenzierocks:
Oh, I just don't like swapping between the README and my IDE, plus it's not instantly clear where to get some things, like ClassName for TypeName parameters.

swankjesse:
Example?

swankjesse:
(or just send a pull request actually!)

PR #314, opened 2015-07-31
Issue 253 - Added support for annotations on enum values

jaybe38:
Per issue [253](https://github.com/square/javapoet/issues/253), this change relaxes the prohibition on annotations for anonymous classes and adds the annotation  to the emitter.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -131,6 +131,7 @@ public final class TypeSpec {
     try {
       if (enumName != null) {
         codeWriter.emitJavadoc(javadoc);
+        codeWriter.emitAnnotations(annotations, false);
         codeWriter.emit("$L", enumName);
         if (!anonymousTypeArguments.formatParts.isEmpty()) {
           codeWriter.emit("(");
@@ -352,7 +353,6 @@ public final class TypeSpec {
     }
 
     public Builder addAnnotations(Iterable<AnnotationSpec> annotationSpecs) {
-      checkState(anonymousTypeArguments == null, "forbidden on anonymous types.");
       checkArgument(annotationSpecs != null, "annotationSpecs == null");
       for (AnnotationSpec annotationSpec : annotationSpecs) {
         this.annotations.add(annotationSpec);
@@ -361,7 +361,6 @@ public final class TypeSpec {
     }
 
     public Builder addAnnotation(AnnotationSpec annotationSpec) {
-      checkState(anonymousTypeArguments == null, "forbidden on anonymous types.");
       this.annotations.add(annotationSpec);
       return this;
     }
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -441,6 +441,31 @@ public final class TypeSpecTest {
         + "}\n");
   }
 
+  /** https://github.com/square/javapoet/issues/253 */
+  @Test public void enumWithAnnotatedValues() throws Exception {
+    TypeSpec roshambo = TypeSpec.enumBuilder("Roshambo")
+        .addModifiers(Modifier.PUBLIC)
+        .addEnumConstant("ROCK", TypeSpec.anonymousClassBuilder("")
+            .addAnnotation(Deprecated.class)
+            .build())
+        .addEnumConstant("PAPER")
+        .addEnumConstant("SCISSORS")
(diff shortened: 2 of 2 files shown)


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -352,7 +353,6 @@ public Builder addJavadoc(String format, Object... args) {
     }
 
     public Builder addAnnotations(Iterable<AnnotationSpec> annotationSpecs) {
-      checkState(anonymousTypeArguments == null, "forbidden on anonymous types.");
```
jaybe38: Happy to be a bit cleverer here if you have some ideas how, but AFAICT I don't know at this point how I'll be used, just that I'm an anonymous class.  That means I can't limit annotation usage to enums only.

swankjesse: That's fine. We made the same relaxation for Javadoc recently.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -441,6 +441,68 @@ private boolean isJava8() {
         + "}\n");
   }
 
+  /** https://github.com/square/javapoet/issues/253 */
+  @Test public void enumWithAnnotatedValues() throws Exception {
+    TypeSpec roshambo = TypeSpec.enumBuilder("Roshambo")
+        .addModifiers(Modifier.PUBLIC)
+        .addEnumConstant("ROCK", TypeSpec.anonymousClassBuilder("")
+            .addJavadoc("Avalanche!\n")
+            .addAnnotation(ClassName.get("javax.annotation", "Nullable"))
```
swankjesse: This is an awkward example because this annotation can't actually go here. Maybe use `@Deprecated` instead?

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -441,6 +441,68 @@ private boolean isJava8() {
...
+            .build())
+        .addEnumConstant("PAPER", TypeSpec.anonymousClassBuilder("$S", "flat")
```
swankjesse: You can simplify this to just `addEnumConstant("PAPER")`. The rest is distracting because it's unrelated to annotations & enums.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -441,6 +441,68 @@ private boolean isJava8() {
...
+            .addMethod(MethodSpec.methodBuilder("toString")
+                .addAnnotation(Override.class)
+                .addModifiers(Modifier.PUBLIC)
+                .returns(String.class)
+                .addCode("return $S;\n", "paper airplane!")
+                .build())
+            .build())
+        .addEnumConstant("SCISSORS", TypeSpec.anonymousClassBuilder("$S", "peace sign")
```
swankjesse: similarly.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -441,6 +441,68 @@ private boolean isJava8() {
...
+            .build())
+        .addField(String.class, "handPosition", Modifier.PRIVATE, Modifier.FINAL)
+        .addMethod(MethodSpec.constructorBuilder()
```
swankjesse: And you can omit this stuff too!

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -441,6 +441,68 @@ private boolean isJava8() {
...
+            .addParameter(String.class, "handPosition")
+            .addCode("this.handPosition = handPosition;\n")
+            .build())
+        .addMethod(MethodSpec.constructorBuilder()
```
swankjesse: And this!

swankjesse:
Please complete our contributor's agreement so we can merge your change!

Also if you do a follow up commit, please http://rebaseandsqua.sh.

jaybe38:
All points addressed.  I _think_ I managed to do a rebaseandsqua.sh, but it didn't give any output.  Maybe I did something in the wrong order.
Let me know if there are problems though and I'll go back and do over.

swankjesse:
Yeah, your PR still has 2 commits, and we want 1.

jaybe38:
OK, this seems to have worked better - squashed commit of both commits.  That what you need?

swankjesse:
Thanks!

PR #315, opened 2015-08-01
Add support for unimportable types.

JakeWharton:
The use case which drove this feature is that I have a bunch of code that generates references to Android IDs. These are either in your package (`com.example.R.id.foo`) or Android's (`android.R.id.bar`). The latter, by convention, is never imported so that it's unambiguously a reference to a platform resource instead of your own. In the cases where an Android reference is first or the only reference, JavaPoet (understandably) imports this type. By marking the `ClassName` as un-importable, we ensure this will never happen allowing the generated code to honor conventions.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -38,8 +38,9 @@ public final class ClassName extends TypeName implements Comparable<ClassName> {
   /** From top to bottom. This will be ["java.util", "Map", "Entry"] for {@link Map.Entry}. */
   final List<String> names;
   final String canonicalName;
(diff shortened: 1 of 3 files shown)


swankjesse:
Not a fan of this implementation. ClassName is a value type, and this confuses how `equals()` works on it. I'd prefer to configure a `JavaFile`. Or even better, I think we need a _new_ class that represents a collection of `JavaFile` instances, and offers options to configure how they're imported as a group. (This is what we need for strictly-correct nested-class imports).

JakeWharton:
Makes sense. Worked around with a `CodeBlock` for now anyway. I'm not sure there's any widespread use case for this feature other than the Android `R` case.

PR #316, opened 2015-08-03
Support enum values in code format

auke-:


swankjesse:
`$T` means _type_, but this is not a type. I think you're better off doing this manually, with an enum class followed by a literal: `$T.$L`.

auke-:
Don't you think it's nicer to refer to the enum value once and let the CodeBuilder handle the imports?

swankjesse:
The problem is one of consistency. Right now `$T` always means type. With this change you'd need to consider the argument also.

auke-:
Would $N be better? The issue with $N is that it currently doesn't handle imports.

swankjesse:
I don't think we need compact syntax for it. Method calls and field accesses are far more common, and they don't have accelerators.

auke-:
Okay

swankjesse:
...Would be interesting to have an accelerator for member references, like `someObject.instanceMethod()` or `someObject.instanceField` or `SomeEnum.CONSTANT` or `SomeType.staticMethod()` etc.

We'd need a new `MemberRef` type that would include either a TypeName (for static members) or a String name (for instance members).

Issue #298, opened 2015-07-05
Conflicting simple names in the same file cause problems


swankjesse:
We reference the wrong class!

swankjesse:
Thanks @auke-.

PR #320, opened 2015-08-07
Use qualified name for conflicting simple names in the same file

auke-:
Fix for #298

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -296,9 +296,9 @@ final class CodeWriter {
       return className.canonicalName;
     }
 
-    // Look for the longest common prefix, which we can omit.
+    // Look for the longest non-conflicting common prefix, which we can omit.
     List<String> classNames = className.simpleNames();
-    int prefixLength = commonPrefixLength(classNames);
+    int prefixLength = nonConflictingCommonPrefixLength(classNames);
     if (prefixLength == classNames.size()) {
       return className.simpleName(); // Special case: a class referring to itself!
     }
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -154,6 +152,82 @@
...
+            .build())
+        .build()
+        .toString();
+    assertThat(source).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "class A {\n"
+        + "  class B {\n"
+        + "    class C {\n"
+        + "      A.Twin.D d;\n"
```
swankjesse: Shouldn't this just be Twin?

swankjesse: err, Twin.D ?

auke-: I stared for hours at this problem (and first at my generated code, why it did't compile) so i might have made a mistake here. The nested-child case seems unnecessary.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -332,12 +332,36 @@ private int commonPrefixLength(List<String> classNames) {
     for (int i = 0; i < size; i++) {
       String a = classNames.get(i);
       String b = typeSpecStack.get(i).name;
-      if (!a.equals(b)) return i;
+      if (!a.equals(b)) {
+        for (; i > 0; i--) {
+          // Does the current type have a parent type with a member type with the same name
+          if (!parentTypeHasMemberType(classNames.get(i), i)) {
```
swankjesse: Looks wrong. My guess is you want to invert this: if any parent has a member type return 0; if none do return i.

auke-: I don't think this is wrong, I'm trying to find the least-qualified non-conflicting name here.

I changed the comment and commit description to make that more clear.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -332,12 +332,37 @@ private int commonPrefixLength(List<String> classNames) {
     for (int i = 0; i < size; i++) {
       String a = classNames.get(i);
       String b = typeSpecStack.get(i).name;
-      if (!a.equals(b)) return i;
+      if (!a.equals(b)) {
```
swankjesse: I think we need another method. This method no longer does what it's named to do, nor what it's documented to do. That new method should delegate to `commonPrefixLength()`.

src/main/java/com/squareup/javapoet/CodeWriter.java
swankjesse: Ahh, got it. The import rules are hard for me to get my head around!

auke-:
Is the latest version of the patch good enough?

swankjesse:
Much apologies for the delay on this. The logic is subtle & I wanted to take my time reviewing it. Seems solid. Thanks so much for the fix.

auke-:
Thanks, no problem. I think the logic can be made more straightforward by using a ClassName instance instead of a list of strings and an index and compare it to a ClassName instance constructed from the typeSpecStack.

Furthermore, the algorithm could be made less greedy. The same goes for the method conflictsWithLocalName, which returns a fully qualified name for a nested class if the outer class has a name conflict.

I should read the specs (https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html) to understand all the naming rules...

swankjesse:
Yep. If you want to send another PR to simplify further, that's always welcome!

Issue #77, opened 2014-11-25
Imports sometimes incorrect when inheriting nested type of same name


eamonnmcmanus:
If you are generating a subclass of a class like this...

``` java
// Parent.java
class Parent {
  static class Optional {}
}

// Generated Child.java
class Child extends Parent {
  java.util.Optional<String> optionalString() {...}
}
```

...then it is not correct to import java.util.Optional and abbreviate to "Optional<String> optionalString()", because the inherited nested Optional takes precedence over the import statement. Since JavaWriter typically doesn't know anything about Parent in this case, there should be some way of telling it to fully-qualify a type even if it would otherwise be inclined to import it.

This is an actual case that occurs with AutoValue in code at Google. (AutoValue doesn't use JavaWriter and can't as long as this bug remains.)

adriancole:
In the dagger variant of JavaWriter, there's a concept of an originating element, which holds context about a generated type's parent.

That said, there was a bug in the dagger implementation, as this context was ignored for enclosed types. Thanks to @JakeWharton for pointing out this issue is very similar.

Anyway, here's the fix to dagger's JavaWriter and tests to go along with it.
https://github.com/google/dagger/pull/99

JakeWharton:
We have originating element as well but at a much more granular level.
While that may be the fix for Dagger, you cannot assume anything about the
relationship of the originating element to the generated type since it's a
completely separate domain.

adriancole:
I see what you mean: Originating Element is not the right concept to stick
to, even if it happened to have the type data I needed.
http://docs.oracle.com/javase/6/docs/api/javax/annotation/processing/Filer.html

Guess I know what I'm also doing tomorrow :P

JakeWharton:
This needs some thought, but I think we have two choices:

 * A concept similar to the explicit imports where you can inform the `TypeWriter` of additional type structure. In this example you would iterate the type hierarchy of the element used to specify the superclass to inform nested types.
 * Make `ClassName` aware of nested types. I don't know how feasible this is, I'm just shooting from the hip here. This way when you create the `ClassName` from the `TypeMirror` you'd get all nested types as well.

I haven't thought about either of these at all. Just spitballing.

swankjesse:
Right now we also explicitly import from `java.lang` because we don't know if the local package overrides those imports. Which means that 99% of the time, we're generating unnecessary imports.

[...]

PR #323, opened 2015-09-01
Use least-qualified names for conflicting class names

auke-:
After the quick fix for conflicting names (#320) this is my first shot at making name resolution more straight forward.

There are a couple of simple scenarios:
- same class
Always use the simple name, e.g. x.y.A referenced from x.y.A resolves to A and x.y.A.B referenced from x.y.A.B resolves to B.

- parent class
Always use the simple name, e.g. x.y.A.B referenced from x.y.A.B.C resolves to B and  x.y.A referenced from x.y.A.B.C resolves to A.

- nested class
Always use the qualified name starting at the nesting, e.g. x.y.A.B referenced from x.y.A resolves to B and x.y.A.B.C referenced from x.y.A resolves to B.C.

And some complex scenarios:
- nested class of a parent class
Use the qualified name starting at the nesting, e.g. x.y.A.D referenced from x.y.A.B.C resolves to D.

- other class in the same package
Use the qualified name, e.g. x.y.B.C referenced from x.y.A. resolves to B.C.

- other class in another package
Use the qualified name and import top-level name, e.g. x.y.z.B.C referenced from x.y.A. resolves to B.C and imports x.y.z.B.

If (in the complex scenarios) the top-level name of the resolved name conflicts with another name (same class, parent, member, member of a parent, imported class or referenced class) the name of the enclosing class is used. If the top-level class conflicts, the fully qualified name is used (package and class names).

There are some issues with my current approach:
- Information about inherited classes and interfaces isn't available (see #77).
- Type information of annotated classes and classes implementing or extending classes isn't available on the type stack (should this be changed in TypeSpec.emit?).
- I'm not sure if the methods nestedName and isNestedClass should be methods of ClassName or CodeWriter.
- Not all scenarios are covered by tests.
- Scenarios with anonymous nested classes aren't covered by tests.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
+
 import javax.lang.model.element.Modifier;
 
 import static com.squareup.javapoet.Util.checkArgument;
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +273,158 @@ private void emitLiteral(Object o) throws IOException {
...
   String lookupName(ClassName className) {
-    // Different package than current? Just look for an import.
-    if (!className.packageName().equals(packageName)) {
-      if (conflictsWithLocalName(className)) {
-        return className.toString(); // A local name conflicts? Use the fully-qualified name.
+
+    // Create ClassName from current stack
+    ClassName stackContext = null;
+    for (TypeSpec typeSpec : typeSpecStack) {
+        if (typeSpec.name == null) {
```
JakeWharton: can you reformat this with two-space indentation?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +273,158 @@ private void emitLiteral(Object o) throws IOException {
    * names visible due to inheritance.
    */
   String lookupName(ClassName className) {
-    // Different package than current? Just look for an import.
-    if (!className.packageName().equals(packageName)) {
-      if (conflictsWithLocalName(className)) {
-        return className.toString(); // A local name conflicts? Use the fully-qualified name.
+
+    // Create ClassName from current stack
```
JakeWharton: Just as a heads up, in general, [comments should be sentences][1]. All your comments are missing periods. Seems minor, I know, but the linked article points out some compelling arguments for it. Plus it help reinforce good commenting habit.

 [1]: nedbatchelder.com/blog/201401/comments_should_be_sentences.html

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +273,158 @@ private void emitLiteral(Object o) throws IOException {
...
+        } else {
+            stackContext = stackContext.nestedClass(typeSpec.name);
+        }
+    }
+
+    if (stackContext == null) {
+      // There is no stack context, e.g. annotation or interface on top-level classes
+      if (className.packageName().equals(packageName)) {
+        return Util.join(".", className.simpleNames());
+      } else {
```
JakeWharton: You don't need `else` after `return`, and it makes the presence of this comment not seem so strange.
```java
if (foo) {
  return "foo";
}
// There's no way to know!
```

swankjesse:
Is there a simpler way to implement this by working backwards?

Start with the fully qualified name `A.B.C.D`.
Drop the prefix and attempt to resolve `B.C.D`.
Does that resolve to `A.B.C.D` ? If no, return the full path. If yes, repeat the process.

swankjesse:
(we'd need a `resolve()` function that would take a partial name and return a full name)

auke-:
The simple scenarios (same class, nested class or parent class) are relatively easy to detect and handled, because the names can't conflict with another name in the same context.

For the other scenarios there are two issues: what is the shortest relative reference and does it conflict with another name in the current context.

I don't see how working backwards makes it easier, but that more because I don't  see how that would work.

auke-:
I've rebased the commit onto master. I'm looking for a way to get this merged. Can you suggest how I should split this up in smaller parts, make the logic more straightforward or change this commit in another way to make it easier to review and merge? Thanks

swankjesse:
I want to make an attempt at a resolution-based implementation to see if what I'm imagining will actually work. If it doesn't (it very likely won't!) I'll merge this! Sorry for the hold up.

auke-:
No problem. Note that there is (at least) one issue with my patch:
* Type information of annotated classes and classes implementing or extending classes isn't available on the type stack

This means that there is no way to detect name conflicts for interfaces:
```
package foo;
class Collection implements java.util.Collection {
...
}
```

Or annotations:
```
package foo;
@java.lang.FunctionalInterface
interface FunctionalInterface {
...
}
```

This is only an issue for top-level classes (not an unusual case) and could be addressed in TypeSpec.emit (invoke codeWriter.pushType before writing annotations and interfaces).

swankjesse:
@auke- here's resolution-based importing. https://github.com/square/javapoet/pull/331

auke-:
I haven't had time to look into your solution until now and it seems you fixed the issue. The latest snapshot version generates a lot cleaner imports and class-references on our codebase. Thanks!

PR #324, opened 2015-09-14
Add logic to avoid infinite recursion in TypeName.get(TypeMirror) …

### LINKED ISSUE #237: StackOverflowError: TypeName.get() for "Foo<T extends Foo<T>>"
While experimenting with JavaPoet I found that  a ```StackOverflowError``` occurs when calling ``TypeName.get(typeMirror)`` for a TypeMirror that represents the following class declaration which has a recursive type variable. 

```java
public class Foo<T extends Foo<T>> {}
```

And here is the stacktrace:-
```
Caused by: java.lang.StackOverflowError
        at javax.lang.model.util.AbstractTypeVisitor6.<init>(AbstractTypeVisitor6.java:69)
        at javax.lang.model.util.SimpleTypeVisitor6.<init>(SimpleTypeVisitor6.java:103)
        at javax.lang.model.util.SimpleTypeVisitor7.<init>(SimpleTypeVisitor7.java:80)
        at com.squareup.javapoet.TypeName$1.<init>(TypeName.java:155)
        at com.squareup.javapoet.TypeName.get(TypeName.java:155)
        at com.squareup.javapoet.TypeVariableName.get(TypeVariableName.java:88)
        at com.squareup.javapoet.TypeName$1.visitTypeVariable(TypeName.java:195)
        at com.squareup.javapoet.TypeName$1.visitTypeVariable(TypeName.java:155)
        at com.sun.tools.javac.code.Type$TypeVar.accept(Type.java:1495)
        at com.squareup.javapoet.TypeName.get(TypeName.java:155)
```

Class Foo in isn't something I can change in this case, so it would be great if JavaPoet could handle this. Please also let me know if there are any workarounds in the mean time.

eamonnmcmanus:
whenn the TypeMirror includes a recursive type bound. Fixes #237.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ArrayTypeName.java
+++ b/src/main/java/com/squareup/javapoet/ArrayTypeName.java
@@ -18,6 +18,10 @@ package com.squareup.javapoet;
 import java.io.IOException;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.type.ArrayType;
 
 import static com.squareup.javapoet.Util.checkNotNull;
@@ -54,7 +58,12 @@ public final class ArrayTypeName extends TypeName {
 
   /** Returns an array type equivalent to {@code mirror}. */
   public static ArrayTypeName get(ArrayType mirror) {
-    return new ArrayTypeName(get(mirror.getComponentType()));
+    return get(mirror, new HashMap<TypeParameterElement, TypeVariableName>());
+  }
+
+  static ArrayTypeName get(
+      ArrayType mirror, Map<TypeParameterElement, TypeVariableName> typeVariables) {
+    return new ArrayTypeName(get(mirror.getComponentType(), typeVariables));
   }
 
   /** Returns an array type equivalent to {@code type}. */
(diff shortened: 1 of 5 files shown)


src/main/java/com/squareup/javapoet/ArrayTypeName.java
```diff
@@ -54,7 +58,12 @@ public static ArrayTypeName of(Type componentType) {
 
   /** Returns an array type equivalent to {@code mirror}. */
   public static ArrayTypeName get(ArrayType mirror) {
-    return new ArrayTypeName(get(mirror.getComponentType()));
+    return get(mirror, new HashMap<TypeParameterElement, TypeVariableName>());
```
swankjesse: please LinkedHashMap. Never HashMap!

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -152,6 +156,11 @@ CodeWriter emit(CodeWriter out) throws IOException {
 
   /** Returns a type name equivalent to {@code mirror}. */
   public static TypeName get(TypeMirror mirror) {
+    return get(mirror, new HashMap<TypeParameterElement, TypeVariableName>());
```
swankjesse: Same.

swankjesse:
Brilliant. Thanks!

swankjesse:
Merged manually with the `LinkedHashMap` fix.

PR #331, opened 2015-09-29
Resolve-based import resolution.

swankjesse:
This introduces a new API that given a string, attempts to match it to
a class name. Then it uses that new API to discover the best name to
use when importing another type.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -27,12 +27,12 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.TreeMap;
 import javax.lang.model.element.Modifier;
 
 import static com.squareup.javapoet.Util.checkArgument;
 import static com.squareup.javapoet.Util.checkNotNull;
 import static com.squareup.javapoet.Util.checkState;
+import static com.squareup.javapoet.Util.join;
 
 /**
  * Converts a {@link JavaFile} to a string suitable to both human- and javac-consumption. This
@@ -47,8 +47,8 @@ final class CodeWriter {
   private boolean comment = false;
   private String packageName;
   private final List<TypeSpec> typeSpecStack = new ArrayList<>();
-  private final Map<ClassName, String> importedTypes;
-  private final Set<ClassName> importableTypes = new LinkedHashSet<>();
+  private final Map<String, ClassName> importedTypes;
+  private final Map<String, ClassName> importableTypes = new LinkedHashMap<>();
   private final Set<String> referencedNames = new LinkedHashSet<>();
   private boolean trailingNewline;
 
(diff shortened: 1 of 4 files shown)


src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -228,6 +228,27 @@
...
+            .build())
+        .build()
+        .toString();
+    assertThat(source).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import com.squareup.wire.Message;\n"
+        + "\n"
+        + "class Taco extends Message {\n"
+        + "  class Builder extends Message.Builder {\n"
```
JakeWharton: Yes. So much yes.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +272,80 @@ private void emitLiteral(Object o) throws IOException {
...
+      referencedNames.add(className.topLevelClassName().simpleName());
+      return join(".", className.simpleNames());
     }
 
-    // Look for the longest non-conflicting common prefix, which we can omit.
-    List<String> classNames = className.simpleNames();
-    int prefixLength = nonConflictingCommonPrefixLength(classNames);
-    if (prefixLength == classNames.size()) {
-      return className.simpleName(); // Special case: a class referring to itself!
+    // We'll have to use the fully-qualified name. Mark the type as importable for a future pass.
```
JakeWharton: Aren't these contradictory?

swankjesse: We actually execute `CodeWriter` 2x per file. On the first pass we don't import anything, but we do remember what we encountered. On the second pass, great success.

src/test/java/com/squareup/javapoet/JavaFileTest.java
swankjesse: Particularly noteworthy: with this change it's now pretty trivial to turn on and off nested-class imports. So we could import `com.squareup.wire.Message.Builder` directly; whereas we couldn't as easily before.

The _catch_ with that is that if we turn it on, it's not very smart, and so we always get the deepest types. I think the current, top-level-policy mostly works.

src/main/java/com/squareup/javapoet/CodeWriter.java
JakeWharton: Got it. Expanding the diff to read the Javaodc helped as well.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +272,80 @@ private void emitLiteral(Object o) throws IOException {
...
     }
-    return false;
   }
 
   /**
-   * Returns the common prefix of {@code classNames} and the current nesting scope. For example,
-   * suppose the current scope is {@code AbstractMap.SimpleEntry}. This will return 0 for {@code
-   * List}, 1 for {@code AbstractMap}, 1 for {@code AbstractMap.SimpleImmutableEntry}, and 2 for
-   * {@code AbstractMap.SimpleEntry} itself.
+   * Returns the class referenced by {@code simpleNames}, using the current nesting context and
```
JakeWharton: nit: `simpleName` (no 's')

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -272,106 +272,80 @@ private void emitLiteral(Object o) throws IOException {
...
+   * imports.
+   *
+   * TODO(jwilson): also honor superclass members when resolving names.
```
JakeWharton: `//`-style comment? Otherwise this shows up in Javadoc and people's IDEs. Unless you're going to fix before release.

src/main/java/com/squareup/javapoet/CodeWriter.java
swankjesse: Fixed.

src/main/java/com/squareup/javapoet/CodeWriter.java
swankjesse: Fixed.

src/test/java/com/squareup/javapoet/JavaFileTest.java
JakeWharton: Yeah. I think it's reasonable, and probably is what you want more cases than not versus the opposite behavior.

Issue #57, opened 2014-04-18
Support Java 8 things


JakeWharton:
Some failing tests:
```java
  @Test public void interfaceStaticMethod() throws IOException {
    javaWriter.emitPackage("com.squareup");
    javaWriter.beginType("com.squareup.Foo", "interface", EnumSet.noneOf(Modifier.class));
    javaWriter.beginMethod("Foo", "empty", EnumSet.of(STATIC));
    javaWriter.emitStatement("return new Foo() {}");
    javaWriter.endMethod();
    javaWriter.endType();
    assertCode(""
        + "package com.squareup;\n"
        + "\n"
        + "interface Foo {\n"
        + "  static Foo empty() {\n"
        + "    return new Foo() {};\n"
        + "  }\n"
        + "}\n");
  }

  @Test public void interfaceDefaultMethod() throws IOException {
    javaWriter.emitPackage("com.squareup");
    javaWriter.beginType("com.squareup.Foo", "interface", EnumSet.noneOf(Modifier.class));
    javaWriter.beginMethod("String", "asString", EnumSet.of(DEFAULT));
    javaWriter.emitStatement("return toString()");
    javaWriter.endMethod();
    javaWriter.endType();
    assertCode(""
        + "package com.squareup;\n"
        + "\n"
        + "interface Foo {\n"
        + "  default String asString() {\n"
        + "    return toString();\n"
        + "  }\n"
        + "}\n");
  }
```

kenzierocks:
Can we get this going? I'd be happy to PR it if needed, just tell me what you need.

swankjesse:
@kenzierocks what do you want specifically? Default methods? Static interface methods? Lambdas? Adding new modifier rules should be straightforward if you include sufficient tests.

Lambdas are already possible, as you express 'em as code strings.

kenzierocks:
I'm just submitting a PR for default methods, except I'm not sure where to add the tests. Advise me on the PR please :)

kenzierocks:
This might be complete? https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L760-L801 seems to cover what this issue was for.

JakeWharton:
I'd say so. Only thing left would be method references using MemberRef, but that's another issue.

sormuras:
@JakeWharton Good point. Will add it to #350 :smile:

PR #334, opened 2015-10-07
Added test for static methods in an interface

sormuras:
#57

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -44,7 +44,6 @@ import org.mockito.Mockito;
 
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.fail;
-import static org.junit.Assume.assumeTrue;
 
 @RunWith(JUnit4.class)
 public final class TypeSpecTest {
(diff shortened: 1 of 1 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -756,6 +756,28 @@ public void classCannotHaveDefaultMethods() throws Exception {
   }
 
   @Test
+  public void interfaceStaticMethods() throws Exception {
+    assumeTrue(isJava8());
```
JakeWharton: This shouldn't be needed, the resulting interfaces are not compiled.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -756,6 +756,28 @@ public void classCannotHaveDefaultMethods() throws Exception {
...
+    TypeSpec bar = TypeSpec.interfaceBuilder("Tacos")
+        .addMethod(MethodSpec.methodBuilder("test")
+            .addModifiers(Modifier.PUBLIC, Modifier.valueOf("STATIC"))
```
JakeWharton: This can just be `Modifier.STATIC` directly

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sormuras: True. 
I removed all other assume-statements in that test class as well.

PR #336, opened 2015-10-11
Add copy() method on NameAllocator

jbgi:
Useful to refine a NameAllocator for use in an inner scope code block.


swankjesse:
Neat! Wanna rename to clone() and implement Cloneable? That's my preference.

jbgi:
done. I usually try to stay away from Cloneable complexity. I guess this is OK in this case: the class does not extend another and is final.

jbgi:
should not have force pushed. opening a new pull-request.

PR #338, opened 2015-10-11
Add clone() method on NameAllocator

jbgi:
Useful to refine a NameAllocator for use in an inner scope code block.


swankjesse:
Brilliant. Thanks!

PR #339, opened 2015-10-12
AnnotationSpec.get(Annotation) added.

sormuras:
As *requested* by Jesse here http://stackoverflow.com/questions/33055732/compare-javapoet-parameterspec-type-with-java-8-annotatedtype - find an AnnotationSpec.get(Annotation) proposal in this PR.

### DIFF
--- a/src/main/java/com/squareup/javapoet/NameAllocator.java
+++ b/src/main/java/com/squareup/javapoet/NameAllocator.java
@@ -72,10 +72,24 @@ import static com.squareup.javapoet.Util.checkNotNull;
  * The underscore is appended to {@code sb} to avoid conflicting with the user-supplied {@code sb}
  * property. Underscores are also prefixed for names that start with a digit, and used to replace
  * name-unsafe characters like space or dash.
+ *
+ * <p>When dealing with multiple independent inner scopes, use a {@link #clone()} of the
+ * NameAllocator used for the outer scope to further refine name allocation for a specific inner
+ * scope.
  */
-public final class NameAllocator {
-  private final Set<String> allocatedNames = new LinkedHashSet<>();
-  private final Map<Object, String> tagToName = new LinkedHashMap<>();
+public final class NameAllocator implements Cloneable {
+  private final Set<String> allocatedNames;
+  private final Map<Object, String> tagToName;
+
+  public NameAllocator() {
+    this(new LinkedHashSet<String>(), new LinkedHashMap<Object, String>());
+  }
+
+  private NameAllocator(LinkedHashSet<String> allocatedNames,
+                        LinkedHashMap<Object, String> tagToName) {
+    this.allocatedNames = allocatedNames;
+    this.tagToName = tagToName;
+  }
 
   public String newName(String suggestion, Object tag) {
     checkNotNull(suggestion, "suggestion");
@@ -121,4 +135,19 @@ public final class NameAllocator {
     }
     return result;
   }
+
+  /**
+   * Create a deep copy of this NameAllocator. Useful to create multiple independent refinements
+   * of a NameAllocator to be used in the respective definition of multiples, independently-scoped,
+   * inner code blocks.
+   *
+   * @return A deep copy of this NameAllocator.
+   */
+  @Override
+  public NameAllocator clone() {
+    return new NameAllocator(
+        new LinkedHashSet<>(this.allocatedNames),
+        new LinkedHashMap<>(this.tagToName));
+  }
+
 }
--- a/src/test/java/com/squareup/javapoet/NameAllocatorTest.java
+++ b/src/test/java/com/squareup/javapoet/NameAllocatorTest.java
@@ -85,4 +85,17 @@ public final class NameAllocatorTest {
       assertThat(expected).hasMessage("unknown tag: 1");
     }
   }
+
+  @Test public void cloneUsage() throws Exception {
+    NameAllocator outterAllocator = new NameAllocator();
+    outterAllocator.newName("foo", 1);
+
+    NameAllocator innerAllocator1 = outterAllocator.clone();
+    assertThat(innerAllocator1.newName("bar", 2)).isEqualTo("bar");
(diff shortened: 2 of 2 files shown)


sormuras:
Some implementation notes:
* Two getters exposed, as with/without defaults use cases are valid, imho.
* Move MethodNameCompartor to Util class?
* The "rendering" of nested annotations in line 135ff is not "smart". Better use internal emitAnnotationValues here?
* The `addAnnotationValue` method could be enhanced to add `f`/`d`/etc type declaration syntax sugar.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,63 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+
+  private static Builder addAnnotationValue(Builder builder, String name, Object value) {
+    if (value instanceof Class<?>) return builder.addMember(name, "$T.class", value);
+    if (value instanceof Enum) return builder.addMember(name, "$T.$L", value.getClass(), value);
+    if (value instanceof String) return builder.addMember(name, "\"$L\"", value);
+    // return literal/toString anyway
+    return builder.addMember(name, "$L", value);
+  }
+
+  private static class MethodNameComparator implements Comparator<Method> {
```
swankjesse: Wanna make this an anonymous inner constant?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,63 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+  }
+
+  public static AnnotationSpec get(Annotation annotation, boolean includeDefaultValues) {
+    Builder builder = builder(annotation.annotationType());
+    try {
+      Method[] methods = annotation.annotationType().getDeclaredMethods();
+      Arrays.sort(methods, new MethodNameComparator());
+      for (Method method : methods) {
+        Object value = method.invoke(annotation);
+        if (!includeDefaultValues) {
```
swankjesse: why is this useful?

swankjesse: ...ahh, I see. if the annotation reports a default value you can omit it. Hmm. I think it's simpler to just universally include everything?

swankjesse: Let's keep your API as is. Interesting stuff.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,63 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+          }
+        }
+        if (value.getClass().isArray()) {
+          int length = Array.getLength(value);
+          for (int i = 0; i < length; i++) {
+            addAnnotationValue(builder, method.getName(), Array.get(value, i));
+          }
+          continue;
+        }
+        if (value instanceof Annotation) {
```
swankjesse: is this simpler as `if/else if/else` rather than `if/continue, if/continue/ last` ?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,63 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+            String s = nested.toString();
+            s = s.substring(s.indexOf('(') + 1, s.lastIndexOf(')'));
+            builder.addMember(method.getName(), "@$T($L)", nested.type, s);
+          }
+          continue;
+        }
+        addAnnotationValue(builder, method.getName(), value);
+      }
+    } catch (Exception e) {
+      throw new IllegalStateException("Reflecting " + annotation + " failed!", e);
```
swankjesse: RuntimeException? IllegalStateException means the call is allowed, just not at this time.

swankjesse:
This is awesome.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,62 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+    }
+    return builder.build();
+  }
+
+  private static Builder addAnnotationValue(Builder builder, String name, Object value) {
+    if (value instanceof Class<?>) return builder.addMember(name, "$T.class", value);
+    if (value instanceof Enum) return builder.addMember(name, "$T.$L", value.getClass(), value);
+    if (value instanceof String) return builder.addMember(name, "\"$L\"", value);
```
swankjesse: I think you want `$S` instead of `$L` here. Otherwise bad things happen if the string has quote characters.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: I prefer the control of "do something with current item, item done, next one". No need for "else".

swankjesse: Okay, but your preferences are wrong!

shipleys law

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Will revert my decision to make this an internal named class just minutes before the commit. If only Java 8 was the minimal target platform... Lambdas!

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Sure. Dooh.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: :+1: Your point.

PR #340, opened 2015-10-12
AnnotationSpec.get(Annotation) added.

sormuras:
Here we go, again. Squashing squashed me, again.
Will add open inline comments below.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -17,11 +17,17 @@ package com.squareup.javapoet;
 
 import java.io.IOException;
 import java.io.StringWriter;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import static com.squareup.javapoet.Util.checkNotNull;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,59 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+            addAnnotationValue(builder, method.getName(), Array.get(value, i));
+          }
+          continue;
+        }
+        if (value instanceof Annotation) {
+          AnnotationSpec nested = get((Annotation) value);
+          if (nested.members.isEmpty()) {
+            builder.addMember(method.getName(), "@$T", nested.type);
+          } else {
+            String s = nested.toString();
```
sormuras: Better use internal `emitAnnotationValues(...)` here or is grabbing the substring between `(...)` ok?
I had to split the nested toString result from the nested type, because otherwise the import mechanic does not kick in.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -98,6 +104,59 @@ private void emitAnnotationValues(CodeWriter codeWriter, String whitespace,
...
+      throw new RuntimeException("Reflecting " + annotation + " failed!", e);
+    }
+    return builder.build();
+  }
+
+  private static Builder addAnnotationValue(Builder builder, String name, Object value) {
+    if (value instanceof Class<?>) return builder.addMember(name, "$T.class", value);
+    if (value instanceof Enum) return builder.addMember(name, "$T.$L", value.getClass(), value);
+    if (value instanceof String) return builder.addMember(name, "$S", value);
+    // return literal/toString anyway
```
sormuras: Add `value instance of Float/float -> "$Lf", value`? Same for double, byte, short?

swankjesse: If the tests are happy, it's probably being inferred.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: You should be able to use `$L` for the nested annotation and have things Just Work.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: The tests don't compile the emitted annotation spec, yet. By "copy & pasting" some generated code I found out, that at least floats are to be marked with a `f`. See `float e() default 9.0f` in `@interface HasDefaultsAnnotation`.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: True.
Strange, my initial implementation used `$L` ... the next will too.

sormuras:
Anything left to do?

swankjesse:
Thanks!

sormuras:
Cheers!

Issue #341, opened 2015-10-13
Methods to access the state of Builders, for lazy members creation


leventov:
I want to design a control flow, when many "mutators" (the set of mutators is coined only in runtime, possibly different for each particular generation) update a `TypeSpec.Builder` one after another, adding some methods.

However, mutators may want some general members to be generated, like helper methods, constants, fields.

If the mutator interface is

```
interface Mutator {
    void generate(TypeSpec.Builder typeBuilder);
}
```

In the implementation I would like to generate general members as needed:

```
class FooMutator {
    void generate(TypeSpec.Builder typeBuilder) {
        if (!typeBuilder.fields.anyMatch(f -> f.name.equals("FOO")) {
            // generate FOO constant..
        }
        // use FOO constant
    }
}
```

I cannot do this, because Builders don't have accessible fields/accessor methods.

Of cause, I could keep the "already generated" state externally, like `void generate(TypeSpec.Builder typeBuilder, GenerationContext generationContext)`, but it seems to be unnecessary complication

swankjesse:
Yeah, we need this. Want to send a pull request?!

swankjesse:
...and if you do, just one type to begin, so we can make sure we agree on the details!

PR #342, opened 2015-10-13
Add fields() and methods() accessors to TypeSpec.Builder #341

leventov:


JakeWharton:
I wonder if we should just make the fields public. This would allow things like calling `Collections.sort` on the raw list with a custom comparator to more tightly control ordering (something that's been discussed a few times).

sormuras:
Making them public exposes the properties to unchecked modifications - but I like the flexibilty of having access to them.
May be some of the checks executed add building/adding stuff has to re-done on `toBuild()`

JakeWharton:
Ah, true. Good point.

leventov:
Yeah personally I would just make the fields public. I didn't do this in PR because if they were made private initially, apparently you decided that encapsulation is more important than agility.

swankjesse:
I like making ’em all public. It’s a _little_ annoying ’cause it prevents us from later changing say, the implementation of `TypeSpec.anonymousTypeArguments`. But that’s okay, and if we do we can rev the major version number.

swankjesse:
@leventov wanna send a change that does public fields? Maybe just for one type to start?

Issue #345, opened 2015-10-14
equals()/hashCode() in JavaPoet incorrect


sormuras:
```java
  @Test public void equalAndHash() {
    AnnotationSpec a = AnnotationSpec.builder(AnnotationC.class).build();
    AnnotationSpec b = AnnotationSpec.builder(AnnotationC.class).build();
    Assert.assertEquals(a, b);
    a = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
    b = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
    Assert.assertEquals(a, b);
  }
```
The last assertion fails as `TypeName` and `CodeBlock` don't override `equals()` and `hashCode()` properly, read not at all.

Both classes need something like:
```java
  @Override public int hashCode() {
    return toString().hashCode();
  }

  @Override
  public boolean equals(Object obj) {
    if (this == obj)
      return true;
    if (obj == null)
      return false;
    if (getClass() != obj.getClass())
      return false;
    return toString().equals(obj.toString());
  }
```

As both are immutable, caching the `toString()` result is an option, right?

sormuras:
This blocks a nice solution for http://stackoverflow.com/questions/33055732/compare-javapoet-parameterspec-type-with-java-8-annotatedtype

swankjesse:
I think using `toString()` for `CodeBlock.equals()` is pretty clever. It could also just compare its internal state. In either case, yes, let's implement these methods on these types.

sormuras:
Let's rename this issue to **equals()/hashCode() in JavaPoet** as tomorrow s/o wants to compare `MethodSpec`s or other types.

PR #347, opened 2015-10-15
equals() and hashCode() implemented

sormuras:
https://github.com/square/javapoet/issues/345

**Implementation outline**
* Used `toString().hashCode()` in all JavaPoet Spec and Name types instead of comparing the internal properties one-by-one.
* Tried to pre-calculate hash codes in constructors where possible.
  * In case of `TypeName` and its subclasses the hash code is calculated on-demand. Could be cached, if performance is an issue.
  * In case of `CodeBlock`, only two tests that assert proper syntax fail. May be rewrite those tests and pre-calculate hash code here, too?
* Add initial bunch of tests, should be extended someday.

Fixed a `NullPointerException` in `ClassName.java` along the way.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -43,10 +43,12 @@ import javax.lang.model.util.SimpleAnnotationValueVisitor7;
 public final class AnnotationSpec {
   public final TypeName type;
   public final Map<String, List<CodeBlock>> members;
+  private final int hashCode;
 
   private AnnotationSpec(Builder builder) {
     this.type = builder.type;
     this.members = Util.immutableMultimap(builder.members);
+    this.hashCode = toString().hashCode(); // always last constructor line!
   }
 
   void emit(CodeWriter codeWriter, boolean inline) throws IOException {
@@ -183,13 +185,17 @@ public final class AnnotationSpec {
   }
 
   @Override public boolean equals(Object o) {
-    return o instanceof AnnotationSpec
-        && ((AnnotationSpec) o).type.equals(type)
-        && ((AnnotationSpec) o).members.equals(members);
+    if (this == o)
+      return true;
+    if (o == null)
+      return false;
+    if (getClass() != o.getClass())
+      return false;
+    return hashCode == o.hashCode();
   }
 
   @Override public int hashCode() {
-    return type.hashCode() + 37 * members.hashCode();
+    return hashCode;
   }
 
   @Override public String toString() {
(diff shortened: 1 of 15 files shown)


src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -44,7 +44,7 @@ private ClassName(List<String> names) {
       checkArgument(SourceVersion.isName(names.get(i)), "part '%s' is keyword", names.get(i));
     }
     this.names = Util.immutableList(names);
-    this.canonicalName = names.get(0).isEmpty()
```
sormuras: `names.get(0)` can be `null` sometimes. Didn't find the cause, though. But if you remove the check `TypesTest` will throw a NPE.

sormuras:
...as the implementation is very similar in all Spec types, create an `AbstractSpec` and collect common functionality there?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -183,13 +185,17 @@ public Builder toBuilder() {
...
-    return o instanceof AnnotationSpec
-        && ((AnnotationSpec) o).type.equals(type)
-        && ((AnnotationSpec) o).members.equals(members);
+    if (this == o)
+      return true;
+    if (o == null)
+      return false;
+    if (getClass() != o.getClass())
+      return false;
+    return hashCode == o.hashCode();
```
swankjesse: this is disgusting and incorrect.

swankjesse: Much better is to cache toString() and compare that

sormuras: That's what it does - with only a 32 bit foot print and not computing `toString()` (with `emit()` etc) everytime `equals(Object)` is called. Somehow _disgusting_, sure.
Concerning correctness: i think, according to `String.hashCode()`s Javadoc, it should be nearly 100% correct.

But, if having the `toString()` result cached in each instance is a) not a memory concern and b) the runtime impact of `this.toString().equals(other.toString())` is not a performance concern, I'll rewrite the code accordingly.

kenzierocks: Unequal strings can have equal hashcodes, since there is >32 bits in many strings. Hashcode is never suitable for `equals`.

sormuras: Ja. Correct. Will convert that incorrect optimization to `String.equals()` calls.

sormuras:
What's your opinion regarding an abstract base class for Specs?

See https://github.com/sormuras/javapoet/blob/spec/src/main/java/com/squareup/javapoet/Spec.java - imagine the correct `equals()` implementation there.

sormuras:
https://github.com/square/javapoet/pull/349 takes over.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: `polygenelibricants` !

sormuras: [_Sorry, we found no matches for your search term(s) polygenelubricants._](http://dict.leo.org/ende/index_en.html#/search=polygenelubricants&searchLoc=0&resultOrder=basic&multiwordShowSingle=on)

PR #349, opened 2015-10-15
equals() and hashCode() implemented.

sormuras:
https://github.com/square/javapoet/issues/345
Take 2

If a common base class for all Specs, Names and CodeBlock is too intrusive/restrictive, I'll unroll the two methods into all types extending Base.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -327,7 +327,7 @@ final class CodeWriter {
 
     // Match the top-level class.
     if (typeSpecStack.size() > 0 && Objects.equals(typeSpecStack.get(0).name, simpleName)) {
-      return ClassName.get(packageName, simpleName);
+      return ClassName.get(getPackageNameNullSafe(), simpleName);
     }
 
     // Match an imported type.
@@ -340,11 +340,15 @@ final class CodeWriter {
 
   /** Returns the class named {@code simpleName} when nested in the class at {@code stackDepth}. */
   private ClassName stackClassName(int stackDepth, String simpleName) {
-    ClassName className = ClassName.get(packageName, typeSpecStack.get(0).name);
+    ClassName className = ClassName.get(getPackageNameNullSafe(), typeSpecStack.get(0).name);
     for (int i = 1; i <= stackDepth; i++) {
       className = className.nestedClass(typeSpecStack.get(i).name);
     }
     return className.nestedClass(simpleName);
+  }
+
+  private String getPackageNameNullSafe() {
+    return packageName == null ? "" : packageName;
   }
 
   /**
--- a/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
@@ -103,6 +103,17 @@ public final class AnnotationSpecTest {
 
   @Rule public final CompilationRule compilation = new CompilationRule();
 
+  @Test public void equalsAndHashCode() {
+    AnnotationSpec a = AnnotationSpec.builder(AnnotationC.class).build();
+    AnnotationSpec b = AnnotationSpec.builder(AnnotationC.class).build();
+    assertThat(a.equals(b)).isTrue();
+    assertThat(a.hashCode()).isEqualTo(b.hashCode());
+    a = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
+    b = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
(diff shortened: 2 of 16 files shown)


src/main/java/com/squareup/javapoet/Base.java
```diff
@@ -0,0 +1,19 @@
+package com.squareup.javapoet;
+
+abstract class Base {
```
swankjesse: Please no implementation inheritance. We use inheritance only for modeling!

src/main/java/com/squareup/javapoet/Base.java
```diff
@@ -0,0 +1,19 @@
+package com.squareup.javapoet;
+
+abstract class Base {
+
+  @Override public boolean equals(Object o) {
+    if (this == o)
```
swankjesse: nit: not consistent with our style

src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
```diff
@@ -103,6 +103,17 @@
 
   @Rule public final CompilationRule compilation = new CompilationRule();
 
+  @Test public void equalsAndHashCode() {
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -21,6 +21,17 @@
 import static org.junit.Assert.fail;
 
 public final class CodeBlockTest {
+  @Test public void equalsAndHashCode() {
```
swankjesse: Nice.

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.io.Serializable;
+import java.rmi.server.UID;
+import java.util.Comparator;
+import java.util.Set;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TypeNameTest {
```
swankjesse: great!

swankjesse:
Awesome. Mind copying equals/hashCode everywhere?

sormuras:
...on may way to copy'em here and there. :microphone:

src/main/java/com/squareup/javapoet/Base.java
sormuras: Inline return statements like:
```java
  @Override public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null) return false;
    return toString().equals(o.toString());
  }
```
or add curly braces?

sormuras:
`JavaFile` received the treatment as well.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -44,7 +44,7 @@ private ClassName(List<String> names) {
       checkArgument(SourceVersion.isName(names.get(i)), "part '%s' is keyword", names.get(i));
     }
     this.names = Util.immutableList(names);
-    this.canonicalName = names.get(0).isEmpty()
+    this.canonicalName = (names.get(0) == null || names.get(0).isEmpty())
```
swankjesse: Eep, does `null` actually happen? We should figure out where that's coming from and replace it with `""`.

swankjesse:
LGTM. We can deal with the null package thing now or later. Squash & I'll merge!

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -340,13 +340,17 @@ private ClassName resolve(String simpleName) {
...
   private ClassName stackClassName(int stackDepth, String simpleName) {
-    ClassName className = ClassName.get(packageName, typeSpecStack.get(0).name);
+    ClassName className = ClassName.get(getPackageNameNullSafe(), typeSpecStack.get(0).name);
     for (int i = 1; i <= stackDepth; i++) {
       className = className.nestedClass(typeSpecStack.get(i).name);
     }
     return className.nestedClass(simpleName);
   }
 
+  private String getPackageNameNullSafe() {
```
sormuras: This prevents the NPE that is thrown, when building nested `TypeSpec`s and the constructor of `TypeSpec` contains a `toString()` call as its last statement.

These 3 tests will fai in `TypeSpecTest.java`:
* `typeVariable()`
* `nestedClass()`
* `referencedAndDeclaredSimpleNamesConflict()`

sormuras:
Squashed.

src/main/java/com/squareup/javapoet/CodeWriter.java
swankjesse: Ahh yeah, that's broken! Thanks for the fix! https://github.com/square/javapoet/pull/351

Issue #317, opened 2015-08-03
MemberRef


swankjesse:
Would be interesting to have an accelerator for member references, like `someObject.instanceMethod()` or `someObject.instanceField` or `SomeEnum.CONSTANT` or `SomeType.staticMethod()` etc.

We'd need a new `MemberRef` type that would include either a TypeName (for static members) or a String name (for instance members).

Would be useful for static imports.

jbgi:
:+1: would be very useful.

sormuras:
What would a possible usage look like? Say, you want to write:
`TimeUnit.DAYS.convert(1701, TimeUnit.SECONDS)`

With `$A` (for anchor like in <a href=...) and `Ref` for the reference type:
```java
"$A(42, $A)", Ref.get(TimeUnit.class, TimeUnit.DAYS, "convert"), Ref.get(TimeUnit.class, TimeUnit.SECONDS)
```
instead of
```java
"$T.$L.convert(42, $T.$L)", TimeUnit.class, TimeUnit.DAYS, TimeUnit.class, TimeUnit.SECONDS
```
or, with `$E` (hint, hint):
```java
"$E.convert(42, $E)", TimeUnit.DAYS, TimeUnit.SECONDS
```

jbgi:
For methods I would want something in the like of ```Ref.get(ExecutableElement method, TypeMirror... typeArguments)```

sormuras:
@jbgi Talking about parameter types, do you? If yes, the parameter types are already included in the `ExecutableElement` instance, iirc.

Anyway, can you give me an example for `someObject.instanceMethod()`. Especially, how do you reference `someObject`. Via a string literal, via NameAllocator?
```java
"$A($A)", Ref.get("this", "equals"), Ref.get(TimeUnit.DAYS)
"$T v = $A", TypeName.get(TimeUnit.class), Ref.get("super", "getCurrentUnit")
```

jbgi:
@sormuras no type arguments. like when I want to generate a call to ```Collections.<String>emptyList()```

sormuras:
I understand. Never used type arguments in that place. Usually the types are infered like:
```java
List<String> strings = Collections.emptyList()
```

jbgi:
@sormuras yes I usually do that. there is cases where type inference is poor however, like when chaining methods:
```Collections.emptyList().stream().findFirst(); // does not compile```
Contrived example, but it does arise sometimes. I agree that the use case is not worth it.

[...]

Issue #73, opened 2014-11-25
Static imports


JakeWharton:
```java
public JavaWriter addStaticImport(Class<?> cls, String methodName) { }
```
```java
private final SetMultimap<ClassName, String> staticImports = LinkedHashMultimap.create();
```
Should be a bit more straightforward to emit since the logic for actually choosing which imports only applies to the current and parent contexts in the same file. A much smaller scope than regular import resolution.

hascode:
As I am writing a code generator for a framework that uses/offers a DSL - using a lot of static imports - I would really be delighted to see this feature being added to javapoet!

Keep up the good work! :+1:

swankjesse:
Definitely want this for Wire also.

swankjesse:
One tricky problem is that we'll need to detect the referenced method import in the code. We'd need to teach `CodeBlock` to look beyond its normal scope at things like `$T.identifier`, inspecting the string to find the identifier. If it does, it would omit `$T` and the dot that follows.

It gets very difficult if we favor in the possibility of name collisions. For example, if the current type declares a method that collides, or even _extends_ a class that has a method that collides, the static import is unsafe.

sormuras:
What do you like more?
 * a) Automatically inserted import static statements for all static referable members?
```java
JavaFile.builder(...).autoStaticImports(true)...
```
 * b) Explicitly named static referable members? See [addStaticImports](https://github.com/sormuras/javapoet/blob/ref/src/test/java/com/squareup/javapoet/MemberRefTest.java#L154)
```java
JavaFile.builder(...).addStaticImports(MemberRef.get(System.class.getMethod("gc")), ...)...
```

hascode:
@sormuras b) is my favourite, too :+1:

sormuras:
a) is #390
b) is #389
c) not implemented, yet

sormuras:
`implementes Closeable` :tada:

PR #350, opened 2015-10-15
MemberRef "$R" and static import support added.

sormuras:
#317 and #73 solved?

* ~~README.md needs some "$R"-love, read paragraph.~~ https://github.com/sormuras/javapoet/tree/ref#r-for-references-and-static-imports
* What about references to `MethodSpec`, `FieldSpec` and `TypeSpec.enumConstants`? Or are they covered by `$N` already?

### DIFF
--- a/src/main/java/com/squareup/javapoet/MemberRef.java
+++ b/src/main/java/com/squareup/javapoet/MemberRef.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+
+/**
+ * Member reference class.
+ *
+ * Use {@code $R} in format strings to insert member references into your generated code.
+ *
+ * @author Christian Stein
+ */
+public abstract class MemberRef implements Comparable<MemberRef> {
+
+  /**
+   * Defines all well-known member flavours, like {@code ENUM}, {@code FIELD} and {@code METHOD},
+   * that can be refered to.
+   */
+  public enum Kind { ENUM, FIELD, METHOD }
+
+  /** Simple getter using JavaPoet-model types only. */
+  public static MemberRef get(Kind kind, ClassName type, String name, Set<Modifier> modifiers,
+      TypeName... typeArguments) {
+    Util.checkNotNull(kind, "kind == null");
+    Util.checkNotNull(type, "type == null");
+    Util.checkNotNull(name, "name == null");
+    Util.checkNotNull(modifiers, "modifiers == null");
+    Util.checkNotNull(typeArguments, "typeArguments == types");
+    switch (kind) {
+    case ENUM:
+      return new EnumRef(type, name, modifiers);
+    case FIELD:
+      return new FieldRef(type, name, modifiers);
+    case METHOD:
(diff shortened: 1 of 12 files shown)


src/main/java/com/squareup/javapoet/MemberRef.java
```diff
@@ -0,0 +1,138 @@
...
+    Util.checkNotNull(types, "types == null");
+    TypeName type = TypeName.get(element.getEnclosingElement().asType());
+    String name = element.getSimpleName().toString();
+    return new MemberRef(type, name, someObject, TypeName.list(types));
+  }
+
+  public final TypeName type;
+  public final List<TypeName> arguments;
+  public final String name;
+  public final Object someObject;
```
swankjesse: I don't quite understand what this is.

src/main/java/com/squareup/javapoet/MemberRef.java
```diff
@@ -0,0 +1,138 @@
...
+    Util.checkNotNull(element, "element == null");
+    Util.checkNotNull(someObject, "someObject == null");
+    Util.checkNotNull(types, "types == null");
+    TypeName type = TypeName.get(element.getEnclosingElement().asType());
+    String name = element.getSimpleName().toString();
+    return new MemberRef(type, name, someObject, TypeName.list(types));
+  }
+
+  public final TypeName type;
+  public final List<TypeName> arguments;
```
swankjesse: Ooh, are these type parameters? Hmmm

swankjesse:
Do you mind sitting on this for a while until I have time to figure out how this is most useful?

The thing that's really tricky for me to reason about is how these things work in practice. Typically you need to dereference either a static thing `Roshambo.ROCK` or an instance thing `roshambo.name()`.

What we have here _works_ but I really want to explore the consequences of the design, including how it'll work for static imports, and chained references.

sormuras:
Rebased and I'm totally fine with exploring.

See #317 for some discussion around the topic.
And https://github.com/sormuras/javapoet/tree/ref#r-for-references for usage examples.

src/main/java/com/squareup/javapoet/MemberRef.java
sormuras: https://github.com/square/javapoet/issues/317#issuecomment-148513048

src/main/java/com/squareup/javapoet/MemberRef.java
sormuras: What is the line, you're referring to?

If it's line 73 `public final Object someObject;` then it's the call site, the target, the instance that the method is called on.

src/test/java/com/squareup/javapoet/MemberRefTest.java
```diff
@@ -0,0 +1,86 @@
...
+  @Test public void readmeExampleWithMethodChaining() throws Exception {
+    // get members
+    Method cbBuilder = CodeBlock.class.getMethod("builder"); // static
+    Method cbAdd = CodeBlock.Builder.class.getMethod("add", String.class, Object[].class);
+    Method cbBuild = CodeBlock.Builder.class.getMethod("build");
+    // create refs explicitly
+    TypeName tnCodeBlock = TypeName.get(CodeBlock.class);
+    MemberRef mrBuilder = MemberRef.get(cbBuilder, tnCodeBlock, new Type[0]);
+    MemberRef mrAdd = MemberRef.get(cbAdd, "");
+    MemberRef mrBuild = MemberRef.get(cbBuild, "");
```
sormuras: Overloaded `get` too much. Having to pass an empty array is not ... nice.

Will introduce more specific getters, like `getStaticMemberRef`, `getInstanceMemberRef`

src/test/java/com/squareup/javapoet/MemberRefTest.java
```diff
@@ -0,0 +1,86 @@
...
+    // wire up
+    CodeBlock block = CodeBlock.builder()
+        .add("$R()$R(\"$$R\", $R)$R()", mrBuilder, mrAdd, Thread.State.NEW, mrBuild)
```
sormuras: Method arguments are part of the format string. Makes the reference instance re-usable for various arguments.
The chaining `.`s are part the `$R` rendering with the empty literal `""` as a target. Needs some more thoughts and work.

sormuras:
#73 ... I'd expect that `JavaFile` takes an user-controlled list of member refs to import, right? Meaning, no auto-magic import guessing here.

sormuras:
Do you mind reviewing again? :beers: 

The implementation of #73 is a initial one - but works nicely at the test cases. Needs a ton o'tests more, though.

sormuras:
Rebased to javapoet/master.

sormuras:
In this context, [Java 8 method references](https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html) need some love/thoughts.

Without thinking in detail, you can't use `Class::member` as an input for `MemberRef.get(...)` as the actual target is hidden by the lambda construct.

sormuras:
Rebased on 1.4.0 release.

sormuras:
@swankjesse The MemberRef property `Set<Modifier> modifiers` could easily be reduced to a `boolean isStatic` flag. At the moment, only that information is needed and used by the emitting logic.

sormuras:
Rebased.

sormuras:
I'l remove the "static import" support from this PR within the next days. It`s only working via manual tagging and I have an auto-import in my mind, and already have-coded.

sormuras:
@Deprecated - use #388 and/or following instead.

PR #351, opened 2015-10-16
Change around how we handle null package names.

swankjesse:
Converting null to empty makes me uncomfortable. Just change the default
so we always have a package.


kenzierocks:
Neat trick :)

PR #354, opened 2015-10-22
AnnotationTypeName added.

sormuras:
#136 proposal.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
+++ b/src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public final class AnnotatedTypeName extends TypeName {
+
+  public static AnnotatedTypeName get(TypeName type, AnnotationSpec... annotations) {
+    return new AnnotatedTypeName(type, Arrays.asList(annotations));
+  }
+
+  public static AnnotatedTypeName get(Type type, Annotation... annotations) {
+    List<AnnotationSpec> specs = new ArrayList<>();
+    for (int i = 0; i < annotations.length; i++) {
+      specs.add(AnnotationSpec.get(annotations[i]));
+    }
+    return new AnnotatedTypeName(TypeName.get(type), specs);
+  }
+
+  public final TypeName type;
+  public final List<AnnotationSpec> annotations;
+
+  AnnotatedTypeName(TypeName type, List<AnnotationSpec> annotations) {
+    this.type = type;
+    this.annotations = Util.immutableList(annotations);
+  }
+
+  @Override CodeWriter emit(CodeWriter out) throws IOException {
+    if (!annotations.isEmpty()) {
+      for (AnnotationSpec annotation : annotations) {
+        annotation.emit(out, true);
+        out.emit(" ");
+      }
+    }
+    return type.emit(out);
+  }
+
+}
--- a/src/test/java/com/squareup/javapoet/TypeNameTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeNameTest.java
@@ -16,63 +16,97 @@
 package com.squareup.javapoet;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 
 import java.io.Serializable;
+import java.lang.annotation.Documented;
 import java.rmi.server.UID;
 import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
-
-import org.junit.Assert;
 import org.junit.Test;
+import com.squareup.javapoet.AnnotationSpecTest.AnnotationA;
+import com.squareup.javapoet.AnnotationSpecTest.AnnotationB;
 
 public class TypeNameTest {
 
   @Test public void equalsAndHashCodePrimitive() {
-    test(TypeName.BOOLEAN, TypeName.BOOLEAN);
-    test(TypeName.BYTE, TypeName.BYTE);
-    test(TypeName.CHAR, TypeName.CHAR);
-    test(TypeName.DOUBLE, TypeName.DOUBLE);
-    test(TypeName.FLOAT, TypeName.FLOAT);
-    test(TypeName.INT, TypeName.INT);
-    test(TypeName.LONG, TypeName.LONG);
-    test(TypeName.SHORT, TypeName.SHORT);
-    test(TypeName.VOID, TypeName.VOID);
+    assertEqualsHashCodeAndToString(TypeName.BOOLEAN, TypeName.BOOLEAN);
+    assertEqualsHashCodeAndToString(TypeName.BYTE, TypeName.BYTE);
+    assertEqualsHashCodeAndToString(TypeName.CHAR, TypeName.CHAR);
+    assertEqualsHashCodeAndToString(TypeName.DOUBLE, TypeName.DOUBLE);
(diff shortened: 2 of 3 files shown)


src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -69,10 +79,29 @@
...
 
+  @Test public void equalsAndHashCodeAnnotated() {
+    test(TypeName.BYTE.annotated(list()), TypeName.BYTE.annotated(list()));
+    test(TypeName.INT.annotated(list(Override.class)), TypeName.INT.annotated(list(Override.class)));
+    test(ClassName.get(Map.class).annotated(list(AnnotationA.class, AnnotationB.class)),
+        ClassName.get(Map.class).annotated(list(AnnotationA.class, AnnotationB.class)));
+    assertNotEquals(TypeName.CHAR, TypeName.CHAR.annotated(list(Documented.class)));
+  }
+
   private void test(TypeName a, TypeName b) {
```
swankjesse: maybe this wants to be `assertEqualsAndHashCode()`?  or something

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -69,10 +79,29 @@
...
-    Assert.assertEquals(a.toString(), b.toString());
+    assertEquals(a.toString(), b.toString());
     assertThat(a.equals(b)).isTrue();
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
   }
 
+  private List<AnnotationSpec> list(Class<?>... types) {
+    if (types.length == 0) {
```
swankjesse: this optimization probably isn't necessary

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -69,10 +79,29 @@
...
+      return Collections.emptyList();
+    }
+    List<AnnotationSpec> annotations = new ArrayList<>();
```
swankjesse: Should this just be `Arrays.asList(types)` ?

swankjesse:
This is great. Rebase & Squash & I'll merge

src/test/java/com/squareup/javapoet/TypeNameTest.java
sormuras: Removed.

src/test/java/com/squareup/javapoet/TypeNameTest.java
sormuras: Hm? Need `AnnotationSpec` instances here, not `java.lang.Class`.

src/test/java/com/squareup/javapoet/TypeNameTest.java
sormuras: Wanted the name to be in line with the other method names. Failed, though. Will try `@Test ... equalsAndHashCodeAnnotatedTypeName()`

Or are you referring to the name of the private `test` method?

sormuras: Ah, your comment is pointing to line 72. Okay, will rename as you wish. :+1:

swankjesse: 100 chars per line!

sormuras: :100: ? Then I'll call it `t()` ... :smile:

swankjesse: Seems like you need some more newline characters. Here's five:
```




```

sormuras: High five!

src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.util.List;
+
+public final class AnnotatedTypeName extends TypeName {
```
JakeWharton: It seems strange to have a public type with no factory methods in contrast to what the rest of the library does. We don't have something like
```java
public ParameterizedTypeName parameterized(List<TypeName> parameters) {
  // ...
}
```
on `TypeName` (maybe we should!).

At the very least I think a static factory is in order here which takes a `TypeName` base an varargs `AnnotationSpec`.

swankjesse: I buy that.

swankjesse: This type needs its own `equals()` and `hashCode()` and `toString()`

sormuras: Like https://github.com/square/javapoet/issues/136#issuecomment-150122774 - the static getter?

sormuras: Why its own equals() and hashCode() and toString()? Those inherited from `TypeName` (using the `toString()` from any derivation of `TypeName`) fits, imho.

sormuras:
Will stream-line the getters. Static and instance-ones in `TypeName`.
`equals`, `hashCode` and/or `toString` clean up should move into another task.

sormuras:
@JakeWharton The `parameterized(List<TypeName> parameters)` factory is better suited in `ClassName`. Or should we add a check for `instanceof ClassName` and cast it on the fly?

sormuras:
Squashed for review.

sormuras:
Fixed style, squashed for review.

swankjesse:
Thanks!

sormuras:
You're welcome. Now, off to #350 with both of you... :smile:

src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
```diff
@@ -0,0 +1,57 @@
...
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public final class AnnotatedTypeName extends TypeName {
```
swankjesse: OH CRAP, this shouldn’t `extends` TypeName. I completely missed that. It should be its own class that extends Object. Please fix!

swankjesse: This is my bad on review. This really, really shouldn't be a TypeName though. I'm tempted to roll your change back and ask you to resubmit another PR that fixes that...

sormuras:
Sure. Revert and I extend Object.

src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
sormuras: @swankjesse Before you roll back the changes, please explain me the difference between `ParameterizedTypeName` and `AnnotatedTypeName`. Why is the former okay extending `TypeName`, why the later not?

JakeWharton: This has to be a `TypeName`. Otherwise you cannot create types like `List<@NonNull String>`

sormuras: Good point, @JakeWharton - I'll submit a test case with all examples from #136 's description.

sormuras: Moving this discussion over to #136 - there're more examples and general thoughts on the use cases.

swankjesse: Ahhh. Ughh.

src/main/java/com/squareup/javapoet/AnnotatedTypeName.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public static AnnotatedTypeName get(TypeName type, AnnotationSpec... annotations) {
+    return new AnnotatedTypeName(type, Arrays.asList(annotations));
+  }
+
+  public static AnnotatedTypeName get(Type type, Annotation... annotations) {
+    List<AnnotationSpec> specs = new ArrayList<>();
+    for (int i = 0; i < annotations.length; i++) {
```
swankjesse: Should this return `type` as is if `annotations.length == 0` ?

sormuras: No. If you call this getter, you expect an `AnnotatedTypeName`, not just a `TypeName` instance.

A not-annotated-AnnotatedTypeName `toString()` representation may look like a simple `TypeName`s representation, but they are not equal in the meaning of `equals`.

Analog empty-collection behaviour: [ParameterizedTypeName.java#L71](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/ParameterizedTypeName.java#L71)

swankjesse: I don't think the two cases are equivalent. A `String` without any annotations is not different than a `String` without any annotations. The two instances will be `equals()` but they'll be of different types. That’s weird.

swankjesse:
https://github.com/square/javapoet/pull/358

PR #358, opened 2015-10-24
Revert "AnnotationTypeName added."

swankjesse:
This reverts commit a6716680d0328420b62ed4bd8ad2949b7f051fb9.

The commit is reasonable, I’m just not quite comfortable with some
model consequences where TypeNames and AnnotatedTypeNames
are intermixed.

In particular I want to decide whether we should just permit annotations
on _any_ type name, and whether we have methods like `withoutAnnotations()`
to strip ’em.


JakeWharton:
> whether we should just permit annotations on any type name

sounds reasonable actually

JakeWharton:
also LGTM

swankjesse:
Yeah, @sormuras sorry for thrashing on this. Wanna just add a `List<Annotation>` to `TypeName`? We’ll probably also want an API to strip the annotations, but that can come later!

sormuras:
See first part of this comment: https://github.com/square/javapoet/issues/136#issuecomment-150122774 ... was already partly implemented.

fup2 #136

PR #359, opened 2015-10-24
TypeName annotations added

sormuras:
#136

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -22,6 +22,7 @@ import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.lang.reflect.WildcardType;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


swankjesse:
Love it. LGTM

swankjesse:
(usual flattening request applies)

sormuras:
Applied.

And love it more than the former solution, too.

sormuras:
`TypeName` needs some more love... wait a sec.

sormuras:
Done.

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -87,14 +88,34 @@
...
   // Package-private constructor to prevent third-party subclasses.
-  TypeName() {
-    this(null);
+  TypeName(List<AnnotationSpec> annotations) {
+    this(null, annotations);
+  }
+
+  public TypeName annotated(AnnotationSpec... annotations) {
+    Util.checkNotNull(annotations, "annotations == null");
+    return new TypeName(keyword, Arrays.asList(annotations));
```
sormuras: This replaces any existing annotations with the new collection. If the user wants to merge the old and new annotations, it has to be done before calling this method. A nice side-effect of this solution, that an empty new collection strips any existing annotations.

PR #361, opened 2015-10-28
Fix a bug when a nested class and sibling class have the same name.

swankjesse:


JakeWharton:
LGTM✈✈✈

eamonnmcmanus:
I guess this doesn't work if the nested class is inherited from an ancestor. Probably it can't be made to work in the general case?

swankjesse:
@eamonnmcmanus yeah, we just need to hook up `CodeWriter` to something like `Elements`. The only part that’s interesting there is coming up with the right API.

PR #362, opened 2015-11-02
Add a method to AnnotationSpec to explicitly add a value member

Kapin:
When I started working with JavaPoet a bit, I didn't really have a strong understanding of the internals of annotations (the fact that the default member name is "value"), so I thought this could be useful for some people that just want to explicitly add an annotation like SupressWarnings or any other annotation that uses the "value" member.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -153,6 +153,7 @@ public final class AnnotationSpec {
     // return literal/toString anyway
     return builder.addMember(name, "$L", value);
   }
+
   public static AnnotationSpec get(AnnotationMirror annotation) {
     TypeElement element = (TypeElement) annotation.getAnnotationType().asElement();
     AnnotationSpec.Builder builder = AnnotationSpec.builder(ClassName.get(element));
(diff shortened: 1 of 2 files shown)


swankjesse:
The thing is, this might be confusing since value is such a meaningless name. What if, instead, there was a factory method to create an annotation with exactly one member, whose key was `value`?

kenzierocks:
I would prefer some sort of  `createSimpleAnnotation` method.

Kapin:
@swankjesse and @kenzierocks I completely agree and was worried about that. I think that's a much better design and I'll rework it.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -165,6 +166,15 @@ public static AnnotationSpec get(AnnotationMirror annotation) {
     return builder.build();
   }
 
+  public static AnnotationSpec createSimpleAnnotation(ClassName type, Object value) {
```
sormuras: May be enhanced to take `Object... values` for producing:
 * tagging annotation (without value) like simply `@AnnotationD`: `createSimpleAnnotation(AnnotationD.class)`
 * and array of values like `@AnnotationD({"a", "b", "c"})`: `createSimpleAnnotation(AnnotationD.class, "a", "b", "c")`

with 
```java
  @Retention(RetentionPolicy.RUNTIME)
  public @interface AnnotationD {
    String[] value() default {"x", "y", "z"};
  }
```

sormuras:
The private helper method
```java
private static Builder addAnnotationValue(Builder builder...)
```
should move to the `AnnotationSpec.Builder` class and be public. Just next to the `addMember` methods. Needs a better name, though.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
Kapin: That makes sense. I'll update tonight.

swankjesse:
Thanks! Closing because these ideas are now implemented.

sormuras:
> ...are now implemented.

Over here: [Poetry.annotation(...)](https://github.com/sormuras/poetry/blob/master/src/main/com/github/sormuras/poetry/Poetry.java#L80#L90) using [Poetry.value(...)] (https://github.com/sormuras/poetry/blob/master/src/main/com/github/sormuras/poetry/Poetry.java#L209)

PR #364, opened 2015-11-02
Build status icon added.

sormuras:


JakeWharton:
We don't really like these badges which is why they're all absent from our projects. It's an especially high-value location (first in the readme) for a low-value piece of information.

sormuras:
Never mind. (-:

PR #365, opened 2015-11-02
AnnotationSpec.get(Class<Annotation>) and Builder.addValue(String) added.

sormuras:
Implemented proposed factory methods from #362 here as: `AnnotationSpec.get`

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -104,6 +104,18 @@ public final class AnnotationSpec {
     codeWriter.emit(whitespace + "}");
   }
 
+  public static AnnotationSpec get(ClassName type, Object... values) {
+    AnnotationSpec.Builder builder = AnnotationSpec.builder(type);
+    for (Object value : values) {
+      builder.addValue(value);
+    }
+    return builder.build();
+  }
+
+  public static AnnotationSpec get(Class<? extends Annotation> type, Object... values) {
+    return get(ClassName.get(type), values);
+  }
+
   public static AnnotationSpec get(Annotation annotation) {
     return get(annotation, false);
   }
@@ -221,7 +233,7 @@ public final class AnnotationSpec {
      * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
      * the class of the given {@code value} object is not supported.
      */
-    private Builder addValue(String memberName, Object value) {
+    Builder addValue(String memberName, Object value) {
       if (value instanceof Class<?>) {
         return addMember(memberName, "$T.class", value);
       }
@@ -235,6 +247,15 @@ public final class AnnotationSpec {
         return addMember(memberName, "$Lf", value);
       }
       return addMember(memberName, "$L", value);
+    }
+
+    /**
+     * Conveniently adds an object value to {@code "value"} member to this annotation builder.
+     * @param value object value to add
+     * @see SuppressWarnings#value
+     */
+    public Builder addValue(Object value) {
+      return addValue("value", value);
     }
 
     public AnnotationSpec build() {
--- a/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
@@ -42,6 +42,11 @@ public final class AnnotationSpecTest {
     String value();
   }
 
+  @Retention(RetentionPolicy.RUNTIME)
+  public @interface AnnotationD {
+    String[] value() default "default";
+  }
+
   public enum Breakfast {
     WAFFLES, PANCAKES;
     public String toString() { return name() + " with cherries!"; };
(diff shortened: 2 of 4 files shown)


sormuras:
Rebased.

Any opinion regarding this PR?

swankjesse:
This is actually changing the definition of `addMember(String, String)` such that the 2nd argument goes from being a format string to being a String literal. The reason the test broke is that this change breaks API compatibility.

I like that this makes the calling code more compact, but I don't actually find the calling code to be more readable. In particular `AnnotationSpec.get(AnnotationC.class, "123", "456")` is particularly surprising.

The biggest difference is that this API favors being concise whereas our current API favors being extensible. Imagine you have a bunch of code that builds an annotation like `AnnotationC("123")`. Changing that code to add a 2nd member is more difficult in this API because you need to first expand out the current code to the full builder form.

swankjesse:
(I’m willing to be convinced on this, but the current design makes the API bigger without making it easier to understand.)

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -226,6 +228,26 @@ public Builder addMember(String name, CodeBlock codeBlock) {
       return this;
     }
 
+    public Builder addMember(String name, Object value) {
```
sormuras: Is changing this method signature to `addValue(String memberName, Object value)` more precise?
 * it's more distant to existing `addMember(String name, String format, Object... args)`
 * it's closer to the new `addValue()` with implicite `memberName="value"`

sormuras: Or make it package private - the static factory `AnnotationSpec get(Annotation annotation, boolean includeDefaultValues)` needs it, but it does not have to be public.

sormuras:
> I like that this makes the calling code more compact, but I don't actually find the calling code to be more readable. In particular AnnotationSpec.get(AnnotationC.class, "123", "456") is particularly surprising.

Depends on the actual annotation you are giving values to. The AnnotationC is not a good example here, imho. Look at these well-known annotations:
```java
// @Retention(RetentionPolicy.RUNTIME)
AnnotationSpec.get(Retention.class, RetentionPolicy.RUNTIME)

// @SuppressWarnings("javadoc")
AnnotationSpec.get(SuppressWarnings.class, "javadoc")

// @SuppressWarnings({"javadoc", "null"})
AnnotationSpec.get(SuppressWarnings.class, "javadoc", "null")
```
The Java code and the JavaPoet code are almost ... similar.

sormuras:
> The biggest difference is that this API favors being concise whereas our current API favors being extensible. Imagine you have a bunch of code that builds an annotation like AnnotationC("123"). Changing that code to add a 2nd member is more difficult in this API because you need to first expand out the current code to the full builder form.

True. But why prohibit some convenience, especially when using never-changing annotation? `SuppressWarnings` and many others will never have a 2nd member. Any many more, have no member at all. At the moment, you always have to write:
```
AnnotationSpec.builder(Deprecated.class).build()
```
instead of
```
AnnotationSpec.get(Deprecated.class)
```

And reaching for the "Java does it too" argument here: member name `value` is meant to be left out. When adding a second member, you need to prepend `value=` at your annotation anyway.

swankjesse:
Would it be better to encourage/build factories for common annotations?

```
  public static AnnotationSpec suppressWarnings(String... identifiers);
```

sormuras:
:smile: Been there.
https://github.com/sormuras/stash/blob/master/src/main/com/github/sormuras/stash/Generator.java#L175
https://github.com/sormuras/stash/blob/master/src/main/com/github/sormuras/stash/Generator.java#L184

Could be a solution regarding this generic factory issue. My main point to add this `AnnotationSpec.get` anyway, was my https://github.com/square/javapoet/pull/362#issuecomment-152974630 - want only to move the private static method to the builder. Thought, the idea from @Kapin was a good one.

sormuras:
I'll split this PR into two separate PRs:
* move `private static Builder addAnnotationValue(Builder builder, String name, Object value)` to the `Annotation.Builder` class, keeping it (package) private.
* add convenient `get(Object...)` factory

sormuras:
Squashed. 

Add more javadoc at the `AnnotationSpec.get` factories? With hint/encouragement to build static factories for complex annotations?

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -221,7 +233,7 @@ public Builder addMember(String name, CodeBlock codeBlock) {
      * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
      * the class of the given {@code value} object is not supported.
      */
-    private Builder addValue(String memberName, Object value) {
+    Builder addValue(String memberName, Object value) {
```
sormuras: `private` removed, some tests need access to this method

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1588,10 +1588,8 @@ private String toString(TypeSpec typeSpec) {
   @Test public void multipleAnnotationAddition() {
     TypeSpec taco = TypeSpec.classBuilder("Taco")
         .addAnnotations(Arrays.asList(
-            AnnotationSpec.builder(SuppressWarnings.class)
-                .addMember("value", "$S", "unchecked")
-                .build(),
-            AnnotationSpec.builder(Deprecated.class).build()))
+            AnnotationSpec.get(SuppressWarnings.class, "unchecked"),
+            AnnotationSpec.get(Deprecated.class)))
```
sormuras: This looks so awesome! :smile:

src/test/java/com/squareup/javapoet/AnnotationSpecTest.java
```diff
@@ -104,16 +109,61 @@
...
-    a = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
-    b = AnnotationSpec.builder(AnnotationC.class).addMember("value", "$S", "123").build();
+    a = AnnotationSpec.get(AnnotationC.class, "123");
+    b = AnnotationSpec.get(AnnotationC.class, "123");
     assertThat(a.equals(b)).isTrue();
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
   }
 
+  @Test public void createSimpleAnnotation() {
+    String pack = "com.squareup.javapoet.AnnotationSpecTest";
```
JakeWharton: Use `AnnotationSpecTest.class.getName()` to isolate from future [package renaming][1].

 [1]: http://jakewharton.com/java-interoperability-policy-for-major-version-updates/

sormuras: Sure.

Replace this one or all 26 occurances of `com.squareup.javapoet.AnnotationSpecTest` in this file? There're many hard-coded package names in here.

JakeWharton: Ignore them all in that case.

JakeWharton: They're in all the generated output and such. No point.

sormuras: Do like your blog post. Will JavaPoet follow that naming scheme as well, some day?

JakeWharton: If there's ever a v4, probably. The whole reason JavaPoet is called JavaPoet (and not JavaWriter v3) was because we needed a new package name to not clash with JavaWriter v2 in classpaths where both are being used!

sormuras:
Moved convenient get factories to [Poetry](https://github.com/sormuras/poetry/blob/master/src/main/com/github/sormuras/poetry/Poetry.java#L76#L86). PR #365 and #362 are closeable.

Issue #329, opened 2015-09-20
Too many arguments does not fail.


JakeWharton:
```java
result.addStatement("$1T $2N = new $1T($2S)", StringBuilder.class, "builder", "Hello!");
```
The final `$2S` should have been a `$3S`, but providing the third accidentally unused argument did not throw an exception.

kenzierocks:
Hmm, looks like the parameter validation needs another look over and a new test. [This](https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L1891) should have caught that.

sormuras:
```
String.format("%1$s %2$s = new %1$s(%2$s)", StringBuilder.class, "builder", "Hello!")
```
...doesn't throw an exception.

If I use explicite indexes, I see the objects passed as a pool of options. I might use all, I might use none. Sometimes, the format string is variable.

No bug for me.

sormuras:
@JakeWharton Is it an issue, or is it not?
As `java.lang.String.format`does not complain, JavaPoet should not either, imho.

JakeWharton:
I would rather be more strict than `String.format`, if it truly does not fail that case.

sormuras:
Added two more test scenarios in which more `$` markers are used than arguments are given. Also tried to be descriptive within the error message text.

eamonnmcmanus:
String.format might be used for things like log messages, where you don't want to cause your program to get an exception if an unusual condition causes it to use the String.format. I'd argue that that's not the case here. A JavaPoet user should have tests that exercise every format instance. Otherwise, how do you know that the generated code is correct? So in other words I agree with @JakeWharton.

sormuras:
I'm convinced, as well. With #375 merged, it now complains about unused indexed arguments.

sormuras:
Closeable.

JakeWharton:
Was fixed!

PR #374, opened 2015-12-09
Too many arguments does fail w/ frequency error message.

sormuras:
Fix for #329 -- even though `String.format()` does really not complain unused indexed arguments.


sormuras:
Nah, too many red lights... closing. Better next time.

PR #375, opened 2015-12-09
Unused indexed arguments do fail now.

sormuras:
Fix for #329 -- even though String.format() does really not complain unused indexed arguments.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -1914,7 +1914,19 @@ public final class TypeSpecTest {
       CodeBlock.builder().add("$1L $2L", "a", "b", "c");
       fail();
     } catch (IllegalArgumentException expected) {
-      assertThat(expected).hasMessage("unused arguments: expected 2, received 3");
+      assertThat(expected).hasMessage("unused argument: $3");
+    }
+    try {
+      CodeBlock.builder().add("$1L $1L $1L", "a", "b", "c");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("unused arguments: $2, $3");
+    }
+    try {
+      CodeBlock.builder().add("$3L $1L $3L $1L $3L", "a", "b", "c", "d");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("unused arguments: $2, $4");
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1914,7 +1914,19 @@ private String toString(TypeSpec typeSpec) {
...
       fail();
     } catch (IllegalArgumentException expected) {
-      assertThat(expected).hasMessage("unused arguments: expected 2, received 3");
+      assertThat(expected).hasMessage("unused arguments are marked as zero: [1, 1, 0]");
+    }
+    try {
+      CodeBlock.builder().add("$1L $2L $1L", "a", "b", "c");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("unused arguments are marked as zero: [2, 1, 0]");
```
swankjesse: I think this messaging is a little confusing. Maybe just say: `'arguments are unused"` ?

sormuras: Or list all unused arguments like `unused arguments: $1, S3`?

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1914,7 +1914,19 @@ private String toString(TypeSpec typeSpec) {
...
+      CodeBlock.builder().add("$1L $1L $1L", "a", "b", "c");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("unused arguments: $2, $3");
+    }
+    try {
+      CodeBlock.builder().add("$3L $1L $3L $1L $3L", "a", "b", "c", "d");
+      fail();
+    } catch (IllegalArgumentException expected) {
+      assertThat(expected).hasMessage("unused arguments: $2, $4");
```
swankjesse: dig.

swankjesse: ... do the dollar signs help or hurt?

swankjesse:
LGTM

src/test/java/com/squareup/javapoet/TypeSpecTest.java
JakeWharton: Hurt if you're not using positional (or rather, using implicitly positional) placeholders.

sormuras:
Squashed.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sormuras: Hurt not. The non-indexed (here relative) case is handled as before. No change there.

swankjesse:
Thanks!

Issue #197, opened 2015-01-22
Create a Builder class using javapoet


josketres:
Let's say I have the class `Person`

```java
package foo.bar;

public class Person {
    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}
```
and I would like to generate a `PersonBuilder` class.

```java
package foo.bar;

import java.lang.String;

public class PersonBuilder {
  private String name;

  public PersonBuilder withName(String name) {
    this.name = name;
    return this;
  }

  public Person build() {
    Person person = new Person();
    person.setName(name);
    return person;
  }
}
```
I tried to generate it using javapoet, but it was not possible to make `withName` return a `PersonBuilder` instance, because the class `PersonBuilder` doesn't exist yet.

```java
TypeSpec builder = TypeSpec.classBuilder("PersonBuilder")
    .addModifiers(Modifier.PUBLIC)
    .addField(FieldSpec.builder(String.class, "name", Modifier.PRIVATE).build())
    .addMethod(MethodSpec.methodBuilder("withName")
            .addModifiers(Modifier.PUBLIC)
             //.returns("foo.bar.PersonBuilder") 
             // how can I make this method return a PersonBuilder
            .addParameter(String.class, "name")
            .addStatement("this.name = name")
            .addStatement("return this")
            .build())
    .addMethod(MethodSpec.methodBuilder("build")
            .addModifiers(Modifier.PUBLIC)
            .returns(Person.class)
            .addStatement("Person person = new Person()")
            .addStatement("person.setName(name)")
            .addStatement("return person")
            .build())
    .build();
```
It would be nice to be able to specify the type using a fake Type
```java
...
    .addMethod(MethodSpec.methodBuilder("withName")
       .returns(TypePlaceholder.forType("foo.bar.PersonBuilder")) 
 ...
```

swankjesse:
The TypePlaceholder class you want exists, and it's called ClassName. Use it here & things should just work.

swankjesse:
(You'll want `$T` though.)

tbroyer:
I think the proposal with `TypePlaceholder.self` is to have a "fake" type that corresponds to the enclosing class, irrespective of where the `MethodSpec` will be used, irrespective of the final package of the `TypeSpec`; just a marker that would be resolved at `emit()` time to the type of `this`.

[...]

PR #382, opened 2015-12-17
Self type

sormuras:
Using existing functionality in `CodeWriter` with a simple TypeName subclass named `SelfName` does the trick.

Open ends
 * ~~Catch NoSuchElement/IndexOutOfBounds when stack is empty and report it to the user. Might happen in plain CodeBlocks.~~
 * ~~Introduce `TypeName.self()` factory instead of `SelfName.get`. Or even move the `get()`to `TypeSpec`?~~
 * Needs more tests.
 * Convince Jesse. :smile:

### DIFF
--- a/src/main/java/com/squareup/javapoet/SelfName.java
+++ b/src/main/java/com/squareup/javapoet/SelfName.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class SelfName extends TypeName {
+
+  private final int depth;
+
+  SelfName(List<AnnotationSpec> annotations) {
+    this(annotations, 0);
+  }
+
+  SelfName(List<AnnotationSpec> annotations, int depth) {
+    super(annotations);
+    this.depth = depth;
+  }
+
+  public SelfName(int depth) {
+    this(new ArrayList<AnnotationSpec>(), depth);
+  }
+
+  @Override public TypeName annotated(List<AnnotationSpec> annotations) {
+    return new SelfName(annotations);
+  }
+
+  @Override CodeWriter emit(CodeWriter out) throws IOException {
+ // ClassName className = out.resolve(name == null ? out.peekType().name : name);
+    ClassName className = out.resolve(out.peekType(depth).name);
+    return emitAnnotations(out).emitAndIndent(out.lookupName(className));
+  }
+
+}
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -121,6 +121,14 @@ final class CodeWriter {
     return this;
   }
 
+  TypeSpec peekType(int depth) {
+    checkState(!typeSpecStack.isEmpty(), "expected to be emitting a TypeSpec");
+    checkArgument(depth >= 0, "depth must not be negative, got %d", depth);
+    checkArgument(depth < typeSpecStack.size(), "depth %d exceeds nesting TypeSpec stack: %d",
+        depth, typeSpecStack.size() - 1);
+    return typeSpecStack.get(typeSpecStack.size() - 1 - depth);
+  }
+
   public void emitComment(CodeBlock codeBlock) throws IOException {
     trailingNewline = true; // Force the '//' prefix for the comment.
     comment = true;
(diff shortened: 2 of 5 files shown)


swankjesse:
Interesting!

src/main/java/com/squareup/javapoet/SelfName.java
```diff
@@ -0,0 +1,26 @@
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+public class SelfName extends TypeName {
```
JakeWharton: `final`

src/main/java/com/squareup/javapoet/SelfName.java
```diff
@@ -0,0 +1,26 @@
+package com.squareup.javapoet;
```
JakeWharton: copyright

src/main/java/com/squareup/javapoet/SelfName.java
```diff
@@ -0,0 +1,26 @@
...
+  public static SelfName get() {
+    return new SelfName(Collections.<AnnotationSpec>emptyList());
+  }
+
+  SelfName(List<AnnotationSpec> annotations) {
+    super(annotations);
+  }
+
+  @Override public TypeName annotated(List<AnnotationSpec> annotations) {
+    return new SelfName(annotations);
```
JakeWharton: Can you add a test or modify the test to exercise this code path? Maybe make something `@Nullable` or `@Nonnull`?

src/main/java/com/squareup/javapoet/SelfName.java
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  @Override CodeWriter emit(CodeWriter out) throws IOException {
+    ClassName name = out.resolve(out.peekType().name);
+    return emitAnnotations(out).emitAndIndent(out.lookupName(name));
```
sormuras: This feels like a circular call might arise. `lookupName` uses `resolve` internally. Can you think of a way simplifying this? I mean, get a `ClassName` instance from peeked `TypeName`.

src/main/java/com/squareup/javapoet/SelfName.java
sormuras: Sure.

But, in the next PR (popping up within the next minutes), the `annotated` methods will be touched anyway.

sormuras: Consider looking at/commenting on/merging #365? It'd help writing those tests (with less code). ;-)

src/main/java/com/squareup/javapoet/SelfName.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+public final class SelfName extends TypeName {
+
+  public static SelfName get() {
```
JakeWharton: This should probably be called `create()` as it returns a new instance each time. Or, since it's immutable it could even be `TypeName.SELF`.

JakeWharton: This was in your PR description!

sormuras: Aye.

How about making SelfName package private and only export via `TypeName.self()`? Or is a static field ok?

sormuras:
Squashed.

Can you please have a look at the `getClassNameFromTypeSpec(...)`-comment? I didn't find any recursion, but I'm not sure if it could happen.

swankjesse:
I don't think we want a self type like this. It won't work as the return value of Builder.build() !

sormuras:
Which Builder are you referring to, Jesse?

swankjesse:
Imagine you had a TypeSpec and a nested Builder class for that TypeSpec. The two types should be able to refer to each other, but I don’t think this approach permits that.

sormuras:
> ... each other ...

Bi-directional. Hm, going down the stack, therefore up in the hierachy, via a `TypeName.self(n)` permits at least on direction - but the other one, where `Outer` refers to `Inner` is hardly feasible. Not talking about `Inner1` referring to `Inner2`.

Literally *self* does not imply *each other*. One could keep `SelfName` as is and create `OtherName` using user-supplied markers later. I think the majority of use cases are handled by `SelfName`, like in linked types with "A#next of type A", or fluent type where "B#add() returns B".

sormuras:
`OtherName` is actually not needed. As stated in #197 and test case `nestedClasses()` in `TypeSpecTest` shows, a pre-defined `ClassName` works nicely here. It's possible to defer the resolving task to last `build()` and therefore exchange the pre-defined lines with ugly an inline marker like:
````java
    TypeSpec typeSpec = TypeSpec.classBuilder("Combo")
        .addField(resolveMarker("Combo.Taco"), "taco")
        ...
        .addType(TypeSpec.classBuilder(taco.simpleName())
            .mark("Combo.Taco")
```

Summarized, pre-defined `ClassName`s work perfectly for all `TypeSpec`-related pointers.

The proposed `SelfName` only tackles the special case, where you refer to *self*, i.e. the current TypeSpec in building scope.

sormuras:
The last commit contains a proof of concept peek depth functionality for demonstration purposes only. I still don't like the look of `new SelfType(2)`, nor a potential `TypeName.self(2)`.

sormuras:
@swankjesse before I close this PR as not generic enough, please comment on my last point from the discussion above:

This proposed `SelfName` ~~only~~ tackles the special case, where you refer to **self**, i.e. the current TypeSpec in building scope. I quote @tbroyer #197 #issuecomment-71085494
> I think the proposal with TypePlaceholder.self is to have a "fake" type that corresponds to the enclosing class, irrespective of where the MethodSpec will be used, irrespective of the final package of the TypeSpec; just a marker that would be resolved at emit() time to the type of **this**.

If you think a single `TypeName.SELF` is a worthy addition, I'll remove the depth parameter and add more tests here.

swankjesse:
I think we wanna drop it because it doesn't scale with the program. You start out with this and it saves some work, but later when you need to return a builder or whatever you basically have to undo this and go back to what we have today.

If Java had a self type, we'd want to model that. But it doesn't, it works the the way JavaPoet works.

sormuras:
Alright. So #197 can be closed.

PR #384, opened 2015-12-19
Refer to local TypeSpecs via simple name markers.

sormuras:
I followed the `OtherName` idea and propose this PR as a comparsion to #382 Self type.

See sample usage, with an empty marker for _self_ behaviour:
```java
@Test public void nestedClasses() throws Exception {
  TypeSpec typeSpec = TypeSpec.classBuilder("Combo")
      .addField(new MarkerName(""), "combo")
      .addField(new MarkerName("Taco"), "taco")
      .addField(new MarkerName("Chips"), "chips")
      .addType(TypeSpec.classBuilder("Taco")
          .addModifiers(Modifier.STATIC)
          .addField(ParameterizedTypeName.get(ClassName.get(List.class), new MarkerName("Topping")), "toppings")
          .addField(new MarkerName("Sauce"), "sauce")
...
```

 * The class name `MarkerName` is not final ... does `LocalName` smell better?
 * Of course, a `TypeName.SELF` constant can initialized with `new MarkerName("")`.

Had to tweak some exptected rendered class sources, as the new recursive resolve seems to evaluate differently. Find more comments inline.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -121,6 +121,11 @@ final class CodeWriter {
     return this;
   }
 
+  TypeSpec peekType() {
+    checkState(!typeSpecStack.isEmpty(), "expected to be emitting a TypeSpec");
+    return typeSpecStack.get(typeSpecStack.size() - 1);
+  }
+
   public void emitComment(CodeBlock codeBlock) throws IOException {
     trailingNewline = true; // Force the '//' prefix for the comment.
     comment = true;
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -320,18 +325,31 @@ private void importableType(ClassName className) {
...
 
+  ClassName resolve(String simpleName, TypeSpec typeSpec, int i) {
+    if (Objects.equals(typeSpec.name, simpleName)) {
+      return stackClassName(i, simpleName);
+    }
+    for (TypeSpec nested : typeSpec.typeSpecs) {
+      if (Objects.equals(nested.name, simpleName)) {
+        return stackClassName(i, simpleName);
+      }
+      return resolve(simpleName, nested, i);
```
sormuras: `i` needs some love, shift, side-step ... before it is passed to the recursive call. As is, it produces slightly different ClassNames, but everything is compilable.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -670,7 +669,7 @@ private boolean isJava8() {
         + "  }\n"
         + "\n"
         + "  static class Chips {\n"
-        + "    Taco.Topping topping;\n"
+        + "    Topping topping;\n"
```
sormuras: Here, the new resolved name is shorter and  still valid.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -839,7 +836,7 @@ public void interfaceDefaultMethods() throws Exception {
         + "\n"
         + "  com.squareup.donuts.Top externalTop;\n"
         + "\n"
-        + "  Bottom externalBottom;\n"
+        + "  com.squareup.donuts.Bottom externalBottom;\n"
```
sormuras: Import does not work as before here ... but the fall-back to fully-qualified class name is never wrong.

swankjesse:
Is this all to avoid providing the package name/

sormuras:
No, just to avoid explicitely define `ClassName`s before the `TypeSpec.Builder` statements. Like the `ClassName` lines removed from the start of the `nestedClass()` test case. It shows, that JavaPoet can resolve local types via simple names and provide a *self* functionality, in one go.

Actually, #382 is a nicer solution for the "reference the TypeSpec i'm building at the moment" - see my last 2-4 comments over there. :smiley:

swankjesse:
Discussed offline.

PR #385, opened 2015-12-21
Internal AnnotationSpec clean up.

sormuras:
Better place, no public API change. This PR was part of #365

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -127,7 +127,7 @@ public final class AnnotationSpec {
         }
         if (value.getClass().isArray()) {
           for (int i = 0; i < Array.getLength(value); i++) {
-            addAnnotationValue(builder, method.getName(), Array.get(value, i));
+            builder.addValue(method.getName(), Array.get(value, i));
           }
           continue;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -226,6 +216,27 @@ public Builder addMember(String name, CodeBlock codeBlock) {
       return this;
     }
 
+    /**
+     * Delegates to {@link #addMember(String, String, Object...)}, with parameter {@code format}
+     * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
+     * the class of the given {@code value} object is not supported.
+     */
+    private Builder addValue(String memberName, Object value) {
```
swankjesse: addMember? Value has a special meaning in annotations.

sormuras: *ganrflrrrg* ...then we're back to the naming conflict discovered by you. See #365 :smile:

swankjesse: Ahhh. Ship it.

sormuras:
Great. Now, I'll rebase #365 and leave it up to you, whether to merge or flush it...

swankjesse:
Yeah, do keep me busy!

sormuras:
> Well, I’ll do my very best

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -226,6 +216,27 @@ public Builder addMember(String name, CodeBlock codeBlock) {
...
+      if (value instanceof Class<?>) {
+        return addMember(memberName, "$T.class", value);
+      }
+      if (value instanceof Enum) {
+        return addMember(memberName, "$T.$L", value.getClass(), ((Enum<?>) value).name());
```
sormuras: Could be replaced by
```java
return addMember(memberName, "$R", value);
```
...if only #350 was merged. :dart:

PR #387, opened 2015-12-24
Reformatting headers of Overriden methods in CodeBlock.java

gnash48:
In CodeBlock.java, the equals(), hashCode(), and toString() methods had an "@Override" annotation that was located on the same line as their respective method headers. This was slightly confusing, and I changed the annotation location to be above the method headers in order to increase readability and conform to Java code conventions.


JakeWharton:
This is not our code style. See github.com/square/java-code-styles.

PR #388, opened 2015-12-28
MemberRef $R implemented.

sormuras:
Solves #317 similar to #350 - but focusses on the `$R` part, only. That means:

* ~~no static import support~~ already base feature of JavaPoet 1.5.0
* using `boolean isStatic` flag instead of `Set<Modifiers>`
* no more internal subclassing of `MemberRef`

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,77 @@ class HelloWorld {
 }
 ```
 
+### $R for References
+
+When generating code, you may point to already compiled elements like methods, fields
+and especially the `enum` constant fields. Enter `$R`:
+
+```java
+CodeBlock.builder().add("$R", Thread.State.NEW).build();
+```
+
+That produces the fully-qualified member name:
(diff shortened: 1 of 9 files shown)


sormuras:
Needs some 1.5.0 love: rebase, provide support around static import.

swankjesse:
I still don’t really think we need this. It’s already confusing deciding when to use `$N` and `$L`, and I think this makes it worse rather than better.

sormuras:
Discuss here - or over at #317 ?

sormuras:
Rebased to current HEAD and I have a chaining proposal under construction:
`MemberRef.get(FIELD).call(METHOD, arg0, arg1, ...).call(...))`

sormuras:
Moves eventually to Poetry -- `$L` is the old and new king! #404 for details.

PR #390, opened 2015-12-31
Import static (automagic) implemented, based on "$R"

sormuras:
Solves #317 and #73 - based on PR #388 using __automagically__ generated static import statements via an additional emitter phase.


sormuras:
Automagic will not prevail. Be gone!

Issue #391, opened 2016-01-07
TestFiler uses restricted API


sormuras:
> Access restriction: The type 'JavacPathFileManager' is not API (restriction on required library '[...]\jdk1.8.0_66\lib\tools.jar')

In case JDK 9 will be the platform... some day.

PR #393, opened 2016-01-07
Private source file object used, instead of com.sun.tools.javac types

sormuras:
Proposal for #391 - all tests remain green.

Is the internal "Source" class too simplistic? Am I missing something?

### DIFF
--- a/src/test/java/com/squareup/javapoet/TestFiler.java
+++ b/src/test/java/com/squareup/javapoet/TestFiler.java
@@ -15,12 +15,12 @@
  */
 package com.squareup.javapoet;
 
-import com.sun.tools.javac.nio.JavacPathFileManager;
-import com.sun.tools.javac.nio.PathFileManager;
-import com.sun.tools.javac.util.Context;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.file.FileSystem;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.spi.FileSystemProvider;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


src/test/java/com/squareup/javapoet/TestFiler.java
```diff
@@ -30,20 +31,28 @@
...
+import javax.tools.SimpleJavaFileObject;
 
 final class TestFiler implements Filer {
+  static class Source extends SimpleJavaFileObject {
+    private final OutputStream stream;
+    protected Source(URI uri, OutputStream stream) {
+      super(uri, Kind.SOURCE);
+      this.stream = stream;
+    }
+    @Override public OutputStream openOutputStream() {
```
swankjesse: I think this method probably wants to create a new stream on every invocation?

swankjesse: it's weird to have a live object like a stream as a field of a value object

swankjesse:
Nice!

src/test/java/com/squareup/javapoet/TestFiler.java
sormuras: As an instance of `Source` is only created by `TestFiler#createSourceFile`, I think it's okay to remember the already opened stream here. That instance will be (and according to all the green lights *is*) only used to collect the Java source lines once. I'm 99% sure, that `openOutputStream` is only called a single time per `TestFiler#createSourceFile`.

sormuras: I move the directory, file and stream creation into `Source`. Feels the 1% better.

PR #395, opened 2016-01-08
Added char/Character to AnnotationSpec.Builder.addValue logic

sormuras:
Missing case `(value instanceof Character)` with test fix added.


sormuras:
Lightning fast!

swankjesse:
Probably needs escaping if the character is `'`.

sormuras:
Escaping should be done be `$L` for every `$L` in CodeBlock, right?

swankjesse:
Nope, `$L` is very very literal.

sormuras:
True. PR #397 filed.

Issue #394, opened 2016-01-08
JavaFile.compile()


sormuras:
What do you think about an on-the-fly, in-memory, runtime ... compile feature?
```java
  JavaFile.builder(...).build().compile().newInstance()...
```
Initial implementation can be peeked here: [JavaFile.compile()](https://github.com/sormuras/javapoet/blob/poetry/src/main/java/com/squareup/javapoet/JavaFile.java#L212)
Exemplary usage: [compile() with annotation processor] (https://github.com/sormuras/javapoet/blob/poetry/src/test/java/com/squareup/javapoet/PoetryTest.java#L52)

sormuras:
Moves to [Poetry](https://github.com/sormuras/poetry)

PR #396, opened 2016-01-08
JavaFile.compile() added.

sormuras:
Proposal for #394

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -22,21 +22,30 @@ import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TreeSet;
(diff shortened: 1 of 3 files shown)


sormuras:
Cleaning up the imports and empty lines and adding a "compile and instantiate" variant...

swankjesse:
What would tools use this for?

Looking at the code I think this is interesting but doesn’t quite fit as public API in JavaPoet. In particular there’s a lot of decisions being made in here in terms of build paths, error messages, etc.

sormuras:
* Quick check if generated code compiles
* `JavaFile.Builder.build(boolean assertCompile)` could perform a silent compile and throw an exception if compilation fails
* Use instance of generated class in tests, mock-ups and even in script-like environments
* Pass custom annotation processor (generating more `JavaFile`s) and test the results without having to introduce a dependecy to `compile-testing.jar` or others. Just use the JDK provided compiler.

Build path and error messages could be parameterized - but if you want to do some fancy stuff, roll your own `JavaFile` compiler.

tbroyer:
Looks like something that could be useful (well, if you have a use-case for it) independently of JavaPoet; something that would ideally live in its own library (and it wouldn't even need to depend on JavaPoet IIUC, you'd just have to call `toJavaFileObject()` on your `JavaFile` and the lib would just expect a `JavaFileObject`, thus being completely independent of JavaPoet)

swankjesse:
Let’s leave this to a separate project.

PR #397, opened 2016-01-08
Escape special char values.

sormuras:
Fixes #395 if the value is a single ``'`` character.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -235,7 +235,11 @@ public final class AnnotationSpec {
         return addMember(memberName, "$Lf", value);
       }
       if (value instanceof Character) {
-        return addMember(memberName, "'$L'", value);
+        String literal = CodeWriter.stringLiteral(value.toString(), "");
+        literal = literal.substring(1, literal.length() - 1);
+        if (literal.equals("\\\"")) literal = "\"";
+        if (literal.equals("'")) literal = "\\'";
+        return addMember(memberName, "'$L'", literal);
       }
       return addMember(memberName, "$L", value);
     }
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -419,6 +419,10 @@ final class CodeWriter {
 
   /** Returns the string literal representing {@code data}, including wrapping quotes. */
   String stringLiteral(String value) {
+    return stringLiteral(value, indent);
+  }
+
+  static String stringLiteral(String value, String indent) {
     StringBuilder result = new StringBuilder();
     result.append('"');
     for (int i = 0; i < value.length(); i++) {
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -235,7 +235,8 @@ private Builder addValue(String memberName, Object value) {
         return addMember(memberName, "$Lf", value);
       }
       if (value instanceof Character) {
-        return addMember(memberName, "'$L'", value);
+        String format = Character.valueOf('\'').equals(value) ? "'\\\''" : "'$L'";
```
kenzierocks: Don't need to escape '\' in the string, though it might add readability?

sormuras: Tried it without and the test fails, because we really need the `'\''` defined by `char g() default '\'';`

kenzierocks: But isn't `"'\\\''"` the same as `"'\\''"`?

sormuras: Ah, now I got your point. Will remove one `\` and commit/squash.

swankjesse:
Too simplistic. Won’t handle many of the other interesting cases, such as `\n`. I think you probably want to call into the code that escapes strings and just swap out the open and close quotes.

swankjesse:
I think that’s probably near `CodeWriter.stringLiteral()`

sormuras:
Ja. Gimme a sec' :sweat:

sormuras:
Testing against these values now: `char[] g() default {0, 0xCAFE, 'z', '€', '"', '\'', '\t', '\n'};`

sormuras:
Hm, `Character.MAX_VALUE` is interesting - it's not esacped by `CodeWriter.stringLiteral`.

sormuras:
Uses static version of `CodeWriter.stringLiteral()` now. New cases added to that method should not break the character rendering here.

sormuras:
Is there no JDK utility method that handles the conversion of every single characters to it's source literal? 
* https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html ?
* Apache Commons ?

JakeWharton:
We intend to keep this library dependency free so only the JDK would be able to provide something. But off hand, I can't think of any. Others wanting that functionality should just use this library :grinning:

sormuras:
Don't want to introduce an external lib! Just curious. 

Found a candidate, though: https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html#escapeJava(java.lang.String)

sormuras:
Note: i do not know the meaning of '0xCAFE', which translates to 쫾

JakeWharton:
Probably just an example of a UTF-16 char which was used because Java class
files start with 0xCAFEBABE

sormuras:
Google says: "who" https://translate.google.de/?#ko/en/%EC%AB%BE

swankjesse:
```new char[] { 'A', 'w', 'e', 's', 'o', 'm', 'e', '!' }```

PR #398, opened 2016-01-09
Static import support added.

sormuras:
Solves #73 like @swankjesse wants it. :sweat_smile:

Also includes a fix for a stack overflow that happend inside/between `hashCode()`s of `ParameterizedTypeName` and `TypeVariableName`.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -23,10 +23,12 @@ import java.util.Formatter;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Modifier;
 
 import static com.squareup.javapoet.Util.checkArgument;
@@ -50,6 +52,8 @@ final class CodeWriter {
   private boolean comment = false;
   private String packageName = NO_PACKAGE;
   private final List<TypeSpec> typeSpecStack = new ArrayList<>();
+  private final Set<String> staticImportClassNames;
+  private final Set<String> staticImports;
   private final Map<String, ClassName> importedTypes;
   private final Map<String, ClassName> importableTypes = new LinkedHashMap<>();
   private final Set<String> referencedNames = new LinkedHashSet<>();
@@ -63,17 +67,23 @@ final class CodeWriter {
   int statementLine = -1;
 
   CodeWriter(Appendable out) {
-    this(out, "  ");
+    this(out, "  ", Collections.<String>emptySet());
   }
 
-  CodeWriter(Appendable out, String indent) {
-    this(out, indent, Collections.<String, ClassName>emptyMap());
+  CodeWriter(Appendable out, String indent, Set<String> staticImports) {
+    this(out, indent, Collections.<String, ClassName>emptyMap(), staticImports);
   }
 
-  CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes) {
+  CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes,
+      Set<String> staticImports) {
     this.out = checkNotNull(out, "out == null");
     this.indent = checkNotNull(indent, "indent == null");
     this.importedTypes = checkNotNull(importedTypes, "importedTypes == null");
+    this.staticImports = checkNotNull(staticImports, "staticImports == null");
+    this.staticImportClassNames = new LinkedHashSet<>();
+    for (String signature : staticImports) {
+      staticImportClassNames.add(signature.substring(0, signature.lastIndexOf('.')));
+    }
   }
 
   public Map<String, ClassName> importedTypes() {
@@ -198,7 +208,10 @@ final class CodeWriter {
 
   public CodeWriter emit(CodeBlock codeBlock) throws IOException {
     int a = 0;
-    for (String part : codeBlock.formatParts) {
+    ClassName deferredTypeName = null; // used by "import static" logic
+    ListIterator<String> partIterator = codeBlock.formatParts.listIterator();
+    while (partIterator.hasNext()) {
+      String part = partIterator.next();
       switch (part) {
         case "$L":
           emitLiteral(codeBlock.args.get(a++));
@@ -218,6 +231,17 @@ final class CodeWriter {
 
         case "$T":
           TypeName typeName = (TypeName) codeBlock.args.get(a++);
+          // defer "typeName.emit(this)" if next format part will be handled by the default case
+          if (typeName instanceof ClassName && partIterator.hasNext()) {
+            if (!codeBlock.formatParts.get(partIterator.nextIndex()).startsWith("$")) {
+              ClassName candidate = (ClassName) typeName;
+              if (staticImportClassNames.contains(candidate.canonicalName)) {
+                checkState(deferredTypeName == null, "pending type for static import?!");
+                deferredTypeName = candidate;
+                break;
+              }
+            }
+          }
           typeName.emit(this);
           break;
 
@@ -247,11 +271,46 @@ final class CodeWriter {
           break;
 
         default:
+          // handle deferred type
+          if (deferredTypeName != null) {
+            if (part.startsWith(".")) {
+              if (emitStaticImportMember(deferredTypeName.canonicalName, part)) {
+                // okay, static import hit and all was emitted, so clean-up and jump to next part
+                deferredTypeName = null;
+                break;
+              }
+            }
+            deferredTypeName.emit(this);
+            deferredTypeName = null;
+          }
           emitAndIndent(part);
           break;
       }
     }
     return this;
+  }
+
+  private static String gleanMemberName(String part) {
+    for (int i = 1; i <= part.length(); i++) {
+      if (!SourceVersion.isIdentifier(part.substring(0, i))) {
+        return part.substring(0, i - 1);
+      }
+    }
+    return part;
+  }
+
+  private boolean emitStaticImportMember(String canonical, String part) throws IOException {
+    String partWithoutLeadingDot = part.substring(1);
+    char first = partWithoutLeadingDot.charAt(0);
+    if (!partWithoutLeadingDot.isEmpty() && Character.isJavaIdentifierStart(first)) {
+      String explicit = canonical + "." + gleanMemberName(partWithoutLeadingDot);
+      String wildcard = canonical + ".*";
+      if (staticImports.contains(explicit) || staticImports.contains(wildcard)) {
+        emitAndIndent(partWithoutLeadingDot);
+        return true;
+      }
+    }
+    return false;
   }
 
   private void emitLiteral(Object o) throws IOException {
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -24,9 +24,11 @@ import java.io.Writer;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeSet;
 import javax.annotation.processing.Filer;
 import javax.lang.model.element.Element;
@@ -56,6 +58,7 @@ public final class JavaFile {
   public final String packageName;
   public final TypeSpec typeSpec;
   public final boolean skipJavaLangImports;
+  private final Set<String> staticImports;
   private final String indent;
 
   private JavaFile(Builder builder) {
@@ -63,17 +66,18 @@ public final class JavaFile {
     this.packageName = builder.packageName;
     this.typeSpec = builder.typeSpec;
     this.skipJavaLangImports = builder.skipJavaLangImports;
+    this.staticImports = Util.immutableSet(builder.staticImports);
     this.indent = builder.indent;
   }
 
   public void writeTo(Appendable out) throws IOException {
     // First pass: emit the entire class, just to collect the types we'll need to import.
-    CodeWriter importsCollector = new CodeWriter(NULL_APPENDABLE, indent);
+    CodeWriter importsCollector = new CodeWriter(NULL_APPENDABLE, indent, staticImports);
     emit(importsCollector);
     Map<String, ClassName> suggestedImports = importsCollector.suggestedImports();
 
     // Second pass: write the code, taking advantage of the imports.
-    CodeWriter codeWriter = new CodeWriter(out, indent, suggestedImports);
+    CodeWriter codeWriter = new CodeWriter(out, indent, suggestedImports, staticImports);
     emit(codeWriter);
   }
 
@@ -128,6 +132,13 @@ public final class JavaFile {
 
     if (!packageName.isEmpty()) {
       codeWriter.emit("package $L;\n", packageName);
+      codeWriter.emit("\n");
+    }
+
+    if (!staticImports.isEmpty()) {
+      for (String signature : staticImports) {
+        codeWriter.emit("import static $L;\n", signature);
+      }
       codeWriter.emit("\n");
     }
 
@@ -204,7 +215,8 @@ public final class JavaFile {
   public static final class Builder {
     private final String packageName;
     private final TypeSpec typeSpec;
-    private CodeBlock.Builder fileComment = CodeBlock.builder();
+    private final CodeBlock.Builder fileComment = CodeBlock.builder();
+    private final Set<String> staticImports = new TreeSet<>();
     private boolean skipJavaLangImports;
     private String indent = "  ";
 
@@ -215,6 +227,25 @@ public final class JavaFile {
 
     public Builder addFileComment(String format, Object... args) {
       this.fileComment.add(format, args);
+      return this;
+    }
+
+    public Builder addStaticImport(Enum<?> constant) {
+      return addStaticImport(ClassName.get(constant.getDeclaringClass()), constant.name());
+    }
+
+    public Builder addStaticImport(Class<?> clazz, String... names) {
+      return addStaticImport(ClassName.get(clazz), names);
+    }
+
+    public Builder addStaticImport(ClassName className, String... names) {
+      checkArgument(className != null, "className == null");
+      checkArgument(names != null, "names == null");
+      checkArgument(names.length > 0, "names array is empty");
+      for (String name : names) {
+        checkArgument(name != null, "null entry in names array: %s", Arrays.toString(names));
+        staticImports.add(className.canonicalName + "." + name);
+      }
       return this;
     }
 
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -163,11 +163,12 @@ public class TypeName {
     if (this == o) return true;
     if (o == null) return false;
     if (getClass() != o.getClass()) return false;
-    return toString().equals(o.toString());
+    return keyword.equals(((TypeName) o).keyword);
   }
 
   @Override public int hashCode() {
-    return toString().hashCode();
+    if (keyword == null) throw new AssertionError();
+    return keyword.hashCode();
   }
 
   @Override public final String toString() {
--- a/src/main/java/com/squareup/javapoet/TypeVariableName.java
+++ b/src/main/java/com/squareup/javapoet/TypeVariableName.java
@@ -34,6 +34,7 @@ import static com.squareup.javapoet.Util.checkNotNull;
 public final class TypeVariableName extends TypeName {
   public final String name;
   public final List<TypeName> bounds;
+  private final int hashCode;
 
   private TypeVariableName(String name, List<TypeName> bounds) {
     this(name, bounds, new ArrayList<AnnotationSpec>());
@@ -43,6 +44,7 @@ public final class TypeVariableName extends TypeName {
     super(annotations);
     this.name = checkNotNull(name, "name == null");
     this.bounds = bounds;
+    this.hashCode = name.hashCode() ^ bounds.hashCode();
 
     for (TypeName bound : this.bounds) {
       checkArgument(!bound.isPrimitive() && bound != VOID, "invalid bound: %s", bound);
@@ -67,7 +69,7 @@ public final class TypeVariableName extends TypeName {
   }
 
   @Override public int hashCode() {
-    return name.hashCode() ^ bounds.hashCode();
+    return hashCode;
   }
 
   @Override CodeWriter emit(CodeWriter out) throws IOException {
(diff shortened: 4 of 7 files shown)


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -63,17 +68,25 @@
...
   }
 
-  CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes) {
+  CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes,
+      List<StaticImportItem> staticImportItems) {
     this.out = checkNotNull(out, "out == null");
     this.indent = checkNotNull(indent, "indent == null");
     this.importedTypes = checkNotNull(importedTypes, "importedTypes == null");
+    this.staticImportClassNames = new HashSet<>();
+    this.staticImportSignatures = new HashSet<>();
```
JakeWharton: `LinkedHashSet` for both

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -63,17 +68,25 @@
...
+    checkNotNull(staticImportItems, "staticImportItems == null");
```
JakeWharton: Move up with other argument validation

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -163,11 +163,12 @@ public TypeName unbox() {
     if (this == o) return true;
     if (o == null) return false;
     if (getClass() != o.getClass()) return false;
-    return toString().equals(o.toString());
+    return keyword.equals(((TypeName) o).keyword);
   }
 
   @Override public int hashCode() {
-    return toString().hashCode();
+    if (keyword == null) throw new AssertionError();
```
JakeWharton: This seems really presumptuous. What if consumers are using a map or set for these types?

sormuras: See `emit(CodeWriter)` below, just mimicked that behaviour. Like `emit`, all subclasses of `TypeName` override `hashCode`, should work well with collections.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Sure.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Moved.

src/main/java/com/squareup/javapoet/TypeName.java
JakeWharton: Got it.

src/main/java/com/squareup/javapoet/TypeVariableName.java
```diff
@@ -43,6 +44,7 @@ private TypeVariableName(String name, List<TypeName> bounds, List<AnnotationSpec
     super(annotations);
     this.name = checkNotNull(name, "name == null");
     this.bounds = bounds;
+    this.boundsHashCode = bounds.hashCode();
```
swankjesse: is this problematic for cycle detection?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
           break;
 
         default:
+          if (potentiallyStaticImported == null) {
+            emitAndIndent(part);
+            break;
+          }
+          ClassName pendingClassName = potentiallyStaticImported;
```
swankjesse: I think you want a helper method here. That way you can avoid the duplication between 268 and 288 ?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+          potentiallyStaticImported = null;
+          if (part.startsWith(".")) {
+              int endIndex = part.indexOf('(');
```
swankjesse: or any non-word character, right? like `System.out`

swankjesse: (The part might be `.out.println()`)

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+              if (endIndex < 0) endIndex = part.length();
+              String memberName = part.substring(1, endIndex);
+              checkState(isIdentifier(memberName), "identifier expected, but got: %s", memberName);
```
swankjesse: I’m guessing this explodes on `System.out.println()`

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+              String candidate = pendingClassName.canonicalName + "." + memberName;
```
swankjesse: Neat

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+              String wildcard = pendingClassName.canonicalName + ".*";
```
swankjesse: Neat.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+              if (staticImportSignatures.contains(candidate)
+                  || staticImportSignatures.contains(wildcard)) {
+                String partTail = part.substring(memberName.length() + 1);
```
swankjesse: Instead of this line and 284, isn't it just, `emitAndIndent(part.substring(1));` ?

swankjesse: Ie. chomp the dot ?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +265,27 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+                emitAndIndent(memberName + partTail);
+                break;
+              }
+          }
+          pendingClassName.emit(this);
```
swankjesse: Neat.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -40,6 +42,22 @@
 
 /** A Java file containing a single top level class. */
 public final class JavaFile {
+  static final class StaticImportItem implements Comparable<StaticImportItem> {
+    final ClassName className;
+    final String memberName;
+    final String signature;
```
swankjesse: Do we need anything other than signature?

swankjesse:
I like where this is going.

src/main/java/com/squareup/javapoet/TypeVariableName.java
sormuras: Both, `TypesTest` and `TypesEclipseTest`, run into a cycle, when line 117 in `AbstractTypesTest.getTypeVariableTypeMirrorRecursive()` the statement `typeVariableName.bounds.toString()` is executed. But only when:
 * `ParameterizedTypeName.hashCode()` and `TypeVariableName.hashCode()` calculate the internal collection hash code *on the fly*
 * it only happens if run via Maven/javac - if started inside Eclipse, there's no stack overflow produced

See failure report:
```
	at com.squareup.javapoet.ParameterizedTypeName.hashCode(ParameterizedTypeName.java:64)
	at java.util.AbstractList.hashCode(AbstractList.java:541)
	at java.util.Collections$UnmodifiableList.hashCode(Collections.java:1307)
	at com.squareup.javapoet.TypeVariableName.hashCode(TypeVariableName.java:72)
	at java.util.AbstractList.hashCode(AbstractList.java:541)
	at java.util.Collections$UnmodifiableList.hashCode(Collections.java:1307)
	at com.squareup.javapoet.ParameterizedTypeName.hashCode(ParameterizedTypeName.java:64)
	at java.util.AbstractList.hashCode(AbstractList.java:541)
	at java.util.Collections$UnmodifiableList.hashCode(Collections.java:1307)
	at com.squareup.javapoet.ParameterizedTypeName.hashCode(ParameterizedTypeName.java:64)


Results :

Tests in error: 
  getTypeVariableTypeMirrorRecursive(com.squareup.javapoet.TypesEclipseTest)
  getTypeVariableTypeMirrorRecursive(com.squareup.javapoet.TypesTest)

Tests run: 237, Failures: 0, Errors: 2, Skipped: 0
```
I move the entire hash code calculation into the constructor. Both classes are immutable and marked as final.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: **BäMM!** It exploded and was removed.

src/main/java/com/squareup/javapoet/JavaFile.java
sormuras: Needed the canonicle class name for indentification of potentially imported `$T`s ... will try to boil it down to the signature string, as the `if (staticImportClassNames.contains(typeName))` condition in `case "$T":` could be written as `if (staticImportSignatures.contains(typeName.toString()))`

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: True.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: I remove the first block, as the logic can be moved to the second.

src/main/java/com/squareup/javapoet/JavaFile.java
sormuras: Worked out.

sormuras:
Some local tests with mixed normal/static import fail. Fixing...

sormuras:
Tests and fixes committed. `SourceVersion.isIdentifier()` is back for good.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -218,6 +227,14 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
 
         case "$T":
           TypeName typeName = (TypeName) codeBlock.args.get(a++);
+          if (typeName instanceof ClassName) {
+            ClassName candidate = (ClassName) typeName;
+            if (staticImportClassNames.contains(candidate.canonicalName)) {
+              checkState(potentiallyStaticImported == null, "pending type for static import?!");
```
swankjesse: ooh interesting. Like in say, `$T$T`.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -131,6 +135,13 @@ private void emit(CodeWriter codeWriter) throws IOException {
       codeWriter.emit("\n");
     }
 
+    if (!staticImports.isEmpty()) {
+      for (String signature : staticImports) {
```
swankjesse: sort these?

swankjesse: or maybe just use a `TreeSet` for ’em? Order is never important, and that saves a List to Set conversion later!

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -218,6 +230,25 @@ public Builder addFileComment(String format, Object... args) {
...
 
+    public Builder addStaticImport(Enum<?> constant) {
+      return addStaticImport(ClassName.get(constant.getDeclaringClass()), constant.name());
+    }
+
+    public Builder addStaticImport(Class<?> clazz, String... names) {
+      return addStaticImport(ClassName.get(clazz), names);
+    }
+
+    public Builder addStaticImport(ClassName className, String... names) {
```
swankjesse: nice.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -218,6 +230,25 @@ public Builder addFileComment(String format, Object... args) {
       return this;
     }
 
+    public Builder addStaticImport(Enum<?> constant) {
```
swankjesse: neat.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -237,6 +268,7 @@ public Builder indent(String indent) {
     }
 
     public JavaFile build() {
+      Collections.sort(staticImports);
```
swankjesse: ah, there it is.

swankjesse: I think it’s slightly neater to keep it in a sorted set, but it doesn’t really matter!

swankjesse: (I think my reluctance about this is that `build()` is mutating. A crazy person might call `build()` from two threads in parallel, and crash.)

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -109,4 +110,19 @@ public static void requireExactlyOneOf(Set<Modifier> modifiers, Modifier... mutu
   public static boolean hasDefaultModifier(Collection<Modifier> modifiers) {
     return DEFAULT != null && modifiers.contains(DEFAULT);
   }
+
+  public static String identifier(String part) {
```
swankjesse: Maybe rename to identifierPrefix? Should this Just Work for any string, and return null if it doesn't start with an identifier character? That might make the calling code simpler.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -247,6 +264,19 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
           break;
 
         default:
+          ClassName pendingClassName = potentiallyStaticImported;
+          potentiallyStaticImported = null;
+          if (pendingClassName != null && part.startsWith(".")) {
+              String partWithoutLeadingDot = part.substring(1);
+              String memberName = Util.identifier(partWithoutLeadingDot);
```
swankjesse: I think this might crash on input like `$T...` ?

swankjesse: because `Util.identifier` assumes partWithoutLeadingDot is an identifer.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -109,4 +110,19 @@ public static void requireExactlyOneOf(Set<Modifier> modifiers, Modifier... mutu
...
+    checkArgument(part != null, "part == null");
+    checkArgument(!part.isEmpty(), "part is empty");
+    checkArgument(Character.isJavaIdentifierStart(part.charAt(0)), "part[0] is not valid");
+    StringBuilder builder = new StringBuilder();
+    for (int i = 0; i < part.length(); i++) {
+      builder.append(part.charAt(i));
+      if (!SourceVersion.isIdentifier(builder.toString())) {
```
swankjesse: Probably not necessary to keep a string builder, you can just do successive calls to `part.substring(0, i)`. With the builder you make more work for yourself by having to pop.

src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -25,6 +26,119 @@
...
+        + "\n"
+        + "class Util {\n"
+        + "  public static long minutesToSeconds(long minutes) {\n"
+        + "    System.gc();\n"
+        + "    return SECONDS.convert(minutes, TimeUnit.MINUTES);\n"
+        + "  }\n"
+        + "}\n");
+  }
+
+  @Test public void importStaticTwice() {
```
swankjesse: These tests are perfect.

src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -25,6 +26,119 @@
...
+            + "\n"
+            + "class Util {\n"
+            + "  public static long minutesToSeconds(long minutes) {\n"
+            + "    System.gc();\n"
+            + "    return SECONDS.convert(minutes, MINUTES);\n"
+            + "  }\n"
+            + "}\n");
+  }
+
+  @Test public void importStaticTwiceAndGC() {
```
swankjesse: The GC is a weird way to name this!

swankjesse:
I like where this is going. I think I found one little bug, otherwise it's perfect. YES!

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Ja. Needs a test case to show...

src/main/java/com/squareup/javapoet/JavaFile.java
sormuras: The list is a left-over from the `StaticImportItem` class. Will convert it to a sorted set.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Good catch. My mind is still using the `$R` syntax, in which these edge cases are not present.

src/test/java/com/squareup/javapoet/JavaFileTest.java
sormuras: Rename inc.

src/main/java/com/squareup/javapoet/Util.java
sormuras: The only usage is quite simple. :smile: Only put the method into `Util` because of it's easier to test. Move back to `JavaFile` and make it private? It's not intended to be a general purpose method.

swankjesse: Yeah, I don’t like special-purpose methods in general-purpose locations.

sormuras: Alright. Will add tests that show the intended (and invalid) input strings. Your `$T...` comment does not apply, as `$T` is already resolved and every varargs method emitted didn't throw an exception for the entire test suite. You can't statically import a sole type name, can you?

sormuras: I mean, if s/o produces syntax errors ... the compiler will tell anyway.
* `$T$T` will always crash the compiler
* `$T...` can't be used to declare the last method parameter in JavaPoet-style? And if it's found inside a generic `CodeBlock` ... well.

sormuras: I revoke my last two comments.

                .addParameter(Thread.State[].class, "states")
                .varargs(true)
                .build())

produces your prediction

	java.lang.IllegalArgumentException: part[0] is not valid
		at com.squareup.javapoet.Util.checkArgument(Util.java:63)
		at com.squareup.javapoet.Util.identifier(Util.java:117)
		at com.squareup.javapoet.CodeWriter.emit(CodeWriter.java:271)

src/main/java/com/squareup/javapoet/JavaFile.java
sormuras: I'm getting closer and closer... it's been [$E](https://github.com/sormuras/javapoet/commit/ba36bff2656b7215d440239b0c73065a3a0cfdfd), it's still [$R](https://github.com/square/javapoet/pull/388) ... for now, I can live with **$T**, when it comes to static imports.

src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -25,6 +26,141 @@
 
 @RunWith(JUnit4.class)
 public final class JavaFileTest {
+  @Test(expected = IllegalStateException.class)
+  public void importStaticFailsForChainedTypeName() {
```
sormuras: Keep this state check "as is", or should I work around this strange case of `TypeName` chaining? Maybe like, emit the old pendine one and remember the current for the next round?

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: It actually throws for anything that looks like: `$T*$T` with `*` is any number of non-default parts:

    addStatement("$1T$2L$1T", Runtime.class, "?") // illegal if $T is static imported

sormuras: I'll fix/guard that by peeking into the next part and check whether `!part.startsWith("$")` -- ensuring that only if a `$T` is followed by a default case, the emittion of `$T` is defered.

sormuras:
Feels like the final iteration (including a ListIterator) now.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -254,6 +290,29 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
     return this;
   }
 
+  private static String gleanMemberName(String part) {
```
swankjesse: I might have to rename this later to avoid the word glean.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -254,6 +290,29 @@ public CodeWriter emit(CodeBlock codeBlock) throws IOException {
...
+      if (!SourceVersion.isIdentifier(part.substring(0, i))) {
+        return part.substring(0, i - 1);
+      }
+    }
+    return part;
+  }
+
+  private boolean emitStaticImportMember(String canonical, String part) throws IOException {
+    String partWithoutLeadingDot = part.substring(1);
+    char first = partWithoutLeadingDot.charAt(0);
```
swankjesse: Does this explode if the part is a single character?

src/main/java/com/squareup/javapoet/WildcardTypeName.java
```diff
@@ -42,6 +43,7 @@ private WildcardTypeName(List<TypeName> upperBounds, List<TypeName> lowerBounds,
     super(annotations);
     this.upperBounds = Util.immutableList(upperBounds);
     this.lowerBounds = Util.immutableList(lowerBounds);
+    this.hashCode = upperBounds.hashCode() ^ lowerBounds.hashCode();
```
swankjesse: I’m assuming theses change are unrelated, right?

swankjesse:
I suspect there’s a few unlikely corner cases, such as when the _last_ format part is a `$T`, but this is certainly better than what we have now. I might take a pass at this code to get rid of the nasty word _glean_ and to try to tighten up those exotic & unlikely cases. I’m also tempted to try to make this work with `$L`, so you don’t need to statically know what you’re statically importing!

swankjesse:
And `$N`.

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Preparing a *static_polish* PR, with `extractMemberName`

src/main/java/com/squareup/javapoet/CodeWriter.java
sormuras: Dooh. Late time optimization, the `.isEmpty()` check is too late now. Gonna fix it with a test case attached.

src/main/java/com/squareup/javapoet/WildcardTypeName.java
sormuras: Yes and no. The precalc of hash code in `ParameterizedTypeName` is neccessary - the precalcs in `TypeVariableName` and `WildcardTypeName` are here, to a) make it look and work similar and b) prevent future stack overflows.

PR #399, opened 2016-01-10
Static import corner case tests.

sormuras:
@swankjesse wrote in #398 
> I suspect there’s a few unlikely corner cases, such as when the last format part is a $T, but this is certainly better than what we have now. I might take a pass at this code to get rid of the nasty word glean and to try to tighten up those exotic & unlikely cases.

* Is `extractMemberName` better?
* Fix and test for the "one character member name" like `$T.X` is included.
* More exotic cases are tested. Which are missing?
* Does [wire](https://github.com/square/wire) work with current SNAPSHOT of JavaPoet?

>  I’m also tempted to try to make this work with $L, so you don’t need to statically know what you’re statically importing!

Sounds like a brave plan. You mean like having `$T.$L` resolve to `.addStaticImport($T, $L)` ? Same with `$T.$N`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 Change Log
 ==========
 
+JavaPoet 1.5.0-SNAPSHOT *(not yet released)*
+----------------------------
+
+ * New: `import static`! See `JavaFile.Builder.addStaticImport()` variants.
+ * New: Overload `NameAllocator.newName(String)` for creating a one-off name without a tag.
+ * Fix: AnnotationSpec escapes character literals properly.
+ * Fix: Don't stack overflow when `TypeVariableName` is part of `ParameterizedTypeName`.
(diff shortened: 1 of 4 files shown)


sormuras:
Added an example section to README.md concerning static import.
Prototyped a SNAPSHOT section in CHANGELOG.md as well.

sormuras:
Find the wording and layout of the example here: https://github.com/sormuras/javapoet/tree/static_polish#import-static

swankjesse:
Release time?

sormuras:
$R? ;-p
Am 10.01.2016 15:50 schrieb "Jesse Wilson" <notifications@github.com>:

swankjesse:
That still needs more time for me to digest, sadly!

sormuras:
Null problemo. 1.6.0 will love `$R` :smile: ... and Poetry will offer it right away.

PR #403, opened 2016-01-14
Literal conversion clean up.

sormuras:
`Character` and `String` literal conversion refactored.
Made `AnnotationSpec.Builder.addMemberForValue(String format, Object value)` public.
Removed redundant public modifiers from package private Util methods.
Added `UtilTest` class.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import static com.squareup.javapoet.Util.characterLiteralWithoutSingleQuotes;
 import static com.squareup.javapoet.Util.checkNotNull;
 
 import javax.lang.model.element.AnnotationMirror;
@@ -127,7 +128,7 @@ public final class AnnotationSpec {
         }
         if (value.getClass().isArray()) {
           for (int i = 0; i < Array.getLength(value); i++) {
-            builder.addValue(method.getName(), Array.get(value, i));
+            builder.addMemberForValue(method.getName(), Array.get(value, i));
           }
           continue;
         }
@@ -135,7 +136,7 @@ public final class AnnotationSpec {
           builder.addMember(method.getName(), "$L", get((Annotation) value));
           continue;
         }
-        builder.addValue(method.getName(), value);
+        builder.addMemberForValue(method.getName(), value);
       }
     } catch (Exception e) {
       throw new RuntimeException("Reflecting " + annotation + " failed!", e);
@@ -150,7 +151,7 @@ public final class AnnotationSpec {
     for (ExecutableElement executableElement : annotation.getElementValues().keySet()) {
       String name = executableElement.getSimpleName().toString();
       AnnotationValue value = annotation.getElementValues().get(executableElement);
-      value.accept(visitor, new Entry(name, value));
+      value.accept(visitor, name);
     }
     return builder.build();
   }
@@ -221,7 +222,9 @@ public final class AnnotationSpec {
      * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
      * the class of the given {@code value} object is not supported.
      */
-    private Builder addValue(String memberName, Object value) {
+    Builder addMemberForValue(String memberName, Object value) {
+      checkNotNull(memberName, "memberName == null");
+      checkNotNull(value, "value == null, constant non-null value expected for %s", memberName);
       if (value instanceof Class<?>) {
         return addMember(memberName, "$T.class", value);
       }
@@ -235,11 +238,7 @@ public final class AnnotationSpec {
         return addMember(memberName, "$Lf", value);
       }
       if (value instanceof Character) {
-        String literal = CodeWriter.stringLiteral(value.toString(), "");
-        literal = literal.substring(1, literal.length() - 1);
-        if (literal.equals("\\\"")) literal = "\"";
-        if (literal.equals("'")) literal = "\\'";
-        return addMember(memberName, "'$L'", literal);
+        return addMember(memberName, "'$L'", characterLiteralWithoutSingleQuotes((char) value));
       }
       return addMember(memberName, "$L", value);
     }
@@ -249,20 +248,10 @@ public final class AnnotationSpec {
     }
   }
 
-  private static class Entry {
-    final String name;
-    final AnnotationValue value;
-
-    Entry(String name, AnnotationValue value) {
-      this.name = name;
-      this.value = value;
-    }
-  }
-
   /**
    * Annotation value visitor adding members to the given builder instance.
    */
-  private static class Visitor extends SimpleAnnotationValueVisitor7<Builder, Entry> {
+  private static class Visitor extends SimpleAnnotationValueVisitor7<Builder, String> {
     final Builder builder;
 
     Visitor(Builder builder) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -221,7 +223,9 @@ public Builder addMember(String name, CodeBlock codeBlock) {
      * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
      * the class of the given {@code value} object is not supported.
      */
-    private Builder addValue(String memberName, Object value) {
+    public Builder addMemberForValue(String memberName, Object value) {
+      checkArgument(memberName != null, "memberName == null");
+      checkArgument(value != null, "only constant non-null values are permitted (%s)", memberName);
```
JakeWharton: Both should be `checkNonNull`

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -106,7 +106,45 @@ public static void requireExactlyOneOf(Set<Modifier> modifiers, Modifier... mutu
         modifiers, Arrays.toString(mutuallyExclusive));
   }
 
-  public static boolean hasDefaultModifier(Collection<Modifier> modifiers) {
+  static boolean hasDefaultModifier(Collection<Modifier> modifiers) {
     return DEFAULT != null && modifiers.contains(DEFAULT);
   }
+
+  // {@linkplain https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.6}
```
JakeWharton: This isn't Javadoc so the `{@linkplain` wrapper isn't needed. But more importantly, I would put this comment inside the method so that if it ever gets moved by automated refactoring tools the comment moves with it.

src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -106,7 +106,45 @@ public static void requireExactlyOneOf(Set<Modifier> modifiers, Modifier... mutu
...
+      case '\b': return "\\b"; /* \u0008: backspace (BS) */
+      case '\t': return "\\t"; /* \u0009: horizontal tab (HT) */
+      case '\n': return "\\n"; /* \u000a: linefeed (LF) */
+      case '\f': return "\\f"; /* \u000c: form feed (FF) */
+      case '\r': return "\\r"; /* \u000d: carriage return (CR) */
+      case '\"': return "\"";  /* \u0022: double quote (") */
+      case '\'': return "\\'"; /* \u0027: single quote (') */
+      case '\\': return "\\";  /* \u005c: backslash (\) */
+      default:
+      return Character.isISOControl(c) ? String.format("\\u%04x", (int) c) : Character.toString(c);
```
JakeWharton: nit: needs indent

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: Or introduce a `Util.checkArgumentNotNull(memberName, "memberName")`? That a) checks for `null` and b) prepares a `argument memberName must not be null` and c-1) throws a `IllegalArgumentException` or c-2) a `NullPointerException`.

JakeWharton:
Cool tests

src/main/java/com/squareup/javapoet/AnnotationSpec.java
JakeWharton: This came up very recently. Our rule is: NPE when null is not expected even if it is an argument, and IAE for non-null expectation violations.

src/main/java/com/squareup/javapoet/Util.java
sormuras: I know ... but than it hits the 100 chars per line border. Will `import static Character.isISOControl()`.

sormuras:
I found (and corrected) another bug in `AnnotationSpec.get(AnnotationMirror)` - it didn't use the correct "Object conversion" path. Please have a close look at the changes.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -221,7 +222,9 @@ public Builder addMember(String name, CodeBlock codeBlock) {
      * depending on the given {@code value} object. Falls back to {@code "$L"} literal format if
      * the class of the given {@code value} object is not supported.
      */
-    private Builder addValue(String memberName, Object value) {
+    public Builder addMemberForValue(String memberName, Object value) {
```
swankjesse: Hmmm . . . . is our approach wrong here?

With this API we offer two ways to do everything:
```
  builder.addMember("foo", "$S", "hello");
  builder.addMemberForValue("foo", "hello");
  builder.addMember("bar", "$T.$N", Roshambo.class, Roshambo.ROCK);
  builder.addMemberForValue("bar", Roshambo.ROCK);
```
Is it better to permit a single API?

```
  builder.addMember("foo", "$S", "hello");
  builder.addMember("foo", "$V", "hello");
  builder.addMember("bar", "$T.$N", Roshambo.class, Roshambo.ROCK);
  builder.addMember("bar", "$V", Roshambo.ROCK);
```

src/test/java/com/squareup/javapoet/UtilTest.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
swankjesse: 2016

src/main/java/com/squareup/javapoet/AnnotationSpec.java
sormuras: You do remember the overload clash with the single method name? The extra
name was on purpose.
Or are you suggesting dropping one of the old methods?

src/test/java/com/squareup/javapoet/UtilTest.java
sormuras: Should be a mass update for all files, imho.

JakeWharton: Changing copyright dates can invalidate the copyright.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: I’m not sure. This change is tricky because it brings us to having two different ways to do everything.

sormuras: Pardon me - I totally missed the `$V` marker. Interesting idea...

sormuras: Maybe the problem is rooted [here](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/AnnotationSpec.java#L199) in the declaration of the member map : `Map<String, List<CodeBlock>> members`
If it was a key-pair consisting of the member type and member name (`ParameterSpec`?)... we could also check for valid types and look up the correct format string like `$S`, `$T` and so on.

swankjesse: Regardless, I’m reluctant to expose a new API here. Wanna remove the new API so we can unblock the rest of this PR ?

sormuras: Package private or private private?

src/test/java/com/squareup/javapoet/UtilTest.java
sormuras: 2016 is in.

src/main/java/com/squareup/javapoet/Util.java
sormuras: Moved inside the method w/o the `{@...}` tag.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
swankjesse: No preference from me. Least visibility is a good rule of thumb.

sormuras: Went with package private, for internal use/testing.

swankjesse:
charles-dickens

PR #404, opened 2016-01-17
$X or custom CodeBlock supplier.

sormuras:
Clutching at straws, seeing `$R` vanish... enter `$X`!

With this generic extension point, JavaPoet provides the tool to implement custom acceleration implementations. Entire #388 (with #317) won't be part of JavaPoet, but at users project scope, it is 100% available.

What do you fellow poets think?

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -46,7 +46,9 @@ import static com.squareup.javapoet.Util.checkArgument;
  *       that. For example, {@code 6" sandwich} is emitted {@code "6\" sandwich"}.
  *   <li>{@code $T} emits a <em>type</em> reference. Types will be imported if possible. Arguments
  *       for types may be {@linkplain Class classes}, {@linkplain javax.lang.model.type.TypeMirror
-,*       type mirrors}, and {@linkplain javax.lang.model.element.Element elements}.
+ *       type mirrors}, and {@linkplain javax.lang.model.element.Element elements}.
+ *   <li>{@code $X} emits a custom <em>code block</em> provided by an implementation of
+ *       {@link Supplier}.
  *   <li>{@code $$} emits a dollar sign.
  *   <li>{@code $&gt;} increases the indentation level.
  *   <li>{@code $&lt;} decreases the indentation level.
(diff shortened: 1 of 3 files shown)


swankjesse:
Doesn’t `$L` do this already?

sormuras:
That'd be great! But how? Via `toString()`?

swankjesse:
It’s in CodeWriter.emitLiteral.

sormuras:
But `CodeBlock` is a final class. Can't do `MemberRef extends CodeBlock`. What works is `MemberRef implements CodeBlock.Supplier` with the following:
```java
  public CodeBlock get() {
    CodeBlock.Builder code = CodeBlock.builder();
    if (isStatic) {
      code.add("$T.", type);
    }
    if (kind == Kind.METHOD) {
      emitTypeArguments(code);
    }
    code.add("$L", name);
    return code.build();
  }
```
Usage looks like:
`codeBlockify("$L.$X($L)", "buffer", methodRef, parameter.name)`

If I replaced that framed `X` with an `L`...

swankjesse:
Are these equivalent?

```
builder.addStatement("return $X;", memberRef);
```

```
builder.addStatement("return $L;", memberRef.toCodeBlock());
```

sormuras:
...and above `get().toString()` as implementation of `MemberRef.toString()`: all test lights green - but what happens to the (statically) importable types emitted by the custom code block? I think, they are emitted without the JavaFile context.

swankjesse:
Yeah, I think we should probably do a first pass of flattening code blocks to make tricks like static imports work better. The flattening pass would convert a tree of nested code blocks into a single list.

sormuras:
Ah, now I see.

Forget about `$X`. What about enhancing the already mighty `$L` with an `instanceOf CodeBlock.Supplier` case? Or a platform-provided `instanceOf Callable<CodeBlock>`. In Java 8, of course `instanceof Supplier<CodeBlock>`.

sormuras:
```java
    } else if (o instanceof CodeBlock.Supplier) {
      CodeBlock codeBlock = ((CodeBlock) o).get();
      emit(codeBlock);
```
No need for an extra pass.

swankjesse:
Thanks to erasure we can’t test if an object is an `instanceof Callable<CodeBlock>`. Only `instanceof Callable` which isn't sufficient.

Isn’t it simpler just to call `get()` when specifying the value? Like this:
```
builder.addStatement("return $L;", memberRef.get());
```

sormuras:
A tad more convenient would be the `instanceof CodeBlock.Supplier` version. :smile:

Will rewrite my local code as you suggest, then I'll have a look. Style-wise.

sormuras:
Replaced every `$R` with `$L`, using the `toCodeBlock()` when specifying the value and ... consider me happy!

https://github.com/sormuras/stash/compare/06e4f09e84bd...501a88ec1875

PR #406, opened 2016-01-18
Static imports fix in progress...

sormuras:
Is this fix going into the right direction?

The crazy format tests complains now...

### DIFF
--- a/src/test/java/com/squareup/javapoet/JavaFileImportStaticTest.java
+++ b/src/test/java/com/squareup/javapoet/JavaFileImportStaticTest.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


swankjesse:
I don’t think this is exactly it. Note that `null` isn’t the same as absent. It’s perfectly reasonable to pass `null` as an argument to `$L` etc.

I’m not exactly sure how this should work. Probably what we want is to change say, `JavaFile.emit()` to create a `CodeBlock` rather than immediately writing to a `CodeWriter`. Only once everything is a `CodeBlock` can we flatten correctly.

sormuras:
Ja, makes more sense to me. Will try it that way.

sormuras:
Moved the `importStaticForCrazyFormatsWorks()` test into an own file, utilizing parameters and checking emitted statements. Two cases fail with assumption errors, because `$N` and `$L` does not work with static imports, yet.

sormuras:
Needs more work.

PR #407, opened 2016-01-19
Support for initializer blocks

sockeqwe:
- Static initializers can be added more than one, so I did the same for instance initializers.
- Unfortunately checkstyle is now complaining that the method `TypeSpec.emit()` has more than 150 lines of code. Indeed now (with initilizer block support) it has 153. I have adjusted the number of max lines for methods to 155 in `checkstyle.xml`
- Bugfix: `toBuilder()` didn't respect static block.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -496,7 +496,7 @@ public final class TypeSpec {
     }
 
     public Builder addInitializerBlock(CodeBlock block) {
-      if (!(kind == Kind.CLASS || kind == Kind.ENUM)) {
+      if ((kind != Kind.CLASS && kind != Kind.ENUM)) {
         throw new UnsupportedOperationException(kind + " can't have initializer blocks");
       }
       initializerBlock.add("{\n")
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -52,6 +52,7 @@ public final class TypeSpec {
   public final Map<String, TypeSpec> enumConstants;
   public final List<FieldSpec> fieldSpecs;
   public final CodeBlock staticBlock;
+  public final CodeBlock initializerBlock;
   public final List<MethodSpec> methodSpecs;
   public final List<TypeSpec> typeSpecs;
   public final List<Element> originatingElements;
@@ -69,6 +70,7 @@ public final class TypeSpec {
     this.enumConstants = Util.immutableMap(builder.enumConstants);
     this.fieldSpecs = Util.immutableList(builder.fieldSpecs);
     this.staticBlock = builder.staticBlock.build();
+    this.initializerBlock = builder.initializerBlock.build();
     this.methodSpecs = Util.immutableList(builder.methodSpecs);
     this.typeSpecs = Util.immutableList(builder.typeSpecs);
 
@@ -118,6 +120,8 @@ public final class TypeSpec {
     builder.fieldSpecs.addAll(fieldSpecs);
     builder.methodSpecs.addAll(methodSpecs);
     builder.typeSpecs.addAll(typeSpecs);
+    builder.initializerBlock.add(initializerBlock);
+    builder.staticBlock.add(staticBlock);
     return builder;
   }
 
@@ -234,6 +238,13 @@ public final class TypeSpec {
         firstMember = false;
       }
 
+      // initializer block
+      if (!initializerBlock.isEmpty()) {
+        if (!firstMember) codeWriter.emit("\n");
+        codeWriter.emit(initializerBlock);
+        firstMember = false;
+      }
+
       // Constructors.
       for (MethodSpec methodSpec : methodSpecs) {
         if (!methodSpec.isConstructor()) continue;
@@ -269,18 +280,21 @@ public final class TypeSpec {
     }
   }
 
-  @Override public boolean equals(Object o) {
+  @Override
+  public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null) return false;
     if (getClass() != o.getClass()) return false;
     return toString().equals(o.toString());
   }
 
-  @Override public int hashCode() {
+  @Override
+  public int hashCode() {
     return toString().hashCode();
   }
 
-  @Override public String toString() {
+  @Override
+  public String toString() {
     StringWriter out = new StringWriter();
     try {
       CodeWriter codeWriter = new CodeWriter(out);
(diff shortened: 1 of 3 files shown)


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -269,18 +280,21 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
...
+  @Override
+  public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null) return false;
     if (getClass() != o.getClass()) return false;
     return toString().equals(o.toString());
   }
 
-  @Override public int hashCode() {
+  @Override
```
swankjesse: Not: no  need to reformat these!

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -480,6 +495,18 @@ public Builder addStaticBlock(CodeBlock block) {
       return this;
     }
 
+    public Builder addInitializerBlock(CodeBlock block) {
+      if (!(kind == Kind.CLASS || kind == Kind.ENUM)) {
```
swankjesse: Mind simplifying this expression? 
```
if (kind != Kind.CLASS && kind != Kind.ENUM) {
```

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -53,6 +51,7 @@
   private static final String donutsPackage = "com.squareup.donuts";
 
   @Rule public final CompilationRule compilation = new CompilationRule();
+  @Rule public final ExpectedException expectedException = ExpectedException.none();
```
swankjesse: Please don't use this Rule. It's strictly worse than the try/catch clauses it attempts to improve upon.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1948,35 +1951,180 @@ private String toString(TypeSpec typeSpec) {
...
+            .build())
+        .build();
+    assertThat(toString(taco)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import java.lang.Override;\n"
+        + "import java.lang.String;\n"
+        + "\n"
+        + "class Taco {\n"
+        + "  private static final String FOO;\n\n"
```
swankjesse: Do me a favor? Use a separate line for each `\n` so we don't have two `\n\n`

src/test/java/com/squareup/javapoet/TypeSpecTest.java
vanniktech: @swankjesse Interesting, do you mind elaborating?

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1948,35 +1951,180 @@ private String toString(TypeSpec typeSpec) {
...
+        + "  static {\n"
+        + "    FOO = \"FOO\";\n"
+        + "  }\n\n"
+        + "  private String mFoo;\n\n"
```
swankjesse: Never use the m prefix, its disgusting.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -1948,35 +1951,180 @@ private String toString(TypeSpec typeSpec) {
...
+            .addAnnotation(Override.class)
+            .addModifiers(Modifier.PUBLIC)
+            .returns(String.class)
+            .addCode("return FOO;\n")
+            .build())
+        .addInitializerBlock(CodeBlock.builder()
+            .addStatement("mFoo = $S", "FOO")
+            .build())
+        .build();
+    assertThat(toString(taco)).isEqualTo(""
```
swankjesse: This assert isn't necessary.

swankjesse:
Excellent. I've got some style nitpicks but overall nice change. Fix em and squash or I'll take care of that later this week.

src/main/java/com/squareup/javapoet/TypeSpec.java
sockeqwe: hm, interesting done automatically by intellij formater with squares java code style

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sockeqwe: Alright, replaced that with a try/catch clause. Would be interesting to hear your opinion on that rule.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sockeqwe: I just have copy & pasted it from static block test. Have moved second `\n` in new line (also in static block test)

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sockeqwe: Agree! I have copy & pasted it from static block test. Thought that was desired by you guys ...

sockeqwe:
Thanks for your feedback! Have changed that

src/main/java/com/squareup/javapoet/TypeSpec.java
JakeWharton: Are you sure? Both the 'Square' and 'SquareAndroid' should not be doing this.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
JakeWharton: * Can't have multiple methods that fail in a single test, can have multiple try/catch blocks.
* Can't assert things like the `getCause()`, regex on the message, elements in the stack trace.
* Can't assert the exact statement that should throw, any matching exception passes test vs. only a single expression with try/catch

src/main/java/com/squareup/javapoet/TypeSpec.java
sockeqwe: You are right, I can't reproduce it.
I guess it was my fault. When I have started working on this, I have used intellij's default formatter and I guess I have pressed the shortcuts for formatting code somehow (well I press this shortcut all the time automatically without noticing it).

Before committing and pushing I have applied 'Square' formatter and have assumed that it will reformat previous changes as well ...

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -234,6 +238,13 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         firstMember = false;
       }
 
+      // initializer block
```
swankjesse: Drop this comment? It doesn’t add much value!

swankjesse:
Merged & squashed manually!

swankjesse:
Thanks!

sockeqwe:
Thank you!

Issue #409, opened 2016-01-20
CodeWriter internals


sormuras:
Is there a reason, why `CodeWriter` uses an [`Appendable`](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/CodeWriter.java#L48) and not a `StringBuilder` for collecting emitted content?

* For an import refactoring i need the current position (which is equals to `StringBuilder.length()`) of the content already emitted. My current solution reads `out.toString().length()` ... which is really awkward.
* Could get rid of all `IOException` declaration and handling around `emit()` methods.

swankjesse:
Unfortunately there's at least one project that's generating gigabyte-sized .java files, and this will break them.

sormuras:
:fearful: Yikes! Mh, then an `Appendable` decorator that count the chars before passing it down to the component will solve my issue here. Because, having an `out.toString()` call will certainly produce an OutOfMemoryException as well...

sormuras:
Is "type import" and "static import" a feature needed if you generate such large files?!

What trying is emitting an entire `JavaFile` with all `$T` as canonical names with collection meta-information of each `$T` rendering in to a StringBuilder instance. After processing the meta-information according to JavaPoet constraints, the "static import" and "import" sections are injected to builder and the imported package (and package+member) string are deleted.

sormuras:
After typing the former comment, I'm realizing how to achieve that w/o the in-memory `StringBuilder`. Will need a `CountingAppendableDecorator` first to produce the meta-information, process'em like now and feed the result into a `ManglingAppenableDecorator` that filters the `$T`s occurences respecting the import statementes...

sormuras:
Updated the PR to only contain the exception handling clean up.

sormuras:
Revoked PR - @tbroyer was convincing.

PR #410, opened 2016-01-20
Centralized IOException handling in CodeWriter

sormuras:
Removing all `throws IOException` from `emit()`-related methods and clear the exception handling from many `toString()` implementations.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -41,6 +41,13 @@ import static com.squareup.javapoet.Util.stringLiteralWithDoubleQuotes;
  * honors imports, indentation, and deferred variable names.
  */
 final class CodeWriter {
+  @SuppressWarnings("serial")
+  static final class EmitRuntimeException extends RuntimeException {
+    EmitRuntimeException(IOException cause) {
+      super(cause);
+    }
+  }
+
   /** Sentinel value that indicates that no user-provided package has been set. */
   private static final String NO_PACKAGE = new String();
 
@@ -131,7 +138,7 @@ final class CodeWriter {
     return this;
   }
 
-  public void emitComment(CodeBlock codeBlock) throws IOException {
+  public void emitComment(CodeBlock codeBlock) {
     trailingNewline = true; // Force the '//' prefix for the comment.
     comment = true;
     try {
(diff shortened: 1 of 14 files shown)


src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -71,14 +59,17 @@ private JavaFile(Builder builder) {
   }
 
   public void writeTo(Appendable out) throws IOException {
+    StringBuilder temp = new StringBuilder();
     // First pass: emit the entire class, just to collect the types we'll need to import.
-    CodeWriter importsCollector = new CodeWriter(NULL_APPENDABLE, indent, staticImports);
+    CodeWriter importsCollector = new CodeWriter(temp, indent, staticImports);
```
sormuras: The content of `temp` is not used - but the allocated memory will be re-used be the second pass.

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -449,21 +448,28 @@ CodeWriter emitAndIndent(String s) throws IOException {
...
     for (int j = 0; j < indentLevel; j++) {
-      out.append(indent);
+      append(indent);
+    }
+  }
+
+  private void append(CharSequence csq) {
+    try {
+      out.append(csq);
+    } catch (IOException e) {
```
sormuras: One-and-only place of exception handling.

tbroyer: Does that mean all `JavaFile.writeTo` methods would now throw `AssertionError`s instead of `IOException`s? Sounds bad.

sormuras: No, the signatures of `JavaFile.writeTo` methods are not affected. All four still declare and throw `IOException`s if there are, for example file system related, i/o problems.

For those exceptions caught here, I could wrap them in an unchecked exception, that can be analyzed by `JavaFile.writeTo(Appendable)` for re-throwing the cause. That'd preserve the original behaviour in all cases - and leaves the `toString()` implementation `IOException`-free.

tbroyer: That sounds like complexity for the sake of complexity.

sormuras: True. But it is small and compact one against a scattered solution effecting all `*Name` and `*Spec` classes. And, imho, the first proposal just transformed that "small complexity" into an unchecked exception - it wasn't lost, as the `JavaFile.writeTo()` still reported `IOException`s.

As an internal in-memory `StringBuilder out` is a no-go anyway, this PR is more like a nice-to-have for the import refactoring mentioned in #409.

PR #412, opened 2016-01-21
Static import refactored.

sormuras:
# Motivation
Current `JavaFile//addStaticImport` implementation "doesn't support members with $L". See https://github.com/square/javapoet/commit/4dd534906a1e8aa8b15883706878a45fdb1437da

## Features
* `$T.$L` and many (all ?) other cases are now supported
* It does not fail on intentional white spaces either, like `"$T . $N = PI"`
* Auto-skips unused `import static` requests

##  Implementation outline
* Hooked into `lookup(ClassName)` and `emitAndIndent(String)` now.
* When a potential hit for a static import is detected, `lookup` returns an empty string and the `ClassName` instance is remembered as deferred.
* While there's a deferred `ClassName`, all strings (coming from any accelerator like `$L`, `$N`, ...) emitted via `emitAndIndent(String)` are gathered and analyzed for a wildcard or perfect match.
* Depending on the match result: if hit, emit  only the member part or emit the deferred class name.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TreeSet;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Modifier;
 
@@ -57,7 +58,10 @@ final class CodeWriter {
   private final Map<String, ClassName> importedTypes;
   private final Map<String, ClassName> importableTypes = new LinkedHashMap<>();
   private final Set<String> referencedNames = new LinkedHashSet<>();
+  private final Set<String> referencedStaticImports = new TreeSet<>();
   private boolean trailingNewline;
+  private ClassName deferredClassName = null;
+  private final StringBuilder deferredLine = new StringBuilder();
 
   /**
    * When emitting a statement, this is the line of the statement currently being written. The first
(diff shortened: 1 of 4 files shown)


swankjesse:
It’ll take me a while to get up to speed on this PR.

sormuras:
Take your time and test-drive other formats in `JavaFileImportStaticTest`. I'm sure, when it comes to line breaks and other whitespaces, there'll be some more work to do. But it's better then before. :8ball: 

This is my favorite, so far: `"$1T$2L$3L($1T.E+$4L$2L$4L)"` with ` Math.class, ".", "cos", 011` produces: `"cos(E+9.9)"`

sormuras:
More crazy formats ticked off.

sormuras:
Better local variable names and inline comments added.

sormuras:
Rebased.

sormuras:
`CodeBlock.of()` used.

sormuras:
Now skips unused `import static` requests.

swankjesse:
I don’t really understand the motivation behind this change. Is the goal to avoid emitting unused static imports? Might help if instead of calling this “refactor ...” you say “improve ...”. Ie. I want to know why, and right now staring at this code it just seems like we’re making CodeWriter more stateful, and putting too much responsibility into `emitAndIndent()`.

sormuras:
Main point was the missing/failing support for `$L`
> addStaticImport doesn't support members with `$L`
See https://github.com/square/javapoet/pull/412/files#diff-0959801b341a3e8d5e98ce499cf56cf3L138

sormuras:
Rebased.

sormuras:
Updated initial PR description.

sormuras:
Moved PR to new branch, created #443

PR #414, opened 2016-01-22
make TypeSpec.Kind enum public

### LINKED ISSUE #413: make TypeSpec.Kind public
the `TypeSpec.kind` field is already public, but the enum is private. Any particular reason why? It'd be nice to be able to know whether a `TypeSpec` is an interface/class/etc.

caplan:
Closes #413


swankjesse:
Well that was easy!

PR #417, opened 2016-02-07
Allow TypeNames to be annotated twice

ronshapiro:
I don't love that this requires each `TypeName` subclass needs to remember to combine the old and new annotations, but I don't see a good way to enforce that since `TypeName#annotated(List<AnnotationSpec>)` needs to be overridable to be able to return the right type. Open to suggestions.

One of the tests assumed that calling `annotatedTypeName.annotated()` would clear this test - I'm not sure if anyone relies on that functionality, but that would be a breaking change. I could add a `clearAnnotations()` method if you'd like.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ArrayTypeName.java
+++ b/src/main/java/com/squareup/javapoet/ArrayTypeName.java
@@ -41,7 +41,7 @@ public final class ArrayTypeName extends TypeName {
   }
 
   @Override public ArrayTypeName annotated(List<AnnotationSpec> annotations) {
-    return new ArrayTypeName(componentType, annotations);
+    return new ArrayTypeName(componentType, prependAnnotations(annotations));
   }
 
   @Override CodeWriter emit(CodeWriter out) throws IOException {
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -55,7 +55,7 @@ public final class ClassName extends TypeName implements Comparable<ClassName> {
   }
 
   @Override public ClassName annotated(List<AnnotationSpec> annotations) {
-    return new ClassName(names, annotations);
+    return new ClassName(names, prependAnnotations(annotations));
   }
 
   /** Returns the package name, like {@code "java.util"} for {@code Map.Entry}. */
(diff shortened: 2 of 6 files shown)


sormuras:
Hi @ronshapiro 

I can think of an enum `Mode {CLEAR, APPEND, PREPEND}` and overload the `annotated(List)` method with a `annotated(Mode, List)`, using `Mode.CLEAR` as default. 

But, i think, having this *how to treat already attached annotation specs* feature implemented externally offers more flexibility:
```java
static List<AnnotationSpec> prepend(TypeName type, AnnotationSpec... annotations) {
  List<AnnotationSpec> annos = new ArrayList<>(type.annotations);
  annos.addAll(Arrays.asList(annotations));
  return annos;
}

@Test public void annotatedTwice() {
  String expected = "@java.lang.Override @" + NN + " java.lang.String";
  TypeName type = TypeName.get(String.class).annotated(NEVER_NULL);
  String actual = type.annotated(prepend(type, AnnotationSpec.builder(Override.class).build()))
          .toString();
  assertEquals(expected, actual);
}
```

JakeWharton:
I can't think of a reason why anyone would want the "clear" behavior. And
as to append or prepend, any code relying on such ordering is broken.

The current (on master) behavior severely violates the principle of least
surprise and will cause subtle bugs. One would expect an already-annotated
type that is instructed to be annotated again to combine those annotations.

sormuras:
@JakeWharton I thought clearing was required by the initial request #136 or by further discussion made in #354 ff. Can't find it now, though.

Point is, current implementation of `annotated()` behaves like a factory method creating a new type name instance, always based on the not annotated base type name. If it were a builder, i'd expect it to collect the annotations from differents `addAnnotation()` calls. So the violation is the *factory-like behaviour*, imho.

Perhaps, declaring the current implementation as package private and expose the factory statically?
```java
public static TypeName annotated(TypeName source, AnnotationSpec... annotations) {
  List<AnnotationSpec> annos = new ArrayList<>(source.annotations);
  annos.addAll(Arrays.asList(annotations));
  return source.annotated(annos);
}
```

sormuras:
**clear** was mentioned by @swankjesse here: #358 ... *and whether we have methods like withoutAnnotations() to strip ’em.*

swankjesse:
Ugh, not sure how I accidentally merged this. I was trying to comment? Mobile web?

Anyway, I think this PR is almost exactly the behavior we want. The one thing I'd like to change is to make the new annotations show up at the end of the list rather than the beginning.

@ronshapiro do you mind sending a follow up PR?

ronshapiro:
Sure, I'll send that soon.

PR #418, opened 2016-02-09
Append annotations from multiple calls to TypeName#annotated

ronshapiro:
Follow up to @swankjesse's last comment on #417.


swankjesse:
Thankyou!

ronshapiro:
Thanks for the fast turnaround!

Issue #422, opened 2016-02-09
java.lang classes fully qualified when annotated


ronshapiro:
I'm seeing some weirdness which I believe happens when JavaFile.Builder.skipJavaLangImports(true) is specified and java.lang classes are being annotated.

This code:

```java
JavaFile.builder(
        "tacos",
        TypeSpec.classBuilder("Tacos")
            .addField(
                ClassName.get(String.class)
                    .annotated(AnnotationSpec.builder(Override.class).build()),
                "string")
           .build())
    .skipJavaLangImports(true)
    .build()
    .writeTo(System.err);
```

prints:
```java
package tacos;

class Tacos {
  @Override java.lang.String string;
}
```

I would expect `java.lang.String` to just be `String`, but it seems that the `skipJavaLangImports` reverses some functionality?

swankjesse:
Yikes, that's terrible. Will fix.

PR #423, opened 2016-02-10
Fix a bug where annotations broke import resolution.

swankjesse:
Closes: https://github.com/square/javapoet/issues/422


ronshapiro:
:+1: thanks for this quick fix!

Issue #420, opened 2016-02-09
CodeBlock convenience methods


ronshapiro:
In migrating Dagger 2 to JavaPoet, I created a `CodeBlocks.format()` which mimics `CodeBlock.of()` (which was deleted in 0f0f80c0eb67ec2d71e6b01d9801480298eaed57). It's already got 89 usages, and I just came across a need for it in migrating AutoFactory as well. Instead of continually redefining this, would you be in favor of bringing this method back?

Less used, but still pretty useful is `CodeBlocks.join(codeBlocks, delimitter)` which has 34 direct+indirect usages. Common uses of this are to join arguments for a method call, type parameters, etc. This would also be useful in the AutoFactory migration.

Here's the code for reference: [CodeBlocks.java](https://github.com/google/dagger/blob/6475f47799f55eb2d22af72b7c8bddf3f7b1d01c/compiler/src/main/java/dagger/internal/codegen/CodeBlocks.java)

swankjesse:
Sure. Send a pull request. I prefer `of()` over `format()`.

JakeWharton:
We have lots of builder().oneThing().build() calls in some to-be-released
things as well. :+1:

JakeWharton:
More evidence: https://github.com/square/javapoet/blob/19fe0a411228683d54e84e9e16b755bcc1a67058/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L2164-L2168

kenzierocks:
I'd like to suggest adding a `addStatement` or similar to match `MethodSpec.Builder.addCode`. I find myself using `addStatement` way more than `addCode`, and I feel that much of my code could benefit from adding a direct version of it to `MethodSpec.Builder`.

sormuras:
@kenzierocks Same for me, see https://github.com/square/javapoet/pull/412/files#diff-f126073fdb367a6757ccfc47a6547ba5R60 - having a `CodeBlock.statementOf(format, args)` would be cool.

swankjesse:
Yep, I think we could definitely use `addStatement()`. We’ve already got most of the other related methods. @ronshapiro would you like the honors?

ronshapiro:
Sure. Could we also discuss the `addLine()` idea as well? 

And would we want parallel static factories: `CodeBlock.statement()` and `CodeBlock.line()`?

swankjesse:
[...]

PR #427, opened 2016-02-10
Revive CodeBlock.of()

ronshapiro:
Requested in #420


src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -32,6 +32,11 @@
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
   }
 
+  @Test public void of() {
+    CodeBlock a = CodeBlock.of("$L taco", "delicious");
```
swankjesse: yes they are

PR #428, opened 2016-02-12
ClassName.canonical and TypeName.binary()

sormuras:
* Published immutable `ClassName.canonicalName` and `ClassName.binaryName` fields.
 * Made `TypeName` report its binary name, which can be used in `Class.forName(className)` calls.

The test looks like:
```java
assertEquals("void", TypeName.VOID.binary());
assertEquals("int", TypeName.INT.binary());
assertEquals("java.lang.Object", TypeName.OBJECT.binary());
assertEquals("java.lang.Thread$State", TypeName.get(Thread.State.class).binary());
assertEquals("[java.util.Map$Entry;", TypeName.get(Map.Entry[].class).binary());
assertEquals("[[[Z", TypeName.get(boolean[][][].class).binary());
assertEquals("java.util.List", ParameterizedTypeName.get(List.class, String.class).binary());
assertEquals("java.util.Set", ParameterizedTypeName.get(Set.class, UUID.class).binary());
```

* `TypeVariableName.binary()` and `WildcardTypeName.binary()` ~~should~~ throw `UnsupportedOperationException`!

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -37,7 +37,8 @@ public final class ClassName extends TypeName implements Comparable<ClassName> {
 
   /** From top to bottom. This will be ["java.util", "Map", "Entry"] for {@link Map.Entry}. */
   final List<String> names;
-  final String canonicalName;
+  /** @see Class#getCanonicalName() */
+  public final String canonicalName;
 
   private ClassName(List<String> names) {
     this(names, new ArrayList<AnnotationSpec>());
@@ -49,7 +50,7 @@ public final class ClassName extends TypeName implements Comparable<ClassName> {
       checkArgument(SourceVersion.isName(names.get(i)), "part '%s' is keyword", names.get(i));
     }
     this.names = Util.immutableList(names);
-    this.canonicalName = names.get(0).isEmpty()
+    this.canonicalName = packageName().isEmpty()
         ? Util.join(".", names.subList(1, names.size()))
         : Util.join(".", names);
   }
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -34,6 +34,7 @@ import javax.lang.model.type.NoType;
 import javax.lang.model.type.PrimitiveType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
 import javax.lang.model.util.SimpleTypeVisitor7;
 
 /**
@@ -166,6 +167,10 @@ public class TypeName {
     if (this.equals(BOXED_FLOAT)) return FLOAT;
     if (this.equals(BOXED_DOUBLE)) return DOUBLE;
     throw new UnsupportedOperationException("cannot unbox " + this);
+  }
+
+  public final TypeMirror toTypeMirror(Elements elements) {
+    return elements.getTypeElement(toString()).asType();
   }
 
   @Override public final boolean equals(Object o) {
--- a/src/test/java/com/squareup/javapoet/TypeNameTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeNameTest.java
@@ -21,10 +21,10 @@ import static org.junit.Assert.assertNotEquals;
 
 import java.io.Serializable;
 import java.lang.reflect.Method;
-import java.rmi.server.UID;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
+import java.util.UUID;
 import org.junit.Test;
 
 public class TypeNameTest {
@@ -70,8 +70,8 @@ public class TypeNameTest {
   @Test public void equalsAndHashCodeParameterizedTypeName() {
     assertEqualsHashCodeAndToString(ParameterizedTypeName.get(Object.class),
         ParameterizedTypeName.get(Object.class));
-    assertEqualsHashCodeAndToString(ParameterizedTypeName.get(Set.class, UID.class),
-        ParameterizedTypeName.get(Set.class, UID.class));
+    assertEqualsHashCodeAndToString(ParameterizedTypeName.get(Set.class, UUID.class),
+        ParameterizedTypeName.get(Set.class, UUID.class));
     assertNotEquals(ClassName.get(List.class), ParameterizedTypeName.get(List.class,
         String.class));
   }
@@ -98,5 +98,4 @@ public class TypeNameTest {
     assertThat(a.equals(b)).isTrue();
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
   }
-
 }
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -2168,4 +2168,14 @@ public final class TypeSpecTest {
     assertThat(TypeSpec.enumBuilder(className).addEnumConstant("A").build().name).isEqualTo("Example");
     assertThat(TypeSpec.annotationBuilder(className).build().name).isEqualTo("Example");
   }
+
+  @Test public void toTypeMirror() {
+    TypeName typeName = ClassName.get("java.lang", "String");
+    String expected = "java.lang.String";
+    assertThat(typeName.toTypeMirror(compilation.getElements()).toString()).isEqualTo(expected);
+    // typeName = ParameterizedTypeName.get(Set.class, String.class);
+    // expected = "java.util.Set<java.lang.String>";
+    // assertThat(typeName.toTypeMirror(compilation.getElements()).toString()).isEqualTo(expected);
+    // NPE in typeName.toTypeMirror() ...
+  }
 }


src/main/java/com/squareup/javapoet/ArrayTypeName.java
```diff
@@ -43,6 +43,26 @@ private ArrayTypeName(TypeName componentType, List<AnnotationSpec> annotations)
...
+    if (componentType.isPrimitive()) {
+      if (componentType.equals(TypeName.BOOLEAN)) return stringBuilder.append('Z');
+      if (componentType.equals(TypeName.BYTE)) return stringBuilder.append('B');
+      if (componentType.equals(TypeName.CHAR)) return stringBuilder.append('C');
+      if (componentType.equals(TypeName.DOUBLE)) return stringBuilder.append('D');
+      if (componentType.equals(TypeName.FLOAT)) return stringBuilder.append('F');
+      if (componentType.equals(TypeName.INT)) return stringBuilder.append('I');
+      if (componentType.equals(TypeName.LONG)) return stringBuilder.append('J');
+      if (componentType.equals(TypeName.SHORT)) return stringBuilder.append('S');
+      Util.checkState(false, "primitive array component type %s not handled?", componentType);
```
JakeWharton: I would prefer if this was just a `throw new...` instead of relying on a convenience method for a constant condition.

sormuras: Fixed.

sormuras:
Related to TypeVariable and Wildcard names. Have it return their string representation or throw an unchecked exception? The following reads ... strange.
```java
    assertEquals("P", TypeVariableName.get("P", Number.class).binary());
    assertEquals("java.lang.Number", WildcardTypeName.get(Number.class).binary());
```

src/main/java/com/squareup/javapoet/ArrayTypeName.java
```diff
@@ -43,6 +43,26 @@ private ArrayTypeName(TypeName componentType, List<AnnotationSpec> annotations)
...
+    if (componentType.isPrimitive()) {
+      if (componentType.equals(TypeName.BOOLEAN)) return stringBuilder.append('Z');
+      if (componentType.equals(TypeName.BYTE)) return stringBuilder.append('B');
+      if (componentType.equals(TypeName.CHAR)) return stringBuilder.append('C');
+      if (componentType.equals(TypeName.DOUBLE)) return stringBuilder.append('D');
+      if (componentType.equals(TypeName.FLOAT)) return stringBuilder.append('F');
+      if (componentType.equals(TypeName.INT)) return stringBuilder.append('I');
+      if (componentType.equals(TypeName.LONG)) return stringBuilder.append('J');
+      if (componentType.equals(TypeName.SHORT)) return stringBuilder.append('S');
+      throw new IllegalStateException(componentType + " is primitive and not handled?!");
```
swankjesse: Typically this is better as a naked `AssertionError`. We don’t want to waste bytecodes or sourcecode building strings that’ll never get built.

swankjesse: (ie. just `throw new AssertionError()`)

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -56,6 +56,10 @@
     return new ParameterizedTypeName(rawType, typeArguments);
   }
 
+  @Override StringBuilder binary(StringBuilder stringBuilder) {
+    return rawType.binary(stringBuilder);
```
swankjesse: Does this case make sense?

swankjesse:
What’s the use case here?

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -99,4 +100,14 @@ private void assertEqualsHashCodeAndToString(TypeName a, TypeName b) {
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
   }
 
+  @Test public void binary() {
+    assertEquals("void", TypeName.VOID.binary());
+    assertEquals("int", TypeName.INT.binary());
+    assertEquals("java.lang.Object", TypeName.OBJECT.binary());
```
swankjesse: What specification are you trying to implement here? This looks a little like the JVM’s field descriptor, but that spec uses a `/` instead of a `.`.

http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -99,4 +100,14 @@ private void assertEqualsHashCodeAndToString(TypeName a, TypeName b) {
...
+    assertEquals("java.lang.Thread$State", TypeName.get(Thread.State.class).binary());
+    assertEquals("[java.util.Map$Entry;", TypeName.get(Map.Entry[].class).binary());
```
swankjesse: Is this format specified somewhere?

sormuras: The Javadoc is over here: https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getName()

sormuras:
The use-case is the reversal of `TypeName.get(...)`. Sometimes, I need to check whether the type specified by a `TypeName` is a) findable, loadable and b) inspect some properties of the loaded class. Like `isEnum()` et al.

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
sormuras: May be not, but it doesn't hurt. It mimics this assertion:

`assertEquals("java.util.HashSet", new HashSet<Byte>().getClass().getName());`

Yes, `getClass()` discards the type variable here. On the JavaPoet side, you'd need to check for "t instanceof ParameterizedTypeName" and call `t.rawType.binary()`. This is just a shortcut for it.

swankjesse:
(And/or, return a a TypeMirror?)

sormuras:
Interesting idea. Having just the type name seems like a half-step... like:
`Type TypeName.getType() throws Exception { return Class.forName(binary()); }`
?
Hm, for me, the name is sufficent. Regarding `TypeMirror`, does the `TypeTool` provide a name-2-mirror method?

swankjesse:
Specifically I don’t think that `Class.getName()` is a good API and I don’t want us to endorse it by treating it like a specification. That method is awkward and inconsistent, particularly around how `int` is `int` but `int[]` is `[I`. The reference arrays are similarly nasty.

I also think it’s weird and bad to try & load a class in the host VM based on a JavaPoet name. That won’t end well in general.

swankjesse:
This gets close:
http://docs.oracle.com/javase/7/docs/api/javax/lang/model/util/Elements.html#getTypeElement%28java.lang.CharSequence%29

(And we _could_ accept an `Elements` instance to the method that goes from TypeName to Element)

swankjesse:
. . . . the other upside of using `Elements` is that the API wouldn’t be braindead for parameterized types, wildcard types, etc. The utility could actually give you the full generic type back.

sormuras:
`Elements#getBinaryName(CharSequence)` ... found it, too.

Alright. The simple *toClass()*-conversion moves to [Poetry](https://github.com/sormuras/poetry) - but keep
 * String ClassName.canonical public and
 * introduce a `TypeMirror toTypeMirror/Element` here?

sormuras:
The Javadoc of `getTypeElement` needs the canonical form as an argument. So, generic information are dead here, too. See the commited test in two secs...

sormuras:
Hm, I'll only leave the `public` (in `ClassName#canonical`) in this PR.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -37,7 +37,8 @@
 
   /** From top to bottom. This will be ["java.util", "Map", "Entry"] for {@link Map.Entry}. */
   final List<String> names;
-  final String canonicalName;
+  /** @see Class#getCanonicalName() */
+  public final String canonicalName;
```
swankjesse: is this ever different than `toString()` ?

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -168,6 +169,10 @@ public TypeName unbox() {
     throw new UnsupportedOperationException("cannot unbox " + this);
   }
 
+  public final TypeMirror toTypeMirror(Elements elements) {
```
swankjesse: I think this goes in Poetry too . . . mostly because the documentation necessary to explain all of the failure cases is way larger than the method body itself.

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -21,10 +21,10 @@
 
 import java.io.Serializable;
 import java.lang.reflect.Method;
-import java.rmi.server.UID;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
+import java.util.UUID;
```
swankjesse: Nice.

sormuras:
Will file the tiny `TypeNameTest` import change in a new PR.

PR #430, opened 2016-02-23
Treat ErrorType like a regular DeclaredType in TypeName.get()

ronshapiro:


ronshapiro:
I'm not sure a great way to get an `ErrorType` in tests without trying to use JavaPoet within a processing round, but that doesn't seem to fit well within a JUnit test. Any suggestions?

JakeWharton:
Maybe just create an implementation yourself for the tests?

ronshapiro:
How's this?

swankjesse:
Won’t this just crash when the type turns out to not be what ClassName expects?

ronshapiro:
I'm not fully sure all the possible cases of ErrorType, but I think they
should all have a TypeElement associated with it, which is all ClassName
needs.

swankjesse:
I don’t really understand the consequences of this. ErrorType itself is strange, since it suggests the generated code is going to be garbage anyway. I’m closing this PR. If you can point me at a sample where this approach is useful, that’d at least help me to understand.

tbroyer:
Won't there be cases where you see an `ErrorType` in a round because the type will be generated later? If so, then the generated code isn't necessarily going to be garbage.

swankjesse:
@tbroyer good point! That’s the part I was missing.

swankjesse:
Thanks @ronshapiro & @tbroyer !

ronshapiro:
I was having a tough time wrapping my head around `ErrorType` and the erroneous source code option. I'm not sure what would be best to do in that case, or how to even get in that state. But the not-yet-generated example was what I had initially thought about this for AutoFactory. Thanks @tbroyer for following up!

Issue #431, opened 2016-03-01
Type annotations on qualified names


cushon:
Type annotations on qualified names have to be applied to the simple name of the type. JLS [§9.7.4](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.4) says:

> It is a compile-time error if an annotation of type T applies to a type (or any part of a type) in a type context, and T is applicable in type contexts, and the annotation is not admissible.
> 
> For example, assume an annotation type TA which is meta-annotated with just `@Target(ElementType.TYPE_USE)`. The terms `@TA java.lang.Object` and `java.@TA lang.Object` are illegal because the simple name to which `@TA` is closest is classified as a package name. On the other hand, `java.lang.@TA Object` is legal.

Incorrect:

``` java
@Target(ElementType.TYPE_USE)
@interface Anno {}

class Test {
  @Anno java.util.Map.Entry e;
}
```

```
error: scoping construct cannot be annotated with type-use annotation: @Anno
```

Correct:

``` java
java.util.Map.@Anno Entry e;
```

JavaPoet doesn't get this right.

Failing test case:

```diff
diff --git a/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java b/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
index 35b2b1f..e420d85 100644
--- a/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
@@ -20,16 +20,24 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;

+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
 import java.util.List;
+import java.util.Map;
 import org.junit.Test;

 public class AnnotatedTypeNameTest {

   private final static String NN = NeverNull.class.getCanonicalName();
   private final AnnotationSpec NEVER_NULL = AnnotationSpec.builder(NeverNull.class).build();
+  private final AnnotationSpec TYPE_USE_ANNOTATION =
+      AnnotationSpec.builder(TypeUseAnnotation.class).build();

   public @interface NeverNull {}

+  @Target(ElementType.TYPE_USE)
+  public @interface TypeUseAnnotation {}
+
   @Test(expected=NullPointerException.class) public void nullAnnotationArray() {
[...]

PR #432, opened 2016-03-01
Fixes location for type annotation on qualified names.

sormuras:
https://github.com/square/javapoet/issues/431

This solution overrides the `TypeName.toString()` in `ClassName` and `ParameterizedTypeName` - had to remove the final modifier from the former.

### DIFF
--- a/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
+++ b/src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
@@ -21,14 +21,17 @@ import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 import java.util.List;
+import java.util.Map;
+
 import org.junit.Test;
 
 public class AnnotatedTypeNameTest {
+  public @interface NeverNull {}
+  // Java 1.8 required for @Target(ElementType.TYPE_USE)
+  public @interface TypeUseAnnotation {}
 
   private final static String NN = NeverNull.class.getCanonicalName();
   private final AnnotationSpec NEVER_NULL = AnnotationSpec.builder(NeverNull.class).build();
-
-  public @interface NeverNull {}
 
   @Test(expected=NullPointerException.class) public void nullAnnotationArray() {
     TypeName.BOOLEAN.annotated((AnnotationSpec[]) null);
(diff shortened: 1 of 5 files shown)


sormuras:
- [x] `$T` works in `JavaFile`-context
- [ ] ParameterizedTypeName needs some love...

sormuras:
`type2` instance works as expected, but looks awkward:

```java
    TypeName type1 = ParameterizedTypeName.get(Map.Entry.class, Byte.class, Byte.class) 
        .annotated(typeUseAnnotation);

    ClassName raw = (ClassName) ClassName.get(Map.Entry.class).annotated(typeUseAnnotation);
    TypeName type2 = ParameterizedTypeName.get(raw, TypeName.BYTE.box(), TypeName.BYTE.box());
```

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -47,6 +50,27 @@
     assertEquals(annotated, annotated.annotated());
   }
 
+  // https://github.com/square/javapoet/issues/431
+  @Test public void annotatedNestedType() {
+    String tuacanon = TypeUseAnnotation.class.getCanonicalName();
```
swankjesse: just inline this and avoid the awkward name?

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -47,6 +50,27 @@
...
+    String expected = "java.util.Map.@" + tuacanon + " Entry";
```
swankjesse: Is this really what’s expected? This doesn't look right

swankjesse: wow!

sormuras: Have a look at https://github.com/square/javapoet/issues/431#issuecomment-190978668 - that's even more "wow!" and I didn't see this notation used in any Java source, yet.

sormuras:
`TypeName.toString()` is final again. Introduced a package-private `toString(out)` default implementation that is overridden in `ClassName` and `ParameterizedTypeName`.

Need to test @cushon s example from https://github.com/square/javapoet/issues/431#issuecomment-190978668 ...

sormuras:
- [ ] `FieldSpec.emit()` still uses fixed annotation location in front of `$T`
- [ ] `ParameterSpec.emit()` still uses fixed annotation location in front of `$T`
- [ ] `$T` rendering might fail in "do not import nothing" case

What else?

sormuras:
Will condense this PR down to (ignored) test cases - the fix feels like a band-aid approach. I think something like `ClassName.annotate(int indexOfSimpleNameToAnnotate, AnnotationSpec...)` would be more appropriate.

sormuras:
Tests moved to #433 
Fix needs more input in #431

PR #433, opened 2016-03-03
Two (ignored) tests for annotations on qualified names added

sormuras:
Added tests for https://github.com/square/javapoet/issues/431


swankjesse:
Thanks!

PR #434, opened 2016-03-04
JavaFile.skipImportsEntirely switch added.

sormuras:
Reasoning for having `JavaFile.skipImportsEntirely` switch:
- if there ever was an error in the internal import logic, the user could fall back to fully qualified type names immediatly, w/o having to wait for bug fix.
- eases testing the fix for #431 ...

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -57,6 +57,7 @@ public final class JavaFile {
   public final CodeBlock fileComment;
   public final String packageName;
   public final TypeSpec typeSpec;
+  public final boolean skipImportsEntirely;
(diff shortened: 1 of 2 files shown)


JakeWharton:
Not a fan of this change.

> if there ever was an error in the internal import logic, the user could fall back to fully qualified type names immediatly, w/o having to wait for bug fix.

This is a weird thing to code defensively against. Especially since it's so specific to one feature of the library when there's about 20 things that could go wrong at any time. Like, for example, the fact that type annotations are currently broken and there's no API to fix them somehow.

> eases testing the fix for #431

Not sure how it helps there. Are you just wanting fully-qualified references to be emitted? You can nest types with the names of the classes you want to import for that.

```java
class Test {
  private java.lang.@Redacted String name;

  static class String {}
}
```

PR #436, opened 2016-03-10
Simple fix for annotated qualified type names 1

sormuras:
Caveat: `ParameterizedTypeName` _projects_ its annotations to the underlying rawtype. See _awkward_ comment at https://github.com/square/javapoet/compare/master...sormuras:annotations_on_qualified_names_fix?expand=1#diff-0286fc2b37521e485d3c5ef3c31fa073R60


sormuras:
Outdated. Closing.

If this solution is chosen, additional tests and changes from #437 should be merged into this PR.

PR #437, opened 2016-03-10
Simple fix for annotated qualified type names

sormuras:
This PR changes the behaviour of annotating a `ParameterizedTypeName`: the annotations passed are collected by the underlying raw type. See https://github.com/square/javapoet/compare/master...sormuras:type_annotations_fix2?expand=1#diff-0286fc2b37521e485d3c5ef3c31fa073R52

### DIFF
--- a/src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
+++ b/src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
@@ -49,7 +49,8 @@ public final class ParameterizedTypeName extends TypeName {
   }
 
   @Override public ParameterizedTypeName annotated(List<AnnotationSpec> annotations) {
-    return new ParameterizedTypeName(rawType, typeArguments, concatAnnotations(annotations));
+    ClassName annotatedRawType = rawType.annotated(annotations);
+    return new ParameterizedTypeName(annotatedRawType, typeArguments);
   }
 
   @Override public TypeName withoutAnnotations() {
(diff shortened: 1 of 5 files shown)


JakeWharton:
Is this a dupe of #436? It's not exactly the same diff, but it's not really different.

sormuras:
Hence the almost same name. (-:

#436 does not change the way `ParameterizedTypeName` collects its annotations as before - but needs an awkward (package private) way to emit the annotations of the underlying raw type. See the `toString(out, annos)` method introduced in `TypeName`

This #437 _does_ change the collector of the annotations: it delegates it to the underlying rawtype.

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -49,7 +49,8 @@
   }
 
   @Override public ParameterizedTypeName annotated(List<AnnotationSpec> annotations) {
-    return new ParameterizedTypeName(rawType, typeArguments, concatAnnotations(annotations));
+    ClassName annotatedRawType = rawType.annotated(annotations);
```
sormuras: Here is the main difference in regard to #436 - the underlying raw type is annotated. This might be somehow ... unexpected. But simplifies the internal annotation emission, when compared to the solution proposed in #436

sormuras:
I like #437 more, as it mimics the following usage pattern:
```java
// from AnnotatedTypeNameTest#annotatedNestedParameterizedType()
TypeName type = ParameterizedTypeName.get(Map.Entry.class, Byte.class, Byte.class)
        .annotated(typeUseAnnotation);
```
is the same as
```java
ClassName raw = (ClassName) ClassName.get(Map.Entry.class).annotated(typeUseAnnotation);
TypeName annotatedByte = TypeName.BYTE.box().annotated(typeUseAnnotation); 
TypeName type = ParameterizedTypeName.get(raw, annotatedByte, TypeName.BYTE.box());
```
Which does yield a green light bulb, atm.

sormuras:
Everythings green again. Fixed a lot of false positive tests, see last comment at #431

sormuras:
An open task: how to handle a `$T` emission? See https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/CodeWriter.java#L234
May be, I could use the `TypeName.toString(out)` here as well, but I'm not sure about the side-effects.

sormuras:
Maybe an `AnnotatedIdentifier` type is a better solution than simply add annotations to the innermost name.

http://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-ReferenceType

In ClassType:
From `final List<String> names` to final List<AnnotatedIdentifier> names` with `AnnotatedIdentifier`:
```java
class AnnotatedIdentifier {
  List<AnnotationSpec> annotations;
  String name;
}
```
Close this PR?

sormuras:
See last comment - closing this PR.

PR #441, opened 2016-03-17
TypeVariableName.withBounds added

### LINKED ISSUE #440: Add TypeVariableName.withBounds
I use a constant for type variables and I want to add bounds when using in a type declaration.

e.g.,
```java
TypeSpec.classBuilder("Example")
    .addTypeVariable(tVar.withBounds(boundTypeName))
    .addField(tVar, "whatever", PRIVATE FINAL)
```

sormuras:
Closes #440

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeVariableName.java
+++ b/src/main/java/com/squareup/javapoet/TypeVariableName.java
@@ -57,6 +57,21 @@ public final class TypeVariableName extends TypeName {
     return new TypeVariableName(name, bounds);
   }
 
+  public TypeVariableName withBounds(Type... bounds) {
+    return withBounds(TypeName.list(bounds));
+  }
+
+  public TypeVariableName withBounds(TypeName... bounds) {
+    return withBounds(Arrays.asList(bounds));
+  }
+
+  public TypeVariableName withBounds(List<TypeName> bounds) {
+    ArrayList<TypeName> newBounds = new ArrayList<>();
+    newBounds.addAll(this.bounds);
+    newBounds.addAll(bounds);
+    return new TypeVariableName(name, newBounds, annotations);
+  }
+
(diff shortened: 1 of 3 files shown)


JakeWharton:
Test?

src/main/java/com/squareup/javapoet/TypeVariableName.java
```diff
@@ -57,6 +57,18 @@ private TypeVariableName(String name, List<TypeName> bounds, List<AnnotationSpec
     return new TypeVariableName(name, bounds);
   }
 
+  public TypeVariableName withBounds(Type... bounds) {
+    return get(name, bounds);
```
JakeWharton: This loses any existing bounds and annotations. I think you'll want to combine just like `annotated` does.

sormuras: Prepend or append new bounds?

JakeWharton: Append, but I don't think it matters. They should probably be a Set in actuality.

tbroyer: Aren't bounds supposed to have a class first (if any) followed by interfaces?
https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.4

sormuras:
Test! :smile:

src/main/java/com/squareup/javapoet/TypeVariableName.java
JakeWharton: True. We always do order-preserving sets, but I was thinking of a case where you're potentially building up interface-based bounds. Might as well leave it as a list until someone complains with an actual use case.

swankjesse:
Thanks!

PR #442, opened 2016-03-18
Annotation methods to be public, abstract and nothing else

Skelehog:
Other modifiers like static are not allowed for annotation
methods.

The exception message no longer suggests that
annotation methods cannot have modifiers, which is misleading
since public and abstract is in fact required.


swankjesse:
Thanks!

PR #443, opened 2016-03-19
Members with $L (and others) support by addStaticImport()

sormuras:
# Motivation

Current `JavaFile//addStaticImport` implementation "doesn't support members with $L". See https://github.com/square/javapoet/commit/4dd534906a1e8aa8b15883706878a45fdb1437da
## Features
- `$T.$L` and many (all ?) other cases are now supported
- It does not fail on intentional white spaces either, like `"$T . $N = PI"`
- Auto-skips unused `import static` requests
## Implementation outline
- Hooked into `lookup(ClassName)` and `emitAndIndent(String)` now.
- When a potential hit for a static import is detected, `lookup` returns an empty string and the `ClassName` instance is remembered as deferred.
- While there's a deferred `ClassName`, all strings (coming from any accelerator like `$L`, `$N`, ...) emitted via `emitAndIndent(String)` are gathered and analyzed for a wildcard or perfect match.
- Depending on the match result: if hit, emit  only the member part or emit the deferred class name.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TreeSet;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Modifier;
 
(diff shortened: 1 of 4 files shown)


jollytoad:
Hi, I've just hit the problem of ("$T.$N", SomeType.class, "methodName") not being supported by addStaticImport too. I've seen there has been many discussion around this, and a possible MemberRef. I'm just wondering what the current state is for this issue? Is it being worked on or has it stalled?

sormuras:
IIRC, last time I spoke with @swankjesse about it, he was almost okay with adding `$R`. Almost. :)

Regarding this PR: it only needs a little lift, and all is green again. Without introducing `$R`. But the integration into the emit-phase is not ... trivial. I guess, why this was never merged.

sormuras:
Rebased and conflicts resolved.

sormuras:
...and works like a charm with the corrected `TYPE_USE` annotations as well.

davidburstrom:
I'm missing this feature right now.

sormuras:
5 years ... and I'm missing it, too.

PR #444, opened 2016-03-20
make JavaFile#writeTo(Filer) thread-safe

gfx:
Parallel execution for `JavaPoet#writeTo(Filer)` makes annotation processing faster, but unfortunately `JavaFile#writeTo(Filer)` is not thread safe because `Filer#createSourceFile()` isn't.

This pull-request makes `JavaFile#writeTo(Filer)` thread-safe to synchronize calls of `Filer#createSourceFile()`.

NOTE: testing this PR is difficult, but I' made sure it works on my annotation processing library: https://github.com/gfx/Android-Orma/pull/230

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -33,6 +33,8 @@ import java.util.TreeSet;
 import javax.annotation.processing.Filer;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
+import javax.tools.FileObject;
+import javax.tools.JavaFileManager;
 import javax.tools.JavaFileObject;
 import javax.tools.JavaFileObject.Kind;
 import javax.tools.SimpleJavaFileObject;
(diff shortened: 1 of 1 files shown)


swankjesse:
What makes you say that Filer isn't thread safe? And if it isn’t, what makes you think that this is sufficient? Docs don't say anything either way.

gfx:
First, `Filer` is not thread safe because it raises errors when I use it in parallel.

I think the details is not defined in `Filer` specification, which is an implementation-defined class. In Oracle JDK and Open JDK, the concrete class of `Filer` is [`com.sun.tooks.javac.processing.JavacFiler`](http://hg.openjdk.java.net/jdk8u/jdk8u/langtools/file/8c3890c90147/src/share/classes/com/sun/tools/javac/processing/JavacFiler.java); I know the class from the stacktrace that I see when I use  `Filer#createSourceFile()` with parallel streams:

```
java.util.ConcurrentModificationException
	at com.sun.tools.javac.processing.JavacFiler.checkFileReopening(JavacFiler.java:531)
	at com.sun.tools.javac.processing.JavacFiler.createSourceOrClassFile(JavacFiler.java:404)
	at com.sun.tools.javac.processing.JavacFiler.createSourceFile(JavacFiler.java:378)
```

Second, because `JavacFiler.createSourceFile()` is not thread-safe and the details is undefined, it is safe to lock a static object, e.g. `JavaFile.class`. In fact, the code no longer throws `ConcurrentModificationException` with this PR equivalent.

swankjesse:
I think the right fix is a Filer equivalent to `Collections.synchronizedList()`. . . . a method that takes a `Filer` and returns a `Filer` that’s safe for concurrent use.

With this code you’ve picked `JavaFile.class` rather arbitrarily. Seems like the solution should be to make the Filer thread safe, not all the code that uses it.

swankjesse:
Is there a tracking bug for OpenJDK to either fix this or document it?

gfx:
Do you mean 3330a48 ? 

I can't find OpenJDK issues about this. It can be a nice chance for me to contribute OpenJDK :innocent:

swankjesse:
Yup, the synchronized filer is the right fix. But I think that should go in your codebase, not in JavaPoet. And I also think you should convince OpenJDK to fix their stuff to be thread safe.

swankjesse:
(We don't want to be responsible for ensuring thread safety of things like Filer.)

gfx:
Hmm, got it. Use the `SynchronizedFile` in my code. Thanks.

PR #445, opened 2016-03-20
add TypeName#unboxIfPossible()

gfx:
Sometimes I need an `unbox()` variation that doesn't throw UnsupportedOperationException. Here is an example: https://github.com/gfx/Android-Orma/blob/v2.3.5/processor/src/main/java/com/github/gfx/android/orma/processor/util/Types.java#L241-L263

I think it is a common need so I make this pull-request. Can you review it please?

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -156,6 +156,20 @@ public class TypeName {
    * @throws UnsupportedOperationException if this type isn't eligible for unboxing.
    */
   public TypeName unbox() {
+    TypeName unboxed = unboxIfPossible();
+    if (keyword != null || this != unboxed) {
+      return unboxed;
+    } else {
+      throw new UnsupportedOperationException("cannot unbox " + this);
+    }
+  }
+
+  /**
+   * Returns an unboxed type if this is a boxed primitive type (like {@code int} for {@code
+   * Integer}) or {@code Void}. Returns this type if it is already unboxed. Otherwise returns
+   * this as is.
+   */
+  public TypeName unboxIfPossible() {
     if (keyword != null) return this; // Already unboxed.
     if (this.equals(BOXED_VOID)) return VOID;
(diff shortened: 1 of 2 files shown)


swankjesse:
Just use `isPrimitive()` ? I have a tough time imagining code that benefits from this.

gfx:
Consider `ClassName.get(Integer.class)`. It is a box class type instead of a primitive type, and as far as I know there is no method to check if it is a box type of a primitive.

kenzierocks:
Also, this is useful if you're using the Stream API, because you can do `map(ClassName::unboxIfPossible)` rather than `map(className -> { try { return className.unbox(); } catch (Exception e) { return className; } })`

swankjesse:
Ahh, I see that. What does the calling code look like? Ie. when do you have `java.lang.Integer` and want `int` ?

tbroyer:
FWIW there are such cases in Immutables.org for instance (similar to AutoValue), where a method returning an `Optional<Float>` can generate a setter and/or wither taking a `float` as argument; and similarly for `List<Integer>` generating an _adder_ taking a `int` as argument.

I suppose we can find quite a few other cases where type arguments are unboxed when possible.

sormuras:
Before mapping both, primitive and wrapper types to some i/o handler, [stash](https://github.com/sormuras/stash) had this logic.
```java
      // Unbox, if possible.
      TypeName type = parameterSpec.type;
      try {
        type = type.unbox();
      } catch (UnsupportedOperationException e) {
        // ignore
      }
```
:+1: for merging

gfx:
I am developing an ORM which maps `int` and `Integer` to SQLite`s `INTEGER`, `float` and `Float` to `REAL`, and so on. In other words, the ORM regards box types as primitive types except for nullability.

swankjesse:
I like this, but I want something symmetric with the current API. Here’s the PR.
https://github.com/square/javapoet/pull/449

swankjesse:
Thanks!

gfx:
Looks good. Thanks.

Issue #446, opened 2016-03-21
Double quotes within string literals are no longer escaped as of 1.6.0


Wyverald:
The string literal

    This is "quoted"!

Is now presented as

    String s = "This is "quoted"!";

Instead of the intended

    String s = "This is \"quoted\"!";

sormuras:
I can not reproduce your issue.

If I add `stringLiteral("This is \"quoted\"!");` to [UtilTest:60](https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/UtilTest.java#L60), everything stays green.

Are you using the `$S` placeholder or `$L`?

Wyverald:
Except that's the wrong test. It should be `stringLiteral("This is \\\"quoted\\\"!", "This is \"quoted\"!", " ");`

sormuras:
Good catch - missed [case '\\\\'](https://github.com/square/javapoet/commit/c810627978b07289c8481d368f501de19c38f598#diff-fc9135b024f89c31f8c8fff2a7602f0aL494) when writing the test class and cleaning up the literal conversion.

swankjesse:
Ugh. Thanks for the fix. 1.6.1 is rolling out.

PR #447, opened 2016-03-21
Fix string literal

Wyverald:
#446


swankjesse:
Merged manually. Thanks!

PR #448, opened 2016-03-22
CodeBlock: add support for character literals with $C

Skelehog:
Character literals are escaped and wrapped in single quotes.
The CodeBlock documentation is updated.

An argument is a character only if it can be casted to
Character, otherwise an IllegalArgumentException is
thrown wrapping the ClassCastException for more information.

In case this PR is merged README should be updated after 
release with a $C for Characters section.


Skelehog:
I just realized that char literals should not be treated as a special case, rather there should be
an additional check in `emitLiteral` for `Character` objects.

There should be also a check for `Long` objects. Long literals must have an L after the value to avoid compiler error when the integer number is too large.

Skelehog:
I closed this because I thought it would be better to handle character and long literals with $L instead. But since $L is specified to be not escaped that may not work for you.

So maybe $C is required after all for character literals, too bad I already deleted my fork :-)

But at the very least I think long literals (they must contain an L after the value) should be handled with $L.

PR #450, opened 2016-03-24
fix dependency versions to test

gfx:
I couldn't build the head of JavaPoet with truth 1.0-SNAPSHOT and compile-testing 1.0-SNAPSHOT.


JakeWharton:
Ew, yeah. Those shouldn't be snapshots.

gfx:
Thanks to the quick merge :+1:

PR #452, opened 2016-03-24
Add LogMessage type to simplify creation of errors and warnings that reference types

ronshapiro:
Open to suggestions on naming, I threw this together pretty quickly. `PlainText`, `PlainTextMessage`?

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -37,23 +37,12 @@ import javax.tools.JavaFileObject;
 import javax.tools.JavaFileObject.Kind;
 import javax.tools.SimpleJavaFileObject;
 
+import static com.squareup.javapoet.Util.NULL_APPENDABLE;
 import static com.squareup.javapoet.Util.checkArgument;
 import static com.squareup.javapoet.Util.checkNotNull;
 
 /** A Java file containing a single top level class. */
 public final class JavaFile {
-  private static final Appendable NULL_APPENDABLE = new Appendable() {
-    @Override public Appendable append(CharSequence charSequence) {
-      return this;
-    }
-    @Override public Appendable append(CharSequence charSequence, int start, int end) {
-      return this;
-    }
-    @Override public Appendable append(char c) {
-      return this;
-    }
-  };
-
   public final CodeBlock fileComment;
   public final String packageName;
   public final TypeSpec typeSpec;
--- a/src/main/java/com/squareup/javapoet/LogMessage.java
+++ b/src/main/java/com/squareup/javapoet/LogMessage.java
@@ -1,0 +1,51 @@
+package com.squareup.javapoet;
+
+import java.io.IOException;
+import java.util.Collections;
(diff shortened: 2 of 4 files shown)


swankjesse:
Can it be a couple of methods instead of a new type?

```
  ClassName concreteType = ClassName.get("com.squareup.tacos", "ConcreteType");
  ClassName interfaceType = ClassName.get("com.squareup.tacos", "InterfaceType");
  CodeBlock codeBlock = CodeBlock.of("$T does not implement $T", concreteType, interfaceType);
  assertThat(codeBlock.toCompactString()).isEqualTo("ConcreteType does not implement InterfaceType");
  assertThat(codeBlock.imports()).containsExactly(concreteType, interfaceType);
```

ronshapiro:
`CodeBlock.imports()` could be confusing since they don't always have context.

```java
ClassName tacosType = ClassName.get("com.squareup.tacos", "Type");
ClassName burritosType = ClassName.get("com.squareup.burritos", "Type");
CodeBlock first = CodeBlock.of("$T first", tacosType);
CodeBlock second = CodeBlock.of("$T second", burritosType);
CodeBlock compound = CodeBlock.builder().append(first).append(" and ").append(second);
```

`first.imports()` and `second.imports()` both would import their single type, but `compound` would only import the first. Also for the typical use case of composing a java file, `imports()` doesn't seem relevant to the API.

Thinking about it now, I'd support having `CodeBlock` be an implementation detail of `LogMessage`, and just expose `.append()` methods on a builder or a `LogMessage.of()`. I know that `CodeBlock` says "Code blocks are not necessarily well-formed Java code" which makes them suitable for error messages, but it also states that it is a "fragment of a .java file", which messages are not. I could imagine that processor code that uses this would look odd by having `CodeBlock` in a validation phase and not a generation phase.

tbroyer:
+1 to CodeBlock as an implementation detail. That was my immediate reaction (but didn't have time to comment)

swankjesse:
Still not sure about this API. The LogMessage is the use case, but doesn't really describe the mechanism. (Sometimes that's okay).

```
public class CodeString {
  public final String code;
  public final Set<ClassName> referencedTypes;
  ...
}

public class CodeBlock {
  ...
  CodeString toCodeString();
}
```

Not sure about this either. Also unsure about how to name the imports. If there's a simple name collision do you want the full set? Or just the ones that were condensed?

ronshapiro:
I think just the names that were condensed, similar to how imports work for `JavaFile`. If there are conflicts, first wins and gets condensed.

`CodeString` is confusing with `CodeBlock` - I'd want something that conveys that the output is not intended to be placed in code. `TextMessage` (lol), `ClassNameText`, `SimpleNameString`?

swankjesse:
```
class CodeBlock {
  String compactString();
  Set<ClassName> compactStringImports();
  ...
}
```
?

swankjesse:
Closing ’cause we don’t have a good next step. If you think we’re okay just teaching `CodeBlock` to expose its compact contents & imports, let me know.

PR #455, opened 2016-03-31
Adding addPackageName method

genyherrera:
Adding addPackageName method and write the tests for this new method. Also adding the encoding on maven-compiler-plugin.

Now you can:

```java
JavaFile javaFile = JavaFile.builder("com.example.helloworld", helloWorld)
    .build();
```

Or:

```java
JavaFile javaFile = JavaFile.builder(helloWorld)
    .addPackageName("com.example.helloworld")
    .build();
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,7 @@
         <configuration>
           <source>${java.version}</source>
           <target>${java.version}</target>
+          <encoding>${project.build.sourceEncoding}</encoding>
         </configuration>
       </plugin>
 
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -204,6 +204,11 @@ public final class JavaFile {
     return new Builder(packageName, typeSpec);
   }
 
+  public static Builder builder(TypeSpec typeSpec) {
+    checkNotNull(typeSpec, "typeSpec == null");
+    return new Builder(typeSpec);
+  }
+
   public Builder toBuilder() {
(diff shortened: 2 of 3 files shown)


swankjesse:
Why?

genyherrera:
I was thinking...

I you want a class with default package, today you had to send an empty String on builder method to then don't translate to nothing. To me, that sounds wrong. You had to send something empty to do do nothing. 

I was think that if we overload the builder method to receive or not the packageName make more sense. Then you don't need to send an empty String to create the default package.

Also, we had method to add almost everything, but to add package is only on the builder method. Also I think that the addPackageName make more readable the code. It's clear that whatever information that you are sending on this new method, would be the package value. On the other hand, you had to know that the first parameter from builder method is the package.

This is exactly what I though, but is your call...

ronshapiro:
I don't think we should do anything to encourage people using the default
package.

genyherrera:
@ronshapiro 

Actually, my though for creating this method, was to bring the capability to add the packageName also as a method, and not only as a parameter from builder method. Is simply to add one more way to add the package to a class, and to me, a more readable way.

I never think in encourage people to use the default package, since I know that is not a good code convention, but at least I need to keep the compliance with previous version, so this is why I overload the builder method.

With my version, they don't need to set the packageName on builder method, but that don't means that I am encouraging to use the default package, it just IF you want to use the default package you don't need to send an empty String anymore. And if you want to declare the package you can now even on builder method or even a more readable solution on addPackageName method.

Regards.

swankjesse:
I think I wanna keep the existing API as is. Sorry!

Issue #456, opened 2016-04-01
Make CodeBlock extensible


Alexander--:
Not sure, how to do it most appropriately, but I'd like for CodeBlock to be extensible (or for some other recursively appendable entity to be added).

Here is an example of code I currently use:

```java
package com.example.customblock

import com.squareup.javapoet.CodeBlock;

import javax.lang.model.type.ArrayType;
import javax.lang.model.type.TypeMirror;

public class ArrayInitBlock {
    private final TypeMirror componentType;
    private final Object lengthLiteral;

    public ArrayInitBlock(TypeMirror component, Object lengthLiteral) {
        this.componentType = component;
        this.lengthLiteral = lengthLiteral;
    }

    @Override
    public String toString() {
        switch (componentType.getKind()) {
            case ARRAY:
                final TypeMirror subcomponentType =
                        ((ArrayType) componentType).getComponentType();

                final ArrayInitBlock nestedBlock =
                        new ArrayInitBlock(subcomponentType, lengthLiteral);

                return CodeBlock.builder()
                        .add("$L[]", nestedBlock)
                        .build()
                        .toString();
            default:
                return CodeBlock.builder()
                        .add("$T[$L]", componentType, lengthLiteral)
                        .build()
                        .toString();
        }
    }
}
```

It can be used with Javapoet like this:

```java
CodeBlock.builder()
  .addStatement("new $L", new ArrayInitBlock(resultType, length))
  .build()
  .toString();
```

…to emit something like:

```java
int[][][] array = new int[96][][];
```

Unfortunately, my quick hack has few downsides to it:

* Lots of String concatenation when widely employed
* No import handling and other cookies, inherent to CodeBlocks

Before you suggested to rewrite the code above without recursion: this is just part of my type-handling code, all of which is recursive (and can not be easily rewritten).

Furthermore there are few other cases, when I'd like to introduce custom CodeBlock subtypes (such as method calls and other expressions, which too can be nested and have associated types).

tbroyer:
You can probably rewrite your code to use `toCodeBlock` instead of `toString`, and therefore benefit from imports handling.

```java
CodeBlock.builder()
  .add("$[new").add(new ArrayInitBlock(resultType, length).toCodeBlock()).add(";\n$]")
  .build();
```

And JavaPoet could possibly be changed to treat a `CodeBlock` specially for a `$L`:

```java
// supposedly in CodeBlock.Builder#add(String,Object...), just before the 'switch (c)'
[...]

PR #457, opened 2016-04-01
Allow CodeBlock to be extended

Alexander--:
Fix for #456. The changes are trivial, so I presume, that CLA won't be needed.


swankjesse:
I think @tbroyer’s recommendation fits this better.

Issue #458, opened 2016-04-18
Inner class of parameterized type does not generate correct code


marcosb:
If we have something like:

class MyClass<T> {
  class InnerClass {}
}

There is currently no way to output MyClass<T>.Inner (ParameterizedTypeName doesn't have any way to specify this, nor does ClassName)

marcosb:
Woops, my brackets were swallowed by the github formatter.

No way to output `MyClass<T>.Inner`is what I meant.

tbroyer:
AFAIK, this is a known limitation; you'll use a `ParameterizedTypeName` for the outer class and then simple strings (simple name) for the inner classes.

marcosb:
Yes, the problem is TypeName.get(TypeMirror mirror, [other params]), which happens to be what I'm using, ends up translating to a raw type.

marcosb:
#460 fixes the issue

marcosb:
Slight update to the patch to actually fix the issue

marcosb:
Gentle ping - does this fix seem reasonable? I'm guessing there's a lot of code generators out there using javapoet to build code from a TypeMirror which are incorrectly generating raw types right now (that was the case with me, and we've gone many months w/o noticing)

marcosb:
Cleaner pull @ https://github.com/square/javapoet/pull/461 after talking to Christian

PR #460, opened 2016-04-19
Fix for https://github.com/square/javapoet/issues/458

marcosb:
Fix for https://github.com/square/javapoet/issues/458


src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -41,6 +72,31 @@
     TypeName.get(recursiveEnum.getGenericParameterTypes()[0]);
   }
 
+  @Test
```
marcosb: FYI this new tests fail before this change, i.e. the ParameterizedTypeName returned is a raw type of TestGeneric.Inner instead of a TestGeneric<String>Inner.

The other 2 additional tests are to make sure that the newly added code doesn't break old behavior.

PR #461, opened 2016-04-20
Avoid using raw types for SomeGeneric<T>.Inner

### LINKED ISSUE #458: Inner class of parameterized type does not generate correct code

marcosb:
Fixes #458


cgruber:
@swankjesse and @JakeWharton - any chance we could fast-track this?  It's a blocking bug for some of our internal code, if it's possible.  We're trying to keep from using internal forks where we can.

cgruber:
While I technically have rights to merge, I don't feel like I've been involved enough to do so at this point, without some Square review by familiar parties.

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -41,6 +69,31 @@
...
+    TypeName.get(genericStringInner.getGenericReturnType());
+    assertNotEquals(TypeName.get(genericStringInner.getGenericReturnType()),
+        TypeName.get(getClass().getDeclaredMethod("testGenericInnerInt").getGenericReturnType()));
+  }
+
+  @Test
+  public void innerStaticInGenericType() throws Exception {
+    Method staticInGeneric = getClass().getDeclaredMethod("testNestedNonGeneric");
+    TypeName.get(staticInGeneric.getReturnType());
+    TypeName.get(staticInGeneric.getGenericReturnType());
```
JakeWharton: I was going to comment how strange it is to see side-effect free statements being used to implicitly test behavior, but it looks like there's other test methods doing the same thing already. We should change these to assert _something_, but I'll leave that for a follow-up to fix everything.

JakeWharton: Filed as https://github.com/square/javapoet/issues/462

marcosb: Yah agreed, I thought the same as well, but wasn't sure what to test for (in my newly added test, the important thing is simply to make sure when the generic args are different, that we actually get different type args in the ParameterizedTypeName

marcosb:
@JakeWharton does it seem reasonable to commit this request, then? I realize this could break some code where raw types were incorrectly being used before, but it seems like TypeName.get(TypeMirror).rawType would work as a solution for places where it was the intended behavior (and where it wasn't intended, this will be a nice bug fix)

cgruber:
@JakeWharton, @swankjesse - any opinions on this p/r?

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +42,101 @@
 
   ParameterizedTypeName(ClassName rawType, List<TypeName> typeArguments,
       List<AnnotationSpec> annotations) {
+    this(
+        rawType,
+        rawType.enclosingClassName() != null ? rawType.enclosingClassName() : rawType,
```
swankjesse: This field definition seems strange. Why are these two interchangeable?

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +42,101 @@
...
+      List<AnnotationSpec> annotations) {
     super(annotations);
     this.rawType = checkNotNull(rawType, "rawType == null");
+    this.enclosingTypeOrRawType =
+        checkNotNull(enclosingTypeOrRawType, "enclosingTypeOrRawType == null");
+    this.nestedClassName = nestedClassName;
     this.typeArguments = Util.immutableList(typeArguments);
 
-    checkArgument(!this.typeArguments.isEmpty(), "no type arguments: %s", rawType);
+    checkArgument(
```
swankjesse: Ugh, it doesn’t seem like this model fits this use case very well. Sigh.

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +42,101 @@
...
+        enclosingTypeOrRawType);
     for (TypeName typeArgument : this.typeArguments) {
       checkArgument(!typeArgument.isPrimitive() && typeArgument != VOID,
           "invalid type parameter: %s", typeArgument);
     }
   }
 
   @Override public ParameterizedTypeName annotated(List<AnnotationSpec> annotations) {
-    return new ParameterizedTypeName(rawType, typeArguments, concatAnnotations(annotations));
+    return new ParameterizedTypeName(
```
swankjesse: nit: could you collapse all of these into fewer lines?

swankjesse: (the rest of this codebase doesn’t generally do line-per-item arguments or parameter declarations)

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
swankjesse: Could you just change this to be the enclosing type only, and we can use rawType if that’s null?

swankjesse: If you do that, then we don't need a member field for nestedClassName, we can just do what you do on 48 when we need that

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -87,7 +156,24 @@ public static ParameterizedTypeName get(ParameterizedType type) {
...
   /** Returns a parameterized type equivalent to {@code type}. */
   static ParameterizedTypeName get(ParameterizedType type, Map<Type, TypeVariableName> map) {
-    return new ParameterizedTypeName(ClassName.get((Class<?>) type.getRawType()),
-        TypeName.list(type.getActualTypeArguments(), map));
+    ParameterizedType toTest = type;
+    Deque<ParameterizedType> enclosingStack = new ArrayDeque<ParameterizedType>();
+    do {
+      enclosingStack.push(toTest);
+      toTest = (toTest.getOwnerType() instanceof ParameterizedType)
+          && !Modifier.isFinal(((Class<?>) type.getRawType()).getModifiers())
```
swankjesse: Don't you mean `isStatic` ?

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -87,7 +156,24 @@ public static ParameterizedTypeName get(ParameterizedType type) {
...
+          ? (ParameterizedType) toTest.getOwnerType() : null;
+    } while (toTest != null);
+
+    ParameterizedTypeName paramTypeName = null;
```
swankjesse: nit: please don’t invent abbreviations like `param` for things. Either drop the term completely `typeName` or spell it out completely.

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -87,7 +156,24 @@ public static ParameterizedTypeName get(ParameterizedType type) {
 
   /** Returns a parameterized type equivalent to {@code type}. */
   static ParameterizedTypeName get(ParameterizedType type, Map<Type, TypeVariableName> map) {
-    return new ParameterizedTypeName(ClassName.get((Class<?>) type.getRawType()),
-        TypeName.list(type.getActualTypeArguments(), map));
+    ParameterizedType toTest = type;
+    Deque<ParameterizedType> enclosingStack = new ArrayDeque<ParameterizedType>();
+    do {
```
swankjesse: Would it be simpler to implement this recursively?

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -265,13 +266,23 @@ static TypeName get(TypeMirror mirror,
...
+          return rawType;
+        }
 
         List<TypeName> typeArgumentNames = new ArrayList<>();
         for (TypeMirror mirror : t.getTypeArguments()) {
           typeArgumentNames.add(get(mirror, typeVariables));
         }
-        return new ParameterizedTypeName(rawType, typeArgumentNames);
+        return (enclosing instanceof ParameterizedTypeName)
+            ? ((ParameterizedTypeName) enclosing).nestedClass(rawType.simpleName())
```
swankjesse: did this drop the type arguments on the enclosed type?

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -32,6 +32,34 @@
...
+
+  protected static TestGeneric<String>.Inner testGenericStringInner() {
+    return null;
+  }
+
+  protected static TestGeneric<Integer>.Inner testGenericIntInner() {
+    return null;
+  }
+
+  protected static TestGeneric<String>.InnerGeneric<String> testGenericInnerString() {
```
swankjesse: Would it be prudent to use different type parameters? So if you have a bug that mixes them up, you catch it?

swankjesse: Like `TestGeneric<Short>.InnerGeneric<Long>` ?

src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -32,6 +32,34 @@
...
+    return null;
+  }
+
+  protected static TestGeneric<String>.InnerGeneric<Integer> testGenericInnerInt() {
```
swankjesse: ahh, here it is

src/test/java/com/squareup/javapoet/TypeNameTest.java
swankjesse: Maybe assert that the toString is reasonable?

swankjesse:
Good fix, but I’ve got lots of feedback. I think the set of fields is surprising. You could probably get by by just adding an enclosingType that’s possibly null.

I’m also wondering if it’s better to _always_ provide the enclosing type if the type is enclosed, but to make that a raw type if the nested class is static. That might be simpler.

marcosb:
@swankjesse thanks for the review! I believe I addressed all your suggestions, PTAL? (I updated the tests to check for something more reasonable; if you had something else in mind, could this be taken care of in #462?)

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
marcosb: Sorry about that; originally I completely removed rawType; but because it was public, there were things accessing it, so I had to add it back.  Updated to use rawType & cleaned up a bit around it.

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
marcosb: Woops, thanks.

marcosb:
@swankjesse missed your point about "I’m also wondering if it’s better to always provide the enclosing type if the type is enclosed" - as we only use the enclosing type in the case where it's parameterized & non-static, I'm not sure if it's much use for other cases (and then we'd have to check isStatic, which is difficult|impossible at this point because TypeName doesn't hold that info; so we'd have to have a new field passed in to the ctor to specify that it's a static & that the enclosing class should be ignored on output)

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +39,77 @@
 
   ParameterizedTypeName(ClassName rawType, List<TypeName> typeArguments,
       List<AnnotationSpec> annotations) {
+    this(rawType, null, typeArguments, annotations);
+  }
+
+  private ParameterizedTypeName(ClassName rawType, ParameterizedTypeName enclosingType,
```
swankjesse: nit: it’s weird that the arguments are (rawType, enclosingType) here, and the opposite in field declarations

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +39,77 @@
...
+        rawType, enclosingType, typeArguments, concatAnnotations(annotations));
   }
 
   @Override public TypeName withoutAnnotations() {
-    return new ParameterizedTypeName(rawType, typeArguments);
+    return new ParameterizedTypeName(
+        rawType, enclosingType, typeArguments, new ArrayList<AnnotationSpec>());
   }
 
   @Override CodeWriter emit(CodeWriter out) throws IOException {
```
swankjesse: nice

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -37,37 +39,77 @@
...
-    rawType.emit(out);
-    out.emitAndIndent("<");
-    boolean firstParameter = true;
-    for (TypeName parameter : typeArguments) {
-      if (!firstParameter) out.emitAndIndent(", ");
-      parameter.emitAnnotations(out);
-      parameter.emit(out);
-      firstParameter = false;
+    if (enclosingType != null) {
+      enclosingType.emitAnnotations(out);
```
swankjesse: Do we always want to get annotations from `rawType` ? That seems like it’s least likely to surprise?

swankjesse:
Merged manually!

swankjesse:
Thanks!

marcosb:
Thank you @swankjesse! I think Ron had asked already, but could we get a release w/ those fixes? Thanks!

swankjesse:
1.7 is out.

marcosb:
Sweet, thanks again!

PR #464, opened 2016-05-06
allow TypeName subclasses in withBounds() input list

### LINKED ISSUE #463: allow TypeVariableName.withBounds() to take TypeName subclasses as input
It'd be nice if this

```java
public TypeVariableName withBounds(List<TypeName> bounds);
```

could be something like this:

```java
public TypeVariableName withBounds(List<? extends TypeName> bounds);
```

I think this is the only method affected, all others take varargs rather than collections as input so we get covariance for free.

caplan:
Closes #463


swankjesse:
Thanks!

PR #469, opened 2016-05-30
Allow fields in annotations

### LINKED ISSUE #467: Cannot add fields to Annotations
I want to generate an Android StringDef like the following
```java
@StringDef({
		Foo.bar,
		Foo.baz
})
public @interface Foo {
	String bar = "bar";
	String baz = "baz";
}
```
Building the annotation with the annotation builder, I wrote
```java
TypeSpec.annotationBuilder("Foo")
				.addModifiers(Modifier.PUBLIC,Modifier.STATIC)
				.addField(FieldSpec.builder(String.class, "bar")
						.initializer("bar")
						.build())
				.addField(FieldSpec.builder(String.class, "baz")
						.initializer("baz")
						.build())
				.build();
```
However, the build fails with
> java.lang.IllegalStateException: ANNOTATION Foo cannot have fields

Disabling the check for fields on annotations generates valid code, and as far as I can tell annotations can have fields.

Is there another way to do this correctly, or is this a bug/limitation?

benjamin-bader:
Fixes #467.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -484,8 +484,7 @@ public final class TypeSpec {
     }
 
     public Builder addField(FieldSpec fieldSpec) {
-      checkState(kind != Kind.ANNOTATION, "%s %s cannot have fields", kind, name);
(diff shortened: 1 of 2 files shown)


JakeWharton:
Thanks!

Issue #470, opened 2016-06-01
Extending/implementing imports of the same name results in a name clash


christianbreynolds:
If I try to make an interface that extends another imported interface of the same name, the produced code has a clash of names. For instance, if I were writing a Person interface in a package com.example, and wanted to extend another Person interface in org.sample, I might write:

```
ClassName name = ClassName.get("org.sample", "Person");
TypeSpec type = TypeSpec.interfaceBuilder("Person").addSuperinterface(name).build();
JavaFile file = JavaFile.builder("com.example", type).build();
```

However, the JavaFile outputs: 

```
package com.example;

import org.sample.Person;

interface Person extends Person {
}
```

which isn't correct unless the line reads `interface Person extends org.sample.Person`. The same occurs for classes, and for already existing types (i.e. `TypeSpec.classBuilder("ArrayList").addSuperinterface(ArrayList.class).build();` )

tbroyer:
It looks like the test introduced in #471 passes only because the type is referenced (in the `Comparable<Taco>` superinterface). If you remove the reference to the type itself, then the bug is reproduced.

I've just been bitten by this, and am now trying to find the least ugly way to reference the type in the file.

PR #471, opened 2016-06-07
Fixed the name clash when implementing/extending classes

christianbreynolds:
Fixed the name clash that arises in this instance (https://github.com/square/javapoet/issues/470):

    TypeSpec ts = TypeSpec.interfaceBuilder("Example")
        .addSuperInterface(ClassName.get("org.pack", "Example")).build();
    System.out.println(JavaFile.Builder("com.ex", ts).build().toString());

Which prints the non-compiling:

    package com.ex;

    import org.pack.Example;

    interface Example extends Example {
    }

It now qualifies org.pack.Example properly:
`interface Example extends org.pack.Example {`

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -149,6 +149,7 @@ public final class TypeSpec {
     codeWriter.statementLine = -1;
 
     try {
+      codeWriter.pushType(this);
       if (enumName != null) {
         codeWriter.emitJavadoc(javadoc);
         codeWriter.emitAnnotations(annotations, false);
@@ -213,7 +214,7 @@ public final class TypeSpec {
         codeWriter.emit(" {\n");
       }
 
-      codeWriter.pushType(this);
+      //codeWriter.pushType(this);
       codeWriter.indent();
       boolean firstMember = true;
       for (Iterator<Map.Entry<String, TypeSpec>> i = enumConstants.entrySet().iterator();
@@ -285,13 +286,14 @@ public final class TypeSpec {
       }
 
       codeWriter.unindent();
-      codeWriter.popType();
+      //codeWriter.popType();
 
       codeWriter.emit("}");
       if (enumName == null && anonymousTypeArguments == null) {
         codeWriter.emit("\n"); // If this type isn't also a value, include a trailing newline.
       }
     } finally {
+      codeWriter.popType();
       codeWriter.statementLine = previousStatementLine;
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -213,7 +214,7 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         codeWriter.emit(" {\n");
       }
 
-      codeWriter.pushType(this);
+      //codeWriter.pushType(this);
```
JakeWharton: Just delete these

jbunting:
This is biting us. Is there anything I can do to push it forward?

JakeWharton:
lgtm. @swankjesse ?

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -622,6 +622,25 @@ private boolean isJava8() {
         + "}\n");
   }
 
+  @Test public void classImplementsExtendsSameName() throws Exception {
+    ClassName javapoetTaco = ClassName.get(tacosPackage, "Taco");
+    ClassName tacoBellTaco = ClassName.get("com.taco.bell", "Taco");
+    ClassName fishTaco = ClassName.get("org.fish.taco", "Taco");
```
swankjesse: omg I love it

swankjesse:
Thanks! Yep I have a bad backlog here. I hope to address that in the next week, and hopefully get a release out shortly thereafter.

jbunting:
Awesome, thanks guys!

swankjesse:
Update – I’m attempting the release and I had to roll this back because the new behavior is broken. But the test still passes, so maybe it’s no big deal. https://github.com/square/javapoet/pull/527

tbroyer:
@swankjesse It really looks to me like the fix here was correct. Could you give more info about what was broken that lead you to revert the change? (and as said on #470, the reason the test passes is because of `Comparable<Taco>`; remove it and the test fails because it'll import `com.fish.taco.Taco` and produce `class Taco extends Taco implements com.taco.bell.Taco`)

tbroyer:
See failing test in #559

tbroyer:
Oh, I see: the change broke `TypeSpecTest#classImplementsNestedClass`!

Should `CodeWriter` have a way to add the type being emitted to `referencedNames`? (maybe as part of an `emittingType()` method?) or should `TypeSpec` push an empty type with the same name (without nested types) when emitting the class "signature" and pop it before pushing the real type? Both would work (I verified).

PR #474, opened 2016-06-19
Allow to specify file encoding

### LINKED ISSUE #473: Allow to specify file encoding
Currently the file encoding is always the system default encoding. This is unreliable. It would be better to have a way to specify the encoding of the generated .java files.

marschall:
Add the ability to specify the file encoding.

Closes #473

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -19,9 +19,9 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
@@ -83,6 +83,14 @@ public final class JavaFile {
 
   /** Writes this to {@code directory} the standard directory structure. */
   public void writeTo(Path directory) throws IOException {
+      writeTo(directory, Charset.defaultCharset());
+  }
+
+  /**
+   * Writes this to {@code directory} the standard directory structure
+   * using {@code charset} (default if null).
+   */
+  public void writeTo(Path directory, Charset charset) throws IOException {
     checkArgument(Files.notExists(directory) || Files.isDirectory(directory),
         "path %s exists but is not a directory.", directory);
     Path outputDirectory = directory;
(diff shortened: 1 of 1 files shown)


marschall:
I agree to and have signed the CLA.

swankjesse:
When you customize the charset, is it to force UTF-8? Or is there another charset you’re using? I’m specifically wondering whether we should just change the defaults to always write UTF-8, since that’s the only good charset there is.

tbroyer:
Well, in an annotation processor, at least, JavaPoet should use the default encoding/charger, otherwise it could break the compilation. We can all agree that everyone should be using UTF-8, but this shouldn't be enforced (at least in this case).

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -94,7 +102,8 @@ public void writeTo(Path directory) throws IOException {
     }
 
     Path outputPath = outputDirectory.resolve(typeSpec.name + ".java");
-    try (Writer writer = new OutputStreamWriter(Files.newOutputStream(outputPath))) {
+    try (Writer writer = Files.newBufferedWriter(outputPath,
+            charset != null ? charset : Charset.defaultCharset())) {
```
tbroyer: Personal opinion: I'd rather have it throw if `null`, and have the overload explicitly pass the default charset in.

sormuras:
Imho, UTF-8 should be the default in JavaPoet.

Even if the encoding was set when writing, the annotation processor calling environment might disregard that and use whatever it likes: https://bugs.eclipse.org/bugs/show_bug.cgi?id=483690

tbroyer:
OK, disregard my comment, `writeTo(Filer)` uses `JavaFileObject#openWriter`. OK to default to UTF-8 for the `File` and `Path` overloads then; anyone wanting a different encoding could still use the `Appendable` overload. Sorry for the noise.

swankjesse:
Does anyone object to making JavaPoet force UTF-8, without the option for further configuration? As far as I know that should make everyone happy.

marschall:
Some of our projects are on UTF-8. Others are on ISO-8859-1. In this specific case it was to force UTF-8 because in Windows (:cry:) the default encoding was CP-1252.

AFAIK Eclipse defaults whatever the platform encoding is (see [bug 108668](https://bugs.eclipse.org/bugs/show_bug.cgi?id=108668)) so hardcoding UTF-8 would cause more configuration effort for Windows Eclipse users.

IMHO the ideal solution would be if JavaPoet generated [unicode escapes](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.3) for everything non-ASCII. That avoids all encoding issues (unless you use EBCDIC).

src/main/java/com/squareup/javapoet/JavaFile.java
marschall: fixed

swankjesse:
I’m still pretty undecided on this. Forcing UTF-8 seems reasonable, and forcing the platform’s default charset also seems reasonable. Would we ever want a different charset than those two? Right now we use the platform’s default charset.

swankjesse:
I’m forcing UTF-8.
https://github.com/square/javapoet/pull/522

PR #476, opened 2016-07-07
Public CodeWriter with Builder

wrmsr:
So I need to generate snippets of code, not entire JavaFiles. Unfortunately CodeWriter and its constructor are package-private. I can somewhat lamely get away with doing this by calling CodeBlock::toString which takes care of instantiating the writer for me but that doesn't let me control indentation (per #198 ) and it's generally not what that method is for. As I need non-default indentation I am currently having to do this via reflection which is... suboptimal :) Here I've made CodeWriter public and as is apparently idiomatic made a simple Builder for it. I looked over its public methods and, to me, they look sane to expose. Notable of course is the lack of defensive copying of the import Set and Map but that is unchanged from the existing code.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -187,7 +187,7 @@ public final class AnnotationSpec {
   @Override public String toString() {
     StringWriter out = new StringWriter();
     try {
-      CodeWriter codeWriter = new CodeWriter(out);
+      CodeWriter codeWriter = CodeWriter.builder(out).build();
       codeWriter.emit("$L", this);
       return out.toString();
     } catch (IOException e) {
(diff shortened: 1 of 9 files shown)


ronshapiro:
This is a similar idea to #452.

I don't think `CodeWriter` is the right level of abstraction though. If you had a `MethodSpec` and wanted to write it as a snippet, how would you do it? And what do you plan on doing with imported types?

One alternative could be a `PartialJavaFile` (with a better name) which allows you to include only certain aspects of what would normally go in a regular java file.

wrmsr:
Good point, to render the Spec's I'd make their emit's public as CodeWriter is itself now public. PartialJavaFile sounds too coarse for what I'm after - I want CodeBlock's explicit lack of any kind of guarantee about being well formed yet also its communication that it is intended to be some chunk of code (even if only, say, a single freestanding open brace) and not just some arbitrary string like an error message.

Regarding imported types I admittedly haven't meditated much on those, I just rephrased the existing code :|

swankjesse:
I still think CodeBlock is what we want for this case . . . it’s potentially misnamed but it intends to model a fragment of a `.java` file.

swankjesse:
No further action on this.

swankjesse:
(Thanks! But we wanna go with CodeBlock for this.)

PR #477, opened 2016-07-11
add README_CN.md

crazy1235:
add README_CN.md for Chinese

### DIFF
--- a/README_CN.md
+++ b/README_CN.md
@@ -1,0 +1,816 @@
+JavaPoet
+========
+
+`JavaPoet` 是一个用来生成 .java源文件的Java API。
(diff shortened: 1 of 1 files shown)


swankjesse:
Although I like this idea it grows the amount of stuff we’re responsible for.

I’m not currently prepared for the extra responsibility of additional languages in any official capacity. It is nice because it is inclusive, but it is not nice because it suggests things that I do not want to suggest In particular it implies we’ll be able to support issue reports in non-English languages, which we cannot do.

Please do write about the project in all of the languages you know! But I don’t want to include them in the project itself.

Issue #480, opened 2016-07-22
MethodSpec.overriding() for interface's default methods


yevhen13:
... is not always usable.

'default' modifier is being (automatically) added to the destination MethodSpec.Builder, and there is no way to remove it afterwards.

Therefore, if destination class is not an interface itself, then [_all hope is lost_](https://github.com/yevhen13/JFwdClassGen/issues/10).

yevhen13:
It seems like there should be a way of either:

- removing the modifier from the MethodSpec.Builder - e.g. _MethodSpec.Builder.removeModifier(modifier)_;
- or _somehow_ dropping the modifier in MethodSpec.overriding().

JakeWharton:
Default should be removed by this method. This is a trivial change. I can
send a PR tomorrow unless someone beats me to it.

PR #481, opened 2016-07-24
Remove DEFAULT modifier from overridden method. #480

sormuras:
Fix for #480 with test added.

### DIFF
--- a/src/test/java/com/squareup/javapoet/MethodSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/MethodSpecTest.java
@@ -39,6 +39,7 @@ import org.junit.Test;
 
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.TruthJUnit.assume;
 import static javax.lang.model.util.ElementFilter.methodsIn;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -145,6 +149,20 @@ private ExecutableElement findFirst(Collection<ExecutableElement> elements, Stri
         + "}\n");
   }
 
+  @Test public void overrideDoesNotCopyDefaultModifier() {
+    TypeElement classElement = getElement(ExtendsIterableWithDefaultMethods.class);
+    DeclaredType classType = (DeclaredType) classElement.asType();
+    List<ExecutableElement> methods = methodsIn(elements.getAllMembers(classElement));
+    ExecutableElement exec = findFirst(methods, "iterator");
+    Assume.assumeTrue(Util.DEFAULT != null);
```
tbroyer: `assumeNotNull`?

Also maybe `import static`? (and possibly use `TruthJUnit.assume()` instead: `assume().that(Util.DEFAULT).isNotNull()`)

sormuras: Done.

swankjesse:
Great!

Issue #378, opened 2015-12-14
Adding to equals() hashCode() toString() 


danielshaya:
Something that would be super useful would be the ability to have default implementations for equals() hashCode() and toString().  In the same way the IDEs generated them for you.  It could be added to the TypeSpec object with methods such as generateToString().

jbgi:
Probably more appropriate for a project like https://github.com/sormuras/poetry

sormuras:
True. ;-)

danielshaya:
I wasn't aware of this other project...
To be honest I think that having all the functions in one library makes it so much intuitive to use. I've used JavaPoet for a big project see my blog here (http://www.rationaljava.com/2015/12/7-tips-for-successful-code-generation.html) and apart from static imports it was perfect.  This suggested sort of added functionality would really take the project to the next level.

jbgi:
I am currently writing an annotation processor (https://github.com/derive4j/derive4j) based on a plugable architecture and javapoet is part of the API of the contract between the main annotation processor and its plugins. As such I would rather prefer javapoet to not be bloated with frequently updated utilities which would require me to update my processor-api contract too frequently (so that plugins can take advantage of added functionalities).

On the contrary, a separated artifact like poetry is better: it can be used by plugins through shadowing without impacting the contract or other plugins (that may use a different/incompatible version of poetry).

But it is only my pov.

danielshaya:
Sounds interesting but a corner case for JavaPoet.

I would have thought that most people who use JavaPoet are looking for a simple, clean, one-stop library to help them generate code. So generating default implementations equals() of hashcode() toString(), which is such a common task that IDEs support generation would be cool.  I have to say that having used JavaPoet on a fairly large project I can't think of anything else I would add at this stage.

But like you that only my pov.

sormuras:
@danielshaya `import static` is in 1.5.0 - and on topic, i added your generator idea to: https://github.com/sormuras/javapoet/wiki - let's implement 'em!

danielshaya:
Before I make a contribution can I ask with which version of the jdk are you compatible? Can I use Objects from Java 7?

jbgi:
yes, javapoet targets jdk7+

sormuras:
https://github.com/sormuras/poetry targets Java 8, btw.

bjakke:
This would be super cool feature! +1

[...]

PR #484, opened 2016-08-02
FieldSpec: add getters and setters

bart452:
- Simplify the procedure to set the standard getter and setter for
fields

### DIFF
--- a/src/main/java/com/squareup/javapoet/FieldSpec.java
+++ b/src/main/java/com/squareup/javapoet/FieldSpec.java
@@ -28,6 +28,7 @@ import javax.lang.model.element.Modifier;
 import static com.squareup.javapoet.Util.checkArgument;
 import static com.squareup.javapoet.Util.checkNotNull;
 import static com.squareup.javapoet.Util.checkState;
+import static com.squareup.javapoet.Util.toCamelCase;
 
 /** A generated field declaration. */
 public final class FieldSpec {
@@ -37,6 +38,8 @@ public final class FieldSpec {
   public final List<AnnotationSpec> annotations;
   public final Set<Modifier> modifiers;
   public final CodeBlock initializer;
+  public final MethodSpec getter;
+  public final MethodSpec setter;
 
   private FieldSpec(Builder builder) {
     this.type = checkNotNull(builder.type, "type == null");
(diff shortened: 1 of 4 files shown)


tbroyer:
How do you manage visibility of the accessors? No “helper” to create a getter or setter for a field inherited from a base class?

This would be better as helper methods creating `MethodSpec`s that you manually add to a `TypeSpec`.

bart452:
@tbroyer Thanks for the feedback, that might be a better idea, I'll update it.

sormuras:
#378 list more generatable `MethodSpec` candidates, based on fields with accessors.

bart452:
#387  Seems similar to what I want to do,  if this is done from `MethodSpec`, what would be the best way to specify what getter and setter you do and don't want?
That's why it might fit better in `FieldSpec`.

sormuras:
Let the user explicitely name the fields. Something like:
```java
FieldSpec name = ...
FieldSpec age = ...

TypeSpec.Builder builder = build("PersonBean");

Util.addFieldsAndGenerateAccessors(builder, name, age, ...);
Util.generateToStringUsingFields(buider, name, ...);
Util.generateEqualsAndHashcode(builder, name, age, ...);

TypeSpec personSpec = builder.build();

```
With `Util` being a placeholder for an arbitrary complex tool that cares for things @tbroyer mentioned. No need to integrate the desired state of turning a field to a bean property into the JavaPoet model.

bart452:
Thanks, seems good to me, will do that. 
Probably tomorrow though.

swankjesse:
I’m pretty reluctant on this one. I don’t think getters and setters make as much sense in 2016 as they did in 2001 when they first became popular.

Setters are good because they can implement field validation, but this code prevents that. Basically it makes the wrong thing easy.

Getters are good because they make it possible to change how the property is stored. But with mechanically generated getters that flexibility is lost.

I’d prefer to just have a helper method that does this external-to-Javapoet.

bart452:
I agree with you here but in some cases simple getters and setters do the job.
If there is some way to do it better, please let me know.
Maybe a `BeanSpec` or something could work.

swankjesse:
I think this is better as an independent class. Maybe in Poetry?
https://github.com/sormuras/poetry

Issue #485, opened 2016-08-04
New API: ClassName.getReflectionName()


swankjesse:
like `java.util.Map$Entry`.

sormuras:
Something like https://github.com/sormuras/poetry/blob/master/src/main/com/github/sormuras/poetry/Poetry.java#L135 ... on `ClassName` level?

sormuras:
- [x] `ClassName` #486
- [ ] `ArrayTypeName` is possible as well: https://github.com/sormuras/poetry/blob/master/src/main/com/github/sormuras/poetry/Poetry.java#L146
- [ ] `ParameterizedTypeName` use raw class type name?

tbroyer:
How about "binary name" rather than "reflection name"?

sormuras:
Or "class name", as it mimics the result of [`Class.getName()`](https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getName()) :smile: 

+1 for `binaryName()`

swankjesse:
Yeah, ugh. Sucks that there are two names. I’m not a fan of binary name ’cause that’s actually something like `Ljava/lang/String;`

gk5885:
The /-delimited form is just a [historical implementation detail](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.2).  "For historical reasons, the syntax of binary names that appear in class file structures differs from the syntax of binary names documented in [JLS §13.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.1)".  Users end up seeing that form in the context of JNI, but separately defined as a "[type signature](http://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/types.html#type_signatures)" (or [from the jvms](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.4)) in that context.  Throughout the JVMS, they're pretty clear drawing that distinction.  E.g. "The ClassName represents a binary class or interface name encoded in internal form"

Basically, I think we're perfectly safe to use "binary name" to mean the the thing from the JLS and referring to the other thing as a "binary name" is a misnomer.

PR #486, opened 2016-08-04
ClassName.reflectionName() added.

sormuras:
Solves #485


swankjesse:
That was quick. Thanks!

sormuras:
You're welcome. In fact, it was slow ... copied the code from an old PR filed earlier this year. :smiley: 

Did you notice the method name discussion in #485 and what about other `xxxName.reflectionName()`?

Issue #482, opened 2016-07-27
MethodSpec.overriding shouldn't copy annotations


gk5885:
For a few reasons:

* It's not necessarily semantically correct.  `@Nullable`, for example, might apply to the method being overridden, but not necessarily to the implementation of the override.
* It doesn't necessarily compile.  A private annotation on the method being overridden nested in the same type that declares the method is not accessible from a different type that might override the method.

sormuras:
Two valid points, Gregory.

What do you guys prefer:
  * replace annotation copy code by just adding an `@Override` tag
  * guard current implementation by boolean switches like:
```
Builder overriding(ExecutableElement method, boolean copyMethodAnnotations, boolean copyParameterAnnotations) {
```
  * or _fix_ by exposing `Builder` state fields and let the caller decide which annotation(s) to kick out? See #478 and #341 #342 and others.

gk5885:
My vote is for the first one.  It's the simple, safe, predictable behavior.

The second one, is unlikely to be all that useful.  You have to know something about the annotations on the method you're overriding in order to be able to predict if copying them is safe/correct.   If you've come that far you might as well just put them on the overriding method yourself.  There _might_ be a version that took a `Predicate` or something that might be useful, but that seems like overkill to me.

The third assumes that a blacklist (rather than a whitelist) is what you're going to want.  I think more often than not, it might be the other way around, but I think a `Predicate` or something similar might work better for that too.

HenryLoenwind:
Is it only me who finds it really annoying that copying a parameter WITH annotations now is a major effort that involves research into how to do that? `VariableElement`s are not the easiest thing to work with...

Aside from that, I find the behaviour of ParamSpec.get() very unexpected. I would have expected it to give me an object that contains the same information as its parameter, not some stripped down version. It's more ParamSpec.getSomethingThatLooksABitLike() now...

PR #487, opened 2016-08-05
Don't copy annotations in MethodSpec.overriding

gk5885:
…for the reasons cited in issue #482.

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -43,7 +43,6 @@ import static com.squareup.javapoet.Util.checkState;
 /** A generated constructor or method declaration. */
 public final class MethodSpec {
   static final String CONSTRUCTOR = "<init>";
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -198,12 +201,7 @@ public static Builder overriding(ExecutableElement method) {
     String methodName = method.getSimpleName().toString();
     MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(methodName);
 
-    methodBuilder.addAnnotation(OVERRIDE);
```
sormuras: Why remove this? The result actually _does_ override a method.

ronshapiro: He added `methodBuilder.addAnnotation(Override.class)` which has the same effect.

sormuras: Ah! Missed that line. Never mind. :)

swankjesse:
Merged manually.

PR #493, opened 2016-08-16
Add the possibility to use CodeBlock in format and adding statement of CodeBlock

lundibundi:
Just a few convenience changes that greatly helped me in my project, I've not encountered any problems with them. 
P.s. It's possible that I've missed something as I'm not very familiar with javapoet's codebase, so please point them out and I'll try to solve them.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -175,10 +175,9 @@ public final class CodeBlock {
           case 'T':
             this.args.add(argToType(args[index]));
             break;
-          case 'B': {
+          case 'B':
             add(argToCodeBlock(args[index]));
             break;
-          }
           default:
             throw new IllegalArgumentException(
                 String.format("invalid format string: '%s'", format));
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -175,6 +175,10 @@ public final class CodeBlock {
           case 'T':
             this.args.add(argToType(args[index]));
             break;
+          case 'B': {
+            add(argToCodeBlock(args[index]));
+            break;
+          }
           default:
             throw new IllegalArgumentException(
                 String.format("invalid format string: '%s'", format));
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -269,6 +278,13 @@ public Builder addStatement(String format, Object... args) {
       return this;
     }
 
+    public Builder addStatement(CodeBlock codeBlock) {
```
kenzierocks: I don't understand why you would ever need this -- you could just call addStatement on the codeBlock argument.

lundibundi: The thing is I create a code for statement through a few iterations, so I first use a codeblock for a few times and then need to wrap it up as a statement. Maybe I've missed a method to do this?

kenzierocks: Could you show me an example block of code that does that?

lundibundi: I have noticed that now that I don't use that method as I've replaced it with "format with CodeBlock s" but let assume we generate code with the following func (as usual smth you don't want to write by yourself)
```java
        public static CodeBlock composeObjectGetter(CodeBlock leftHandSide, CodeBlock identifier, String... properties) {
        CodeBlock.Builder builder = leftHandSide.toBuilder();
        for (int i = properties.length - 1; i >= 0; i--) {
            builder.add("(($T) ", ClassName.get(JSObject.class));
        }
        builder.add(identifier);
        for (int i = 0; i < properties.length; i++) {
            builder.add(OBJECT_GETTER, properties[i]);
        }
        return builder.build();
    }
```

then you might want to add it as Statement but I haven't found a method to do that. Is there any other way? btw at first I used String for generation and then just added it but unfortunately I didn't get any class imports with that and haven't found a way to import manually.

Now I generate only right hand side and the use something like `builder.addStatement("$L = ($B).something()", name, codeblock)`

kenzierocks: In that case I usually just put the `';\n'` myself instead of using addStatement.

lundibundi: What about proper indentation?

kenzierocks: There's no indentation to worry about? It always came out correct, because it's a single line.

lundibundi: Sorry for the late response. It's not really important for me whether we include that function or not, so if it's necessary I'll remove it. 

A few notes about those lines:
- It looks much better: `builder.addStatement(codeBlock)` vs `codeBlockBuilder.add(";\n")` and then `builder.addCode(codeBlock)`
- It gives a system more information about our code (I'm talking about `add("$[")` and `add(";\n$]")`

swankjesse:
I think you want `$L`.

PR #503, opened 2016-09-02
Check kind state when adding super class. #502

### LINKED ISSUE #502: Throw when superclass is used on an interface.
This is currently silently ignored.

sormuras:
`superclass` needs `Kind.CLASS`

Closes #502.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -2014,6 +2014,24 @@ public final class TypeSpecTest {
     }
   }
 
+  @Test public void superClassOnlyValidForClasses() {
+    try {
+      TypeSpec.annotationBuilder("A").superclass(ClassName.get(Object.class));
+      fail();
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -2014,6 +2014,25 @@ private String toString(TypeSpec typeSpec) {
     }
   }
 
+  @Test public void superClassOnlyValidForClasses() {
+    TypeSpec.classBuilder("C").superclass(Object.class);
```
JakeWharton: This line isn't needed. It's already tested elsewhere to be working.

JakeWharton:
lgtm

src/test/java/com/squareup/javapoet/TypeSpecTest.java
sormuras: Line removed.

JakeWharton:
Thanks!

PR #509, opened 2016-09-29
add multy class support

JackyAnn:
generate code such as:HelloWorld.java
import java.io.Serializable;
public class HelloWorld {

}
abstract class Parent {
  protected abstract String getMessage();
}
abstract class Child {
}

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -24,6 +24,7 @@ import java.io.Writer;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -56,7 +57,7 @@ public final class JavaFile {
 
   public final CodeBlock fileComment;
   public final String packageName;
-  public final TypeSpec typeSpec;
+  public final List<TypeSpec> typeSpecs;
   public final boolean skipJavaLangImports;
   private final Set<String> staticImports;
   private final String indent;
@@ -64,7 +65,7 @@ public final class JavaFile {
   private JavaFile(Builder builder) {
     this.fileComment = builder.fileComment.build();
     this.packageName = builder.packageName;
-    this.typeSpec = builder.typeSpec;
+    this.typeSpecs = Util.immutableList(builder.typeSpecs);
     this.skipJavaLangImports = builder.skipJavaLangImports;
     this.staticImports = Util.immutableSet(builder.staticImports);
     this.indent = builder.indent;
@@ -93,7 +94,7 @@ public final class JavaFile {
       Files.createDirectories(outputDirectory);
     }
 
-    Path outputPath = outputDirectory.resolve(typeSpec.name + ".java");
+    Path outputPath = outputDirectory.resolve(typeSpecs.get(0).name + ".java");
     try (Writer writer = new OutputStreamWriter(Files.newOutputStream(outputPath))) {
       writeTo(writer);
     }
@@ -107,9 +108,12 @@ public final class JavaFile {
   /** Writes this to {@code filer}. */
   public void writeTo(Filer filer) throws IOException {
     String fileName = packageName.isEmpty()
-        ? typeSpec.name
-        : packageName + "." + typeSpec.name;
-    List<Element> originatingElements = typeSpec.originatingElements;
+        ? typeSpecs.get(0).name
+        : packageName + "." + typeSpecs.get(0).name;
+    List<Element> originatingElements = new ArrayList<>();
+    for(TypeSpec typeSpec : typeSpecs){
+		originatingElements.addAll(typeSpec.originatingElements);
+    }
     JavaFileObject filerSourceFile = filer.createSourceFile(fileName,
         originatingElements.toArray(new Element[originatingElements.size()]));
     try (Writer writer = filerSourceFile.openWriter()) {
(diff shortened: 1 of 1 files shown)


JackyAnn:
Multiple class declarations in one .java file
sample:
        
        String PACKAGE = "com.example.generate";
        TypeSpec child = TypeSpec.classBuilder(ClassName.get(PACKAGE, "Child"))
                                  .addModifiers(Modifier.ABSTRACT)
                                  .build();

        MethodSpec getMessage = MethodSpec.methodBuilder("getMessage")
                                          .addModifiers(Modifier.PROTECTED)
                                          .addModifiers(Modifier.ABSTRACT)
                                          .returns(String.class)
                                          .build();


        TypeSpec parent = TypeSpec.classBuilder(ClassName.get(PACKAGE, "Parent"))
                                  .addModifiers(Modifier.ABSTRACT)
                                  .addMethod(getMessage)
                                  .build();
        FieldSpec  message = FieldSpec.builder(String.class, "message")
                                      .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
                                      .addModifiers(Modifier.FINAL)
                                      .initializer("\"Hello, JavaPoet!\"")
                                      .build();
        MethodSpec getMessageOvrerride = MethodSpec.methodBuilder("getMessage")
                                                   .addAnnotation(Override.class)
                                                   .addModifiers(Modifier.PROTECTED)
                                                   .returns(String.class)
                                                   .addStatement("return message")
                                                   .addModifiers()
                                                   .build();
        MethodSpec main = MethodSpec.methodBuilder("main")
                                    .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
                                    .returns(void.class)
                                    .addParameter(String[].class, "args")
                                    .addStatement("System.out.println(\"Hello, JavaPoet!\")")
                                    .build();
        TypeSpec typeSpec = TypeSpec.classBuilder("HelloWorld")
                                    .addModifiers(Modifier.FINAL)
                                    .superclass(ClassName.get(PACKAGE, parent.name))
                                    .addSuperinterface(Serializable.class)
                                    //.addType(parent)
                                    .addField(message)
                                    .addMethod(main)
                                    .addMethod(getMessageOvrerride)
                                    .build();

        JavaFile javaFile = JavaFile.builder(PACKAGE, typeSpec).addType(parent).addType(child).build();
        System.out.println(javaFile.toString());

the generate code is:
package com.example.generate;

import java.io.Serializable;
import java.lang.Override;
import java.lang.String;

public final class HelloWorld extends Parent implements Serializable {
  public static final String message = "Hello, JavaPoet!";

  public static void main(String[] args) {
    System.out.println("Hello, JavaPoet!");
  }

  @Override
  protected String getMessage() {
    return message;
  }
}
abstract class Parent {
  protected abstract String getMessage();
}
abstract class Child {
}

JakeWharton:
What advantages does this have over separate files?

sockeqwe:
@JakeWharton For example `AutoValueExtension` has one method `String generateClass()` which expects one String as return value containing the code for the generated class. The problem is that there is no simple way (as far as I know) to generate 2 classes as two different files in AutoValue  (except inner classes).


```java 
  @Override
    public String generateClass(
            Context context, String className, String classToExtend, boolean isFinal) {

        TypeSpec subclass = generateSomeClass();
        return JavaFile.builder(context.packageName(), subclass).build().toString();
    }
```

My (naive) solution would be to generate one java file containing one public class and the second class in the same java file (with package private visibility which would be ok in my use case). 

Is there already a way to do that with JavaPoet or do you know a better way to generate 2 classes (not inner classes) out of one `@AutoValue` annotated class?

swankjesse:
I think that’s a problem with the AutoValue API. Defining multiple top-level classes in a single `.java` file is a bad idea, and causes grief for incremental compilers and other tools. I do not wanna support that in JavaPoet!

swankjesse:
http://stackoverflow.com/a/2336762/40013

JakeWharton:
Yeah that's an interesting use case. Is there another reason you don't want
to nest? Are both classes referenced by non-generated code?

sockeqwe:
I agree! The stackoverflow answer gives an excellent explanation (I haven't thought about incremental compilation yet and I'm wondering how kotlin compiler deals with that, but that is off topic). I also consider multiple class declaration in one file as bad practice. It was just the first "workaround" that came in my mind for the autovalue API limitation, but you are absolutely right that this is not JavaPoet's problem. I was just trying to give a (very rare) use case when someone might need multiple java class definition in one file but you definitely have good arguments to not support that in JavaPoet.

eamonnmcmanus:
The AutoValue extension API makes it easy for you to return the definition of a class that goes into the inheritance hierarchy underneath the original `@AutoValue` class. But it also gives you [access](https://github.com/google/auto/blob/facfc7e62777310282d9be1870e4b00facf692b1/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java#L62) to a [ProcessingEnvironment](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/ProcessingEnvironment.html) and you can use the [Filer](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/Filer.html) from that ProcessingEnvironment to generate new classes that are not in that inheritance hierarchy.

sockeqwe:
@JakeWharton It's a long story. In short: I use a third party library that uses reflection to load a class with a specific name (therefore I can't nest the class) and I would like to build some kind of bridge between auto value and this third party library. As already said, a very rare use case.

@eamonnmcmanus Excellent! Completely overlooked that. Thank you very much!

PR #513, opened 2016-10-24
Add methods for controlling scope of variables

int02h:
Utility methods that help to wrap code blocks with braces to indent variable scope:
```java
{
  int i = 0;
  String s = "";
  // and so on
}
```

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -262,6 +262,18 @@ public final class CodeBlock {
       return this;
     }
 
+    public Builder beginScope() {
+      add("{\n");
+      indent();
+      return this;
+    }
+
+    public Builder endScope() {
+      unindent();
+      add("}\n");
+      return this;
+    }
+
     public Builder addStatement(String format, Object... args) {
       add("$[");
       add(format, args);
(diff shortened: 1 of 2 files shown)


swankjesse:
I don’t think this is going to be very common.

int02h:
Yes, I'm agree. But when it's necessary there is no way to do such thing. Method 'addCode' do not indent following code. Method 'beginControlFlow' inserts one space before opening brace. It looks ugly. Maybe it is more preferable to modify 'beginControlFlow' method to handle situation with whitespace at the begging of the line.

tbroyer:
If it's only a matter of indentation (which I think it is), then use `$>` and `$<`.

```
b.addCode("{\n$>");
b.addCode("int i = 0;\n");
…
b.addCode("$<}");
```

int02h:
Ok. You're right.

ronshapiro:
Is using "$[" and "$]" part of the public API? If so you could use that.

JakeWharton:
Can you do `addCode("{$[\n")` and `addCode("$]}\n")`?

swankjesse:
Yep. Public API and documented here: https://square.github.io/javapoet/1.x/javapoet/com/squareup/javapoet/CodeBlock.html

Issue #518, opened 2016-11-02
Support named parameters


cconroy:
The positional parameter API can become unwieldy in long statements or code blocks. 

If we had named parameters, then a lot of generated code could more easily be represented by block templates instead of statement-by-statement poet API calls.

The exact syntax is TBD, but something like the following:

```
$someType:T $someName:N = new $someType:T();
while (etc) {
  $someName.foo();
  ...
}
```

swankjesse:
I particularly like this because it gets us the best of both worlds – JavaPoet’s composability, plus the code-familiarity you usually get with template systems. We might even convince our AutoValue friends to use this!

swankjesse:
Shipped!

PR #520, opened 2016-11-08
Add support for named arguments to CodeBlock

cconroy:
Relative and positional arguments are fine for small statements, but
larger code blocks benefit from having named arguments.

https://github.com/square/javapoet/issues/518

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
     .build();
 ```
 
+### Code block format strings
+Code blocks may specify the values for their placeholders in a few
+ways. Only one style may be used for each addition to a code block.
+
+#### Relative Arguments
+Pass an argument value for each placeholder in the format string to `CodeBlock.add`
+
+In each example, we generate code to say "I ate 3 tacos"
+
+```java
+CodeBlock.builder().add("I ate $L $L", 3, "tacos")
+```
+
+#### Positional Arguments
+Place an integer index after the placeholder in the format string to specify which argument to use.
+
+```java
+CodeBlock.builder().add("I ate $L1 $L0", "tacos", 3)
+```
+
+#### Named Arguments
+Use the syntax `$argumentName:X` where X is the formatc haracter and
+call `CodeBlock.addNamed` with a map containing all argument keys in the
+format string.
+
+```java
+Map<String, Object> map = new HashMap<>();
+map.put("food", "tacos");
+map.put("count", 3);
+CodeBlock.builder().addNamed("I ate $count:L $food:L", map)
+```
+
 ### Methods
 
 All of the above methods have a code body. Use `Modifiers.ABSTRACT` to get a method without any
--- a/README.md
+++ b/README.md
@@ -422,7 +422,7 @@ CodeBlock.builder().add("I ate $2L $1L", "tacos", 3)
 #### Named Arguments
 Use the syntax `$argumentName:X` where X is the format character and
 call `CodeBlock.addNamed` with a map containing all argument keys in the
-format string.
+format string. Argument names must begin with a lowercase character.
 
 ```java
 Map<String, Object> map = new LinkedHashMap<>();
--- a/README.md
+++ b/README.md
@@ -166,10 +166,10 @@ public static void main(String[] args) throws Exception {
       .addMethod(whatsMyName("eminem"))
       .addMethod(whatsMyName("marshallMathers"))
       .build();
-      
+
   JavaFile javaFile = JavaFile.builder("com.example.helloworld", helloWorld)
       .build();
-      
+
   javaFile.writeTo(System.out);
 }
 
@@ -210,15 +210,15 @@ MethodSpec today = MethodSpec.methodBuilder("today")
     .returns(Date.class)
     .addStatement("return new $T()", Date.class)
     .build();
-    
+
 TypeSpec helloWorld = TypeSpec.classBuilder("HelloWorld")
     .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
     .addMethod(today)
     .build();
-    
+
 JavaFile javaFile = JavaFile.builder("com.example.helloworld", helloWorld)
     .build();
-    
+
 javaFile.writeTo(System.out);
 ```
 
@@ -387,7 +387,7 @@ MethodSpec hexDigit = MethodSpec.methodBuilder("hexDigit")
     .returns(char.class)
     .addStatement("return (char) (i < 10 ? i + '0' : i - 10 + 'a')")
     .build();
-    
+
 MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
     .addParameter(int.class, "b")
     .returns(String.class)
@@ -412,19 +412,20 @@ CodeBlock.builder().add("I ate $L $L", 3, "tacos")
 ```
 
 #### Positional Arguments
-Place an integer index after the placeholder in the format string to specify which argument to use.
+Place an integer index (1-based) before the placeholder in the format string to specify which
+argument to use.
 
 ```java
-CodeBlock.builder().add("I ate $L1 $L0", "tacos", 3)
+CodeBlock.builder().add("I ate $2L $1L", "tacos", 3)
 ```
 
 #### Named Arguments
-Use the syntax `$argumentName:X` where X is the formatc haracter and
+Use the syntax `$argumentName:X` where X is the format character and
 call `CodeBlock.addNamed` with a map containing all argument keys in the
 format string.
 
 ```java
-Map<String, Object> map = new HashMap<>();
+Map<String, Object> map = new LinkedHashMap<>();
 map.put("food", "tacos");
 map.put("count", 3);
 CodeBlock.builder().addNamed("I ate $count:L $food:L", map)
@@ -439,7 +440,7 @@ body. This is only legal if the enclosing class is either abstract or an interfa
 MethodSpec flux = MethodSpec.methodBuilder("flux")
     .addModifiers(Modifier.ABSTRACT, Modifier.PROTECTED)
     .build();
-    
+
 TypeSpec helloWorld = TypeSpec.classBuilder("HelloWorld")
     .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
     .addMethod(flux)
@@ -472,7 +473,7 @@ MethodSpec flux = MethodSpec.constructorBuilder()
     .addParameter(String.class, "greeting")
     .addStatement("this.$N = $N", "greeting", "greeting")
     .build();
-    
+
 TypeSpec helloWorld = TypeSpec.classBuilder("HelloWorld")
     .addModifiers(Modifier.PUBLIC)
     .addField(String.class, "greeting", Modifier.PRIVATE, Modifier.FINAL)
@@ -529,7 +530,7 @@ Like parameters, fields can be created either with builders or by using convenie
 FieldSpec android = FieldSpec.builder(String.class, "android")
     .addModifiers(Modifier.PRIVATE, Modifier.FINAL)
     .build();
-    
+
 TypeSpec helloWorld = TypeSpec.classBuilder("HelloWorld")
     .addModifiers(Modifier.PUBLIC)
     .addField(android)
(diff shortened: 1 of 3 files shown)


cconroy:
cc: @gmalmquist

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -81,27 +83,82 @@
...
+    map.put("text", "taco");
+    CodeBlock block = CodeBlock.builder().addNamed("$text:S", map).build();
+    assertThat(block.toString()).isEqualTo("\"taco\"");
+  }
+
+  @Test public void repeatedNamedArgument() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("text", "tacos");
+    CodeBlock block = CodeBlock.builder()
+        .addNamed("\"I like \" + $text:S + \". Do you like \" + $text:S + \"?\"", map)
```
swankjesse: Yes!

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -81,27 +83,82 @@
...
+
+  @Test public void namedAndNoArgFormat() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("text", "tacos");
+    CodeBlock block = CodeBlock.builder()
+        .addNamed("$>\n$text:L for $$3.50", map).build();
+    assertThat(block.toString()).isEqualTo("\n  tacos for $3.50");
+  }
+
+  @Test public void missingNamedArgument() {
```
swankjesse: dig

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -81,27 +83,82 @@
...
+    try {
+      Map<String, Object> map = new HashMap<>();
```
swankjesse: supernit: HashMap is [dead](https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/) to me, and it should be dead to you too. LinkedHashMap is the only way to live.

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -81,27 +83,82 @@
...
+    }
+  }
+
+  @Test public void multipleNamedArguments() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("pipe", System.class);
+    map.put("text", "tacos");
+
+    CodeBlock block = CodeBlock.builder()
+        .addNamed("$pipe:T.out.println(\"Let's eat some $text:L\");", map)
```
swankjesse: off topic: right now we’re pretty dumb and won’t do the right thing if you embed `$L` inside a quoted string. Perhaps something to consider fixing

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -200,6 +245,31 @@ public Builder add(String format, Object... args) {
       return this;
     }
 
+    private boolean isNoArgPlaceholder(char c) {
```
swankjesse: dig

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -200,6 +245,31 @@ public Builder add(String format, Object... args) {
...
+      return c == '$' || c == '>' || c == '<' || c == '[' || c == ']';
+    }
+
+    private void addArgument(String format, char c, Object arg1) {
+      Object arg = arg1;
```
swankjesse: why?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +118,64 @@ public Builder toBuilder() {
...
+              "unknown format $%s at %s in '%s'", format.charAt(p + 1), p + 1, format);
+          formatParts.add(format.substring(p, p + 2));
+          p += 2;
+        }
+      }
+
+      return this;
+    }
+
+    /**f
```
swankjesse: f

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +118,64 @@ public Builder toBuilder() {
...
 
+    /**
+     * Adds code using named arguments.
+     *
+     * <p> Named arguments specify their name after the '$' followed by : and the corresponding
+     * type character. For example, to refer to the class {@link java.lang.String} with the argument
+     * name 'text' use a format string containing {@code $text:S} and include the key 'text' with
+     * value {@link java.lang.String} in the argument map.
+     */
+    public Builder addNamed(String format, Map<String, Object> arguments) {
```
swankjesse: Probably easier to call than `Map<String, Object>` is `Map<String, ?>`. That way callers can provide a `Map<String, String>` and they won’t get a type error.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +118,64 @@ public Builder toBuilder() {
     private Builder() {
     }
 
+    /**
+     * Adds code using named arguments.
+     *
+     * <p> Named arguments specify their name after the '$' followed by : and the corresponding
+     * type character. For example, to refer to the class {@link java.lang.String} with the argument
+     * name 'text' use a format string containing {@code $text:S} and include the key 'text' with
```
swankjesse: this is a really distracting example! Maybe simpler to give an example that doesn’t require a type conversion?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +118,64 @@ public Builder toBuilder() {
...
+        if (matcher.matches()) {
+          String argumentName = matcher.group("argumentName");
+          checkArgument(arguments.containsKey(argumentName), "Missing named argument for $%s",
+              argumentName);
+          char formatChar = matcher.group("typeChar").charAt(0);
+          addArgument(format, formatChar, arguments.get(argumentName));
+          formatParts.add("$" + formatChar);
+          p += matcher.regionStart() + argumentName.length() + 3;
+        } else {
+          checkArgument(p < format.length() - 1, "dangling $ at end");
```
swankjesse: test for this?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +118,64 @@ public Builder toBuilder() {
...
+        }
+        Matcher matcher = NAMED_ARGUMENT.matcher(format.subSequence(p, format.length()));
+        if (matcher.matches()) {
+          String argumentName = matcher.group("argumentName");
+          checkArgument(arguments.containsKey(argumentName), "Missing named argument for $%s",
+              argumentName);
+          char formatChar = matcher.group("typeChar").charAt(0);
+          addArgument(format, formatChar, arguments.get(argumentName));
+          formatParts.add("$" + formatChar);
+          p += matcher.regionStart() + argumentName.length() + 3;
```
swankjesse: supercool

README.md
```diff
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
     .build();
 ```
 
+### Code block format strings
+Code blocks may specify the values for their placeholders in a few
+ways. Only one style may be used for each addition to a code block.
+
+#### Relative Arguments
```
swankjesse: thanks for completing these cases

README.md
```diff
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
...
+
+```java
+CodeBlock.builder().add("I ate $L $L", 3, "tacos")
+```
+
+#### Positional Arguments
+Place an integer index after the placeholder in the format string to specify which argument to use.
+
+```java
+CodeBlock.builder().add("I ate $L1 $L0", "tacos", 3)
```
swankjesse: these are actually 1-indexed and the char comes after the digit.

```
CodeBlock.builder().add("I ate $2L $1L", "tacos", 3)
```

README.md
```diff
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
...
+```
+
+#### Named Arguments
+Use the syntax `$argumentName:X` where X is the formatc haracter and
```
swankjesse: haracter

README.md
```diff
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
...
+call `CodeBlock.addNamed` with a map containing all argument keys in the
+format string.
+
+```java
+Map<String, Object> map = new HashMap<>();
```
swankjesse: LinkedHashMap

README.md
```diff
@@ -398,6 +398,38 @@ MethodSpec byteToHex = MethodSpec.methodBuilder("byteToHex")
...
+map.put("food", "tacos");
+map.put("count", 3);
+CodeBlock.builder().addNamed("I ate $count:L $food:L", map)
```
swankjesse: I wonder if `addNamed()` is the right method name, since we have `$N` which pulls the name of the thing. 

Hmmm.

Maybe just overload `add()` ? It’s very unlikely that anyone would pass a `Map` to the existing `add()` method... hmm.

src/test/java/com/squareup/javapoet/CodeBlockTest.java
cconroy: tru

src/main/java/com/squareup/javapoet/CodeBlock.java
cconroy: whoops! clearly i got distracted while in the process of moving this around

README.md
cconroy: Even if it's not common, you should be able to pass a map to the normal `add`

Perhaps we could refer to it as `addKwargs` instead?

README.md
cconroy: Got it: fixing

src/test/java/com/squareup/javapoet/CodeBlockTest.java
cconroy: I don't follow: it's fine in this case... Anyways, sounds like that would be a separate issue.

cconroy:
Just want to throw this out there: would `$T:clazz` be preferable to the current implementation of `$clazz:T` ?

src/main/java/com/squareup/javapoet/CodeBlock.java
cconroy: done

src/main/java/com/squareup/javapoet/CodeBlock.java
cconroy: fixed

src/main/java/com/squareup/javapoet/CodeBlock.java
cconroy: fixed

README.md
swankjesse: I think this is fine. Just griping.

swankjesse:
I think the scheme you’ve got is perfect. Maybe only request is to enforce all names start with a lowercase character so the syntax is completely unambiguous.

cconroy:
We can certainly enforce lowercase if you want, but I think the `:` is enough to disambiguate

swankjesse:
My anxiety is that today `$That:T` could be interpreted as `$T` followed by `hat:T`, which is valid syntax today.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -112,9 +119,71 @@ public Builder toBuilder() {
...
+     *
+     * <p> For example, to refer to the type {@link java.lang.Integer} with the argument
+     * name 'clazz' use a format string containing {@code $clazz:T} and include the key 'clazz' with
+     * value {@code java.lang.Integer.class} in the argument map.
+     */
+    public Builder addNamed(String format, Map<String, ?> arguments) {
+      int p = 0;
+
+      for (String argument : arguments.keySet()) {
+        checkArgument(LOWERCASE.matcher(argument).matches(),
```
swankjesse: this looks like it’s checking that the entire thing is lowercase

cconroy: fixed

swankjesse:
Squashed and merged. Thanks @cconroy !

PR #522, opened 2016-11-09
Always write UTF-8.

swankjesse:
Sometimes the system default charset is not UTF-8 and this is sad.
We always want UTF-8, and we want to encourage anyone doing tools
to use UTF-8.

Closes: https://github.com/square/javapoet/pull/474

### DIFF
--- a/src/main/java/com/squareup/javapoet/Util.java
+++ b/src/main/java/com/squareup/javapoet/Util.java
@@ -15,8 +15,6 @@
  */
 package com.squareup.javapoet;
 
-import static java.lang.Character.isISOControl;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/Util.java
```diff
@@ -28,11 +27,15 @@
...
 import javax.lang.model.element.Modifier;
 
+import static java.lang.Character.isISOControl;
+
 /**
  * Like Guava, but worse and standalone. This makes it easier to mix JavaPoet with libraries that
  * bring their own version of Guava.
  */
 final class Util {
+  static final Charset UTF_8 = Charset.forName("UTF-8");
```
JakeWharton: Can be replaced with `StandardCharsets.UTF_8`

swankjesse: Fixed!

Issue #105, opened 2014-12-05
Sensible wrapping, where appropriate


JakeWharton:
There are parts of code to which we can apply sensible wrapping. Right column break to be configured on `JavaWriter`.

Candidates include:
- Class declaration (extends, interfaces)
- Field initializer (if length exceeds wrap&indent initializer)
- Method declaration (params, exceptions)

roykachouh:
This was exactly what I was looking for. Any updates on when to expect this. I'm using JavaPoet (which i love!) to automatically generate server stubs and clients for my project based on specifications. 

There are times when things look a little messy, think builder pattern with a constructor with 10 arguments. Ideally, I was hoping to find something that can take checkstyle and automatically reformat based on our coding conventions, but it doesn't seem there is an appropriate solution.

Is this enhancement still on your milestone list?

roykachouh:
I can also try to take a shot at this with a pr.

swankjesse:
Really difficult to do well. Some teams have recommended coupling JavaPoet with another code formatter.

roykachouh:
I agree. That would be ideal as there can potentially be so many different code style preferences. Do you happen to know of any promising auto format tools out there? Googling has come up short for me so far.

tbroyer:
@roykachouh Google has https://github.com/google/google-java-format specific to their style.

sormuras:
They released version [1.0 Alpha](https://github.com/google/google-java-format/releases) with an all-inclusive jar. Just include it in your project and add a last line like:
```java
String formattedSource = new Formatter().formatSource(javapoetJavaFile.toString());
```

PR #525, opened 2016-11-09
Implement line wrapping.

swankjesse:
Currently only method parameters are wrapped. Anything else can be wrapped
by replacing a space with $W.

https://github.com/square/javapoet/issues/105

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -45,7 +45,7 @@ final class CodeWriter {
   private static final String NO_PACKAGE = new String();
 
   private final String indent;
-  private final Appendable out;
+  private final LineWrapper out;
   private int indentLevel;
 
   private boolean javadoc = false;
@@ -76,7 +76,7 @@ final class CodeWriter {
 
   CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes,
       Set<String> staticImports) {
-    this.out = checkNotNull(out, "out == null");
+    this.out = new LineWrapper(out, indent, 100);
     this.indent = checkNotNull(indent, "indent == null");
     this.importedTypes = checkNotNull(importedTypes, "importedTypes == null");
     this.staticImports = checkNotNull(staticImports, "staticImports == null");
@@ -278,6 +278,10 @@ final class CodeWriter {
           statementLine = -1;
           break;
 
+        case "$W":
+          out.wrappingSpace(indentLevel + 2);
+          break;
+
         default:
           // handle deferred type
           if (deferredTypeName != null) {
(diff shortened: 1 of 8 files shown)


swankjesse:
Sadly I think I need one more symbol, like `$Z` for a zero-width wrapping space. You’d use that anywhere you’d want to hint that it’s a good wrapping point without necessarily adding a space otherwise.

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -2210,29 +2211,53 @@ private String toString(TypeSpec typeSpec) {
...
     } catch (UnsupportedOperationException e) {
     }
   }
 
+  @Test public void lineWrapping() {
+    MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder("call");
+    methodBuilder.addCode("$[call(");
+    for (int i = 0; i < 32; i++) {
+      methodBuilder.addParameter(String.class, "s" + i);
+      methodBuilder.addCode(i > 0 ? ",$W$S" : "$S", i);
```
swankjesse: we probably want a helper that takes a `List` and gives you “comma separated” things

swankjesse: ... or we could just make it so that all commas do this? That’s insane right?

ronshapiro: yes plz (to the first comment): https://github.com/google/dagger/blob/9adf6be2c59d671876f9bf1ab31bc0a3261d78da/compiler/src/main/java/dagger/internal/codegen/CodeBlocks.java#L41

src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.javapoet;
+
+import java.io.IOException;
+
+import static com.squareup.javapoet.Util.checkNotNull;
+
+final class LineWrapper {
```
cconroy: top level javadoc?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -76,7 +76,7 @@
 
   CodeWriter(Appendable out, String indent, Map<String, ClassName> importedTypes,
       Set<String> staticImports) {
-    this.out = checkNotNull(out, "out == null");
+    this.out = new LineWrapper(out, indent, 100);
```
cconroy: make the width configurable?

cconroy:
just a couple of nits. otherwise LGTM

src/main/java/com/squareup/javapoet/LineWrapper.java
swankjesse: Good call. Done.

src/main/java/com/squareup/javapoet/CodeWriter.java
swankjesse: I’ll wait for somebody to request it. Opinionated software FTW.

PR #526, opened 2016-11-09
Support anonymous inner classes in ClassName#get

kroepke:
ClassName#get incorrectly used Class#getSimpleName for anonymous inner classes, which is an empty string.
With this change it falls back to using Class#getName for anonymous classes and taking the compiler generated name for the class.

`mvn clean verify` passed, signed the CLA.

### DIFF
--- a/src/test/java/com/squareup/javapoet/ClassNameTest.java
+++ b/src/test/java/com/squareup/javapoet/ClassNameTest.java
@@ -16,16 +16,13 @@
 package com.squareup.javapoet;
 
 import com.google.testing.compile.CompilationRule;
-
+import java.util.Map;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.util.Elements;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-
-import java.util.Map;
-
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.util.Elements;
 
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


kroepke:
I'm not overly happy with how the class' canonical name is assembled again, mainly because `$` is a valid identifier character and might accidentally follow a `.` in the name.
However, any other solution blows up the code required to keep track of the anonymous class name parts, so I chose to keep it simple.

swankjesse:
Just curious – what’s your use case for this?

kroepke:
I am generating code for a rule language, in which I need to refer to potentially anonymous inner classes.
Those classes are basically user defined functions, that I need a named reference to, to later directly call them (a little like in this gist: https://gist.github.com/kroepke/33b6c7131ebdde419adca83322fb133d#file-rule-1-java-L89)
This came up in a test case where those UDFs were defined inline (which is typically not the case but also not forbidden).

src/test/java/com/squareup/javapoet/ClassNameTest.java
```diff
@@ -107,6 +107,12 @@ private void assertBestGuessThrows(String s) {
         .isEqualTo("java.lang.Object");
     assertThat(ClassName.get(OuterClass.InnerClass.class).toString())
         .isEqualTo("com.squareup.javapoet.ClassNameTest.OuterClass.InnerClass");
+    try {
+      assertThat((ClassName.get(new Object() {}.getClass())).toString())
+          .isEqualTo("com.squareup.javapoet.ClassNameTest$1");
+    } catch (IllegalArgumentException e) {
+      fail();
```
JakeWharton: Just remove the try/catch. An exception will fail the test.

kroepke:
@JakeWharton Do you need anything more from me, just so I don't lose the context?
Thanks!

JakeWharton:
I defer to @swankjesse to keep this library in his head. I'm fine with the change though.

PR #527, opened 2016-11-10
Fix imports of nested types.

swankjesse:
This adds a test and rolls back the changes made in a136eaa777b7750e16b3c6e8bd6d0228442e3e1d.
The tests contributed with that PR still pass.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.Callable;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -629,18 +630,41 @@ private boolean isJava8() {
...
-            + "import java.lang.Comparable;\n"
-            + "\n"
-            + "class Taco extends org.fish.taco.Taco "
-            + "implements Comparable<Taco>, com.taco.bell.Taco {\n"
-            + "}\n");
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "import java.util.concurrent.Callable;\n"
+        + "\n"
+        + "class Outer extends Callable<Outer.Inner> {\n"
```
swankjesse: This fixes a regression. Without this PR this incorrectly emits `extends Callable<Inner>` which doesn’t work because `Inner` isn’t in scope yet.

PR #528, opened 2016-11-10
Rollback support for retaining parameter annotations.

swankjesse:
We recently merged this because it was demanded, but looking through our
history we've had vocal users complaining about this exact kind of
behavior.

https://github.com/square/javapoet/pull/487/files


gk5885:
This almost certainly saves me from having to spend a bunch of time working around the broken edge cases the version release.  Blindly copying annotations _seems_ good, but has actually broken our users.  Thanks for catching this.

swankjesse:
Yep. JavaPoet is best when it’s dumb. Even simple-seeming things like `overriding()` implies tons of policy that we can’t get right.

PR #529, opened 2016-11-10
Only send the relevant characters into the matcher

cconroy:


gmalmquist:
lgtm

swankjesse:
Test case?

tbroyer:
Can't you simply use `Matcher#lookingAt` instead of `Matcher#matches`? (with the same trimmed-down regex as in this change, but without the need for the `subSequence`)

(and then use `end()` or `regionEnd()` btw instead of `regionStart() + argumentName.length() + 3)

cconroy:
@tbroyer good call. i'll update accordingly.

cconroy:
@swankjesse this should be covered by the existing tests.

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -151,15 +151,21 @@ public Builder addNamed(String format, Map<String, ?> arguments) {
...
           p = nextP;
         }
-        Matcher matcher = NAMED_ARGUMENT.matcher(format.subSequence(p, format.length()));
-        if (matcher.matches()) {
+
+        Matcher matcher = null;
+        int colon = format.indexOf(':', p);
+        if (colon != -1) {
+          int endIndex = Math.min(colon + 2, format.length());
+          matcher = NAMED_ARGUMENT.matcher(format.substring(p, endIndex));
```
cconroy: I also realize that `subSequence` was not implemented how I had originally hoped (as an iterator) and instead it's just a straight `substring` call... So, we do want to retain the colon lookup to avoid creating a substring with the remainder of the code block each time.

PR #537, opened 2017-01-18
There is a missing symbol "?" in [$T for Types] - [Import static]

mcxtzhang:
There is a missing symbol "?" in [$T for Types] - [Import static]

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ MethodSpec beyond = MethodSpec.methodBuilder("beyond")
     .addStatement("result.add($T.createNimbus(\"2001\"))", hoverboard)
     .addStatement("result.add($T.createNimbus($T.THUNDERBOLT))", hoverboard, namedBoards)
     .addStatement("$T.sort(result)", Collections.class)
-    .addStatement("return result.isEmpty() $T.emptyList() : result", Collections.class)
+    .addStatement("return result.isEmpty() ? $T.emptyList() : result", Collections.class)
     .build();
 
 TypeSpec hello = TypeSpec.classBuilder("HelloWorld")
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -320,7 +320,7 @@ MethodSpec beyond = MethodSpec.methodBuilder("beyond")
     .addStatement("result.add($T.createNimbus(\"2001\"))", hoverboard)
     .addStatement("result.add($T.createNimbus($T.THUNDERBOLT))", hoverboard, namedBoards)
     .addStatement("$T.sort(result)", Collections.class)
-    .addStatement("return result.isEmpty() $T.emptyList() : result", Collections.class)
+    .addStatement("return result.isEmpty() ? $T.emptyList() : result", Collections.class)
```
tbroyer: More importantly, I think this `isEmpty` check is useless (technically, in the generated code it definitely is; but it doesn't even add anything to the example either). I suggest replacing the line with:
```java
    .addStatement("return result")
```

mcxtzhang: This is an example for using `Import static`.
I think `$T.emptyList()` is more importantly.

tbroyer: There's already the `$T.sort(result)` above.

sormuras: Now I tend to agree with @mcxtzhang ... if only `Collections.sort` was to be imported, it'd be better to use a single static type import instead of the `*` version.
The additional value is to demonstrate the usage of `.addStaticImport(Collections.class, "*")` with a `class` instance.

sormuras: Now, only an `Enum` constant [overload](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/JavaFile.java#L234) demo is missing... 😉

sormuras:
@mcxtzhang can please insert two more `?`s into [JavaFileTest](https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/JavaFileTest.java)? This is where I copied the wrong documentation text section from. Line [44](https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/JavaFileTest.java#L44) and [72](https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/JavaFileTest.java#L72).

mcxtzhang:
I've inserted them.

sormuras:
Merci!

PR #544, opened 2017-02-01
Improve test on TypeName

danglotb:
Hello,

I open this pull request to specify the line [197](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/TypeName.java#L197) in the ```equals()``` method of ```com.squareup.javapoet.TypeName```.

```java
if (o == null) return false;
```

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeNameTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeNameTest.java
@@ -25,6 +25,7 @@ import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 
 public class TypeNameTest {
@@ -178,5 +179,6 @@ public class TypeNameTest {
     assertEquals(a.toString(), b.toString());
     assertThat(a.equals(b)).isTrue();
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
+    assertFalse(a.equals(null));
   }
 }


src/test/java/com/squareup/javapoet/TypeNameTest.java
```diff
@@ -178,5 +179,6 @@ private void assertEqualsHashCodeAndToString(TypeName a, TypeName b) {
     assertEquals(a.toString(), b.toString());
     assertThat(a.equals(b)).isTrue();
     assertThat(a.hashCode()).isEqualTo(b.hashCode());
+    assertFalse(a.equals(null));
```
ronshapiro: `assertThat(a).isNotNull();`

tbroyer: He, no: this specifically about testing the `.equals(null)` contract 😉

sormuras: See https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/TypeName.java#L197

ronshapiro: Oh fair. What about `assertThat(a).isNotEqualTo(null)'?

sormuras: Would be more inline with
```java
```
I think `assertThat(a.equals(a)).isTrue();` is missing, too. Testing the [if (this == o) return true;](https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/TypeName.java#L196) contract.

JakeWharton: True, but that's just an optimization. If `a.equals(b)` works then `b.equals(a)` must also work, thus we then know `a.equals(a)` will work.

sormuras: True. True. Stays true. 🥂

PR #547, opened 2017-02-14
Add error-prone compiler.

JakeWharton:


pom.xml
```diff
@@ -94,11 +94,25 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.0</version>
+        <version>3.3</version>
```
tbroyer: Why not use 3.6?

pom.xml
```diff
@@ -94,11 +94,25 @@
...
+          <compilerId>javac-with-errorprone</compilerId>
+          <forceJavacCompilerUse>true</forceJavacCompilerUse>
           <source>${java.version}</source>
           <target>${java.version}</target>
         </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-compiler-javac-errorprone</artifactId>
+            <version>2.8</version>
```
tbroyer: FYI, there's a 2.8.1.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: java
 
 jdk:
-  - oraclejdk7
```
tbroyer: One possibility in case you still want to test on JDK7 would be to use a Maven profile for errorprone that's only activated when JDK==8 (errorprone won't work with JDK 9 either)

pom.xml
JakeWharton: I just copied whatever was on the error-prone instructions. I can update it, but i've already merged this into about 10 projects with these same versions... 😶

tbroyer: FWIW, if you use annotation processing in those other projects, then you'll have to `mvn clean` before you can `mvn compile` again. This is a regression introduced in 3.2 that was fixed in 3.5.1. Look for “As an anecdote” in https://blog.ltgt.net/maven-is-broken-by-design-2/

.travis.yml
JakeWharton: Couldn't we also use `-version 7` since error-prone is based on javac 9?

tbroyer: I'm almost certain the version of JavaC that error-prone uses does not have that feature (just like it doesn't have modules/jigsaw) unfortunately.

PR #551, opened 2017-03-21
Verify that null is not passed as vararg parameter

ashdavies:
Encountered unexpected behaviour when printing a class as the result of a null being passed from the result of a method call into `TypeSpec` modifiers


swankjesse:
Thanks!

Are we making the same mistake elsewhere?
One option to save code would be to call this instead of a loop:

```
modifiers.contains(null)
```

sormuras:
Hm, introduce `Util.checkNoEntryIsNull(Object...)` (or better name) and insert that check in every vararg API method?

ashdavies:
@swankjesse np, I had thought about using `checkState(!Arrays.asList(...).contains(null))` instead, but I wasn't so sure about having to double the iteration, but having a `check...` method as @sormuras mentioned would seem more consistent with the code style

PR #559, opened 2017-05-12
Fix conflicts with types with same name as type being declared 

tbroyer:


JakeWharton:
If you mark it as `@Ignored` we can merge it in for now...

tbroyer:
Added more tests (it turned out that annotations and type variable bounds had issues too; actually, anything used while declaring the type, before declaring its members), and a proposed fix.

I put the fix in a separate commit, to (maybe) make it easier to review;  but this means that the first commits introduces failing tests (broken build). Tell me how you'd prefer the commits to be organized and I'll make the changes (or you could squash-merge the PR and chose/write your own commit message)

tbroyer:
Ouch, that checkstyle rule is annoying; any suggestion how to refactor the code? (that last `else` body could be extracted in a method, any suggestion for the method name?)

NightlyNexus:
Is there value in using Checkstyle for method line count? That is implementation-altering, not style-altering.

JakeWharton:
It's a sanity check. Feel free to bump it up.

tbroyer:
> Feel free to bump it up.

Bumped to 160 lines.

Also actually fixed the bug (hadn't run the whole testsuite yesterday and it broke with enums), reducing allocations as a side-effect (no TypeSpec.Builder, and no lists and sets when building the builder).

swankjesse:
Nice fix.

jessevdam:
Thank you very much for fixing this one :+1:

tbroyer:
1. Be bitten by bug on Friday
 2. Debug and send PR in the afternoon
 3. Polish during the weekend
 4. Profit from new release when back to work on Monday, so you can continue your project where you left it with almost no interruption 🎉

PR #560, opened 2017-05-12
Add convenience anonymousClassBuilder(CodeBlock)

### LINKED ISSUE #558: Proposal: Add a convenience `anonymousClassBuilder` method to TypeSpec that accepts a CodeBlock
In some cases - I already have a formatted `CodeBlock` (either borrowed from another `TypeSpec` with the same signature or otherwise). If this sounds good, I can make a PR

hzsweers:
Resolves #558

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -112,8 +112,8 @@ public final class TypeSpec {
 
   public static Builder anonymousClassBuilder(String typeArgumentsFormat, Object... args) {
     return anonymousClassBuilder(CodeBlock.builder()
-            .add(typeArgumentsFormat, args)
-            .build());
+        .add(typeArgumentsFormat, args)
+        .build());
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -111,9 +111,13 @@ public static Builder enumBuilder(ClassName className) {
   }
 
   public static Builder anonymousClassBuilder(String typeArgumentsFormat, Object... args) {
-    return new Builder(Kind.CLASS, null, CodeBlock.builder()
-        .add(typeArgumentsFormat, args)
-        .build());
+    return anonymousClassBuilder(CodeBlock.builder()
+            .add(typeArgumentsFormat, args)
+            .build());
```
JakeWharton: Wrong indent

hzsweers: derp, will fix

hzsweers: 9da74b5

hzsweers:
Thanks!

PR #562, opened 2017-05-23
Fix ClassName regression for class names starting with a $

PhilGlass:
Fixes a regression introduced in #526 where classes whose names start with a `$`would incorrectly have `.` separators removed - for example, a class `$Bar` in package `foo` would be named `foo$Bar` rather than the expected `foo.$Bar`.

This is only really a problem when generated code references other generated code, but I imagine that's fairly common for consumers of this library.


swankjesse:
Thanks!

PR #577, opened 2017-08-25
Unify behavior of writeTo() on file encoding.

lowZoom:
Let **com.squareup.javapoet.JavaFile#writeTo(javax.annotation.processing.Filer)** behave like other writeTo() methods which save files in UTF-8 encoding by default instead of system default encoding.  

That is more intuitive and more useful for common use.

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -105,7 +105,7 @@ public final class JavaFile {
     writeTo(directory.toPath());
   }
 
-  /** Writes this to {@code filer}. */
+  /** Writes this to {@code filer} as UTF-8. */
   public void writeTo(Filer filer) throws IOException {
     String fileName = packageName.isEmpty()
         ? typeSpec.name
@@ -113,7 +113,7 @@ public final class JavaFile {
     List<Element> originatingElements = typeSpec.originatingElements;
     JavaFileObject filerSourceFile = filer.createSourceFile(fileName,
         originatingElements.toArray(new Element[originatingElements.size()]));
-    try (Writer writer = filerSourceFile.openWriter()) {
+    try (Writer writer = new OutputStreamWriter(filerSourceFile.openOutputStream(), UTF_8)) {
       writeTo(writer);
     } catch (Exception e) {
       try {


tbroyer:
The difference is that `Filer` can give you a `Writer` **with the appropriate encoding** such that the generated code will compile (we're in an annotation processor where generated code is compiled right away). Switching to UTF-8 has the potential to break people without any good reason.

lowZoom:
I'm sorry that we can't agree on the fact that UTF-8 is a more appropriate encoding than the system default ones. Then why do the other `writeTo()` methods use UTF-8 ? I'm really confused.

tbroyer:
If you use `Filer` it means you're in an annotation processor, which means the generated classes will be immediately compiled, possibly without even be written to disk. You absolutely need to respect the encoding for the current compilation or it'll break. Forcing UTF-8 here means forcing users to use UTF-8 for their compilation, i.e. for their own sources too. While this is probably a good thing, that's a big breaking change; and an arbitrary change backed by no compelling argument.

The other write methods in the other hand initially used the default encoding, but that caused issues to some people **who specifically couldn't change the default encoding**, and expected UTF-8 output (because the output is processed/compiled by other tasks/tools), so it was changed to UTF-8 (the initial request was to be able to provide an encoding). The other methods also will write to disk, so if you don't want UTF-8, you can post-process the files to change their encoding. Those other methods have to somehow chose an encoding, whether the default one or another one, and that's an arbitrary choice anyway. The `Filer` on the other hand knows the **appropriate** encoding for the ongoing task, so you **don't want** to arbitrary choose one.

lowZoom:
I'm indeed generating code in an annotation processor. My project uses UTF-8 for sources (forcing by Maven configuration). I want to generate code containing non-Ascii characters (Chinese exactly).

The code generated by the current `writeTo(Filer)` break the compilation because it uses the system encoding (which is GBK). 

So it's not true that 
> The Filer on the other hand knows the appropriate encoding for the ongoing task.

---

But this PR will actually break some existing code. If avoiding it is the first priority, maybe not changing this method is a better option. However, my situation is able to be worked around ~~in some dirty way~~ somehow.

tbroyer:
@lowZoom Attached is a simple Gradle project with an annotation processor generating a file containing some Chinese ([test.zip](https://github.com/square/javapoet/files/1256519/test.zip)). You can tweak the project's encoding (`-encoding` flag passed to JavaC) in the `build.gradle` file (see comment there), and the JVM settings by setting the `GRADLE_OPTS` environment variable (e.g. `GRADLE_OPTS="-Dfile.encoding=ISO-8859-1"`). The file generated by the annotation processor is written to disk in `build/generated/sources/apt/main/test/generated/Foo.java`.
Clean and build the project with `./gradlew clean build` (or `GRADLE_OPTS="-Dfile.encoding=ISO-8859-1" ./gradlew clean build`) and inspect the generated file.
I tried several combinations, and all had the expected behavior, i.e. respecting the specified encoding (or using the default encoding if the line in the `build.gradle` file was commented out).
You can of course go further and try to change the `src/main/java/test/Test.java` file to use non-ASCII chars and try several encodings, and see how the resulting class behaves at runtime.

BTW, unless you control the environment it will run in, or can somehow require the use of a given source encoding, an annotation processor should only ever output US-ASCII (lowest common denominator), and use escape sequences for non-ASCII chars in string or character literals.

Maybe there's a bug in the maven-compiler-plugin; or maybe your JDK behaves differently than OpenJDK (tested with OpenJDK 8u131).

sormuras:
Maybe you are hit by an Eclipse compiler bug? https://bugs.eclipse.org/bugs/show_bug.cgi?id=483690

tbroyer:
@lowZoom [Here](https://github.com/square/javapoet/files/1256585/test.zip)'s the same project, adapted to Maven (I also kept the Gradle-based build, moving things down into an `app/` subproject).
`mvn clean package -Dproject.build.sourceEncoding=xxx` changes the encoding of `app/target/generated-sources/annotations/test/generated/Foo.java`, as expected (using Maven 3.5.0 and latest maven-compiler-plugin).

lowZoom:
Well, it seems like it's an Intellij problem:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000069750-UTF-8-Annotation-Processor-Compiler-source-encoding-on-Windows

@tbroyer @sormuras Thank you all for the explanation and the information. It's very helpful.

PR #579, opened 2017-08-31
License and inline annotation members

davti:
I have modified the code to add the following License header for JavaFile and inline AnnotationSpec members at runtime.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -27,7 +27,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.ExecutableElement;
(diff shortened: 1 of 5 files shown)


JakeWharton:
This PR is not really reviewable because you changed the code style. Please use https://github.com/square/java-code-styles.

JakeWharton:
Is the file comment feature not adequate for this? It will emit a series of single-line comments instead of a multi-line one. Is that a problem?

davti:
The comments block id not in the standard java file License format so, no the  file comment feature not adequate for this.

JakeWharton:
It's a bit annoying to need to support both. I wonder if we can just change the file comment to be multi-line if there's more than one line in the supplied string.

davti:
"This PR is not really reviewable because you changed the code style" I use Netbeans not IntelliJ can you please provide the coding style config?

I have fixed the formatting manually please re review

JakeWharton:
Ah, sorry. We only have available styles for IntelliJ (and thus Android Studio). I'd love to figure out a way to create them for Eclipse and NetBeans too, but it's not very highly prioritized.

JakeWharton:
@swankjesse @ronshapiro any thoughts about https://github.com/square/javapoet/pull/579#issuecomment-326447812 ?

davti:
apologies missed the formatting of the test files, can you send me the InteliJ formatter file I can have a look and replicate for Netbeans, Thanks,  Tim

davti:
PS: Note the License header is slightly different from standard javadoc the comment field starts with /* as opposed to /** for javadoc

JakeWharton:
Yep. We'd do /* for a traditional multi-line comment, not a Javadoc comment.

davti:
I think that a multi line comment block would be a good edition as well, with the License being a specialisation of that block...

davti:
I have had a look at the failing unit tests, I can fix the unit test string results to pass, the fixes for the most part can be achieved by adding the inlineMembers(true) to the unit test but i'm not sure if the is what the tests are to achieve...

swankjesse:
Sorry I'm late to the review on this. I think this is completely unnecessary. Nobody should care what type of comment you use to comment your license text. This is a made up use case.

PR #594, opened 2017-12-27
Fix search of annotated type parameters in imports

zhurs:
Right now generated code is 
`
package com.squareup.tacos;

import com.squareup.meat.Chorizo;
import java.util.List;

class Taco {
  List<@Spicy com.squareup.meat.Chorizo> chorizo;
}
`
and javac can't compile it (see also https://github.com/square/javapoet/issues/431)

Patch does not fix root of problem, but works for common cases.

### DIFF
--- a/src/test/java/com/squareup/javapoet/JavaFileTest.java
+++ b/src/test/java/com/squareup/javapoet/JavaFileTest.java
@@ -17,6 +17,7 @@ package com.squareup.javapoet;
 
 import java.util.Collections;
 import java.util.Date;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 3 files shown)


src/test/java/com/squareup/javapoet/JavaFileTest.java
```diff
@@ -290,6 +291,29 @@ private TypeSpec importStaticTypeSpec(String name) {
         + "}\n");
   }
 
+  @Test public void annotatedTypeParam() throws Exception {
+    String source = JavaFile.builder("com.squareup.tacos",
```
swankjesse: nit: we wrap +4 in JavaPoet

swankjesse:
Thanks, this is a nice fix. Merged manually.

PR #596, opened 2017-12-30
Add isEmpty to CodeBlock.Builder.

### LINKED ISSUE #595: CodeBlock.Builder.isEmpty()
This is useful when building things that are concatenated to check when you need to prepend a `+`, for example.

JakeWharton:
Closes #595.


src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -20,6 +20,8 @@
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertFalse;
```
Egorand: nit: can stick to Truth to avoid extra imports

Issue #588, opened 2017-11-27
Zero-width version of $W


ronshapiro:
I'm looking into the possibility of removing google-java-format from Dagger (https://github.com/google/dagger/issues/368) and `$W` is a useful tool in adding line breaks when necessary, but sometimes I want the option of breaking if necessary, or no space at all. 

Consider this example: I want to write the following method invocations without thinking about how many arguments I'm about to pass:

```java
Foo_Factory.create(arg0, arg1, arg2);

Foo_Factory.create(
    arg0,
    arg1,
    arg2,
    arg3,
    arg4,
    arg5,
    ...,
    argN);
```

`$W` works for args `1..N` but in the first example, adding in a `$W` between `create(` and `arg0` would create an awkward space.

https://github.com/square/kotlinpoet/pull/281 is a similar attempt that maybe can be applied here as well?

JakeWharton:
This is also needed for parameter lists. The first parameter never wraps.

ronshapiro:
What are your thoughts on having `$W` inspect the previous character for insights on whether it should be zero-width or one-width?

In `foo($Warg1,$Warg2)`

swankjesse:
I think that might be too clever. Maybe just like `$B` for optional-break?

ronshapiro:
Fine by me. Cool if I send a PR?

swankjesse:
Please do!

PR #598, opened 2018-01-12
Adds `$B` as an optional newline if a line will exceed 100 chars

ronshapiro:
Closes https://github.com/square/javapoet/issues/588

I left a few comments in the code with questions that I had.

As I was implementing it, I was thinking that `$Z` may make more sense (`$W` is for _whitespace, `$Z` is for zero-whitespace?). But it's late, so maybe I'm thinking sideways.

I didn't love the `FlushType`/`nextFlush` names. Open to other suggestions (again, it was late, so `FlushType` made tons of sense to me)

### DIFF
--- a/src/main/java/com/squareup/javapoet/LineWrapper.java
+++ b/src/main/java/com/squareup/javapoet/LineWrapper.java
@@ -90,7 +90,7 @@ final class LineWrapper {
   }
 
   /** Emit a newline character if the line will exceed it's limit, otherwise do nothing. */
-  void wrappingZeroWidth(int indentLevel) throws IOException {
+  void zeroWidthSpace(int indentLevel) throws IOException {
     if (closed) throw new IllegalStateException("closed");
 
     // DO NOT SUBMIT: multiple zero-width chars in a row - should that cause a flush?
--- a/src/main/java/com/squareup/javapoet/LineWrapper.java
+++ b/src/main/java/com/squareup/javapoet/LineWrapper.java
@@ -35,8 +35,15 @@ final class LineWrapper {
   /** The number of characters since the most recent newline. Includes both out and the buffer. */
   private int column = 0;
 
-  /** -1 if we have no buffering; otherwise the number of spaces to write after wrapping. */
+  /**
+   * -1 if we have no buffering; otherwise the number of {@code indent}s to write after wrapping.
+   */
   private int indentLevel = -1;
+
+  /** {@code null} if we have no buffering; otherwise the type to pass to the next call to {@link
+   * #flush(FlushType)}.
+   */
+  private FlushType nextFlush;
 
   LineWrapper(Appendable out, String indent, int columnLimit) {
     checkNotNull(out, "out == null");
@@ -49,7 +56,7 @@ final class LineWrapper {
   void append(String s) throws IOException {
     if (closed) throw new IllegalStateException("closed");
 
-    if (indentLevel != -1) {
+    if (nextFlush != null) {
       int nextNewline = s.indexOf('\n');
 
       // If s doesn't cause the current line to cross the limit, buffer it and return. We'll decide
@@ -62,7 +69,7 @@ final class LineWrapper {
 
       // Wrap if appending s would overflow the current line.
       boolean wrap = nextNewline == -1 || column + nextNewline > columnLimit;
-      flush(wrap);
+      flush(wrap ? FlushType.WRAP : nextFlush);
     }
 
     out.append(s);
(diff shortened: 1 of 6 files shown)


src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -49,7 +56,7 @@
   void append(String s) throws IOException {
     if (closed) throw new IllegalStateException("closed");
 
-    if (indentLevel != -1) {
+    if (nextFlush != null) {
```
ronshapiro: `nextFlush != null` was way more clear to me than `indentLevel != -1`. I can't tell you how long I spent staring at the `indentLevel != -1` looking confused at what it was doing.

tbroyer:
Fwiw, I think I like `$Z` better: https://en.wikipedia.org/wiki/Zero-width_space

JakeWharton:
Once this is finalized and merged, we should modify the class header, parameter lists, initializers to use this.

src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -35,9 +35,16 @@
   /** The number of characters since the most recent newline. Includes both out and the buffer. */
   private int column = 0;
 
-  /** -1 if we have no buffering; otherwise the number of spaces to write after wrapping. */
+  /**
+   * -1 if we have no buffering; otherwise the number of {@code indent}s to write after wrapping.
```
swankjesse: Thanks.

src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -35,9 +35,16 @@
...
+   */
   private int indentLevel = -1;
 
+  /** {@code null} if we have no buffering; otherwise the type to pass to the next call to {@link
```
swankjesse: supernit: we like `/**`  on its own line

swankjesse: Also nit, no need to use `{@code }` for null; it’s a word

src/test/java/com/squareup/javapoet/LineWrapperTest.java
```diff
@@ -96,6 +128,19 @@
...
   }
 
+  @org.junit.Ignore
+  @Test public void overlyLongLinesWithLeadingZeroWidth() throws Exception {
+    StringBuffer out = new StringBuffer();
+    LineWrapper lineWrapper = new LineWrapper(out, "  ", 10);
+    lineWrapper.zeroWidthSpace(2);
+    lineWrapper.append("abcdefghijkl");
+    lineWrapper.close();
+    // DO NOT SUBMIT: what should the functionality here be? if we don't want a newline, we can check:
```
swankjesse: Yeah, I think wrapping at column 0 is a mistake. Good catch.

src/test/java/com/squareup/javapoet/LineWrapperTest.java
```diff
@@ -118,6 +163,28 @@
     assertThat(out.toString()).isEqualTo("abcde\n    fghij\nklmnopqrstuvwxy");
   }
 
+  @Test public void noWrapEmbeddedNewlines_ZeroWidth() throws Exception {
```
swankjesse: nice test

src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -76,31 +83,56 @@ void append(String s) throws IOException {
   void wrappingSpace(int indentLevel) throws IOException {
     if (closed) throw new IllegalStateException("closed");
 
-    if (this.indentLevel != -1) flush(false);
-    this.column++;
+    if (this.nextFlush != null) flush(nextFlush);
+    column++; // increment the column even though the space is deferred to next call to flush()
```
swankjesse: supernit: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

src/main/java/com/squareup/javapoet/LineWrapper.java
```diff
@@ -76,31 +83,56 @@ void append(String s) throws IOException {
...
+    this.nextFlush = FlushType.SPACE;
+    this.indentLevel = indentLevel;
+  }
+
+  /** Emit a newline character if the line will exceed it's limit, otherwise do nothing. */
+  void zeroWidthSpace(int indentLevel) throws IOException {
+    if (closed) throw new IllegalStateException("closed");
+
+    // DO NOT SUBMIT: multiple zero-width chars in a row - should that cause a flush?
```
swankjesse: I don’t think there’s a behavior consequence, is there?

swankjesse:
Logic looks great. I have some small style nitpicks. If you’d like to send another PR to address those; please do. Otherwise I’ll comb ’em out next time I’m through these files.

Sorry for the slow turnaround.

PR #599, opened 2018-01-12
Implement CodeBlock.join() and CodeBlock.joining()

ronshapiro:
Closes https://github.com/square/javapoet/issues/595

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <java.version>1.7</java.version>
+    <java.version>1.8</java.version>
     <junit.version>4.12</junit.version>
     <truth.version>0.28</truth.version>
     <compile-testing.version>0.9</compile-testing.version>
--- a/src/main/java/com/squareup/javapoet/CodeBlock.java
+++ b/src/main/java/com/squareup/javapoet/CodeBlock.java
@@ -22,6 +22,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collector;
+import java.util.stream.StreamSupport;
 import javax.lang.model.element.Element;
 import javax.lang.model.type.TypeMirror;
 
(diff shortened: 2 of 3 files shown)


src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -102,6 +104,53 @@ public static CodeBlock of(String format, Object... args) {
...
+   * For example, joining {@code String s}, {@code Object o} and {@code int i} using {@code ", "}
+   * would produce {@code String s, Object o, int i}.
+   */
+  public static CodeBlock join(Iterable<CodeBlock> codeBlocks, String separator) {
+    Collection<CodeBlock> collection;
+    if (codeBlocks instanceof Collection) {
+      collection = (Collection<CodeBlock>) codeBlocks;
+    } else {
+      collection = new ArrayList<>();
+      codeBlocks.forEach(collection::add);
```
tbroyer: Use `StreamSupport.stream(codeBlocks.spliterator(), false)` to create a stream from the iterable instead of making a copy?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -102,6 +104,53 @@ public static CodeBlock of(String format, Object... args) {
...
+  }
+
+  /**
+   * A {@link Collector} implementation that joins {@link CodeBlock} instances together into one
+   * separated by {@code separator}. For example, joining {@code String s}, {@code Object o} and
+   * {@code int i} using {@code ", "} would produce {@code String s, Object o, int i}.
+   */
+  public static Collector<CodeBlock, ?, CodeBlock> joining(
+      String separator, String prefix, String suffix) {
+    Builder builder = builder().add("$N", prefix);
```
tbroyer: Why `$N` here? Should it rather take `CodeBlock`s instead?

src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -281,4 +283,42 @@
...
+
+  @Test public void joiningSingle() {
+    List<CodeBlock> codeBlocks = new ArrayList<>();
+    codeBlocks.add(CodeBlock.of("$S", "hello"));
+
+    CodeBlock joined = codeBlocks.stream().collect(CodeBlock.joining(" || "));
+    assertThat(joined.toString()).isEqualTo("\"hello\"");
+  }
+
+  @Test public void joiningWithSeparator() {
```
tbroyer: Did you mean `joiningWithPrefixAndSuffix`?

ronshapiro: d'oh. yes.

src/main/java/com/squareup/javapoet/CodeBlock.java
ronshapiro: I figured that passing a string (with a literal) is likely easier, and `$N` maintains the exact string.

I thought about having code blocks. I think that's probably less common (at that point, you might just want to add the joined code block to a `CodeBlock.Builder`?)

src/main/java/com/squareup/javapoet/CodeBlock.java
ronshapiro: Good call

ronshapiro: I'm not sure why, but Github isn't picking up my latest commit which added this...

JakeWharton: That is weird. It's definitely on the branch that created this PR...

ronshapiro: I squashed them all into one commit and it seems to be getting it all now.

ronshapiro:
This PR is getting all sorts of weird. The Travis logs seem to be clean, but Github is reporting an error

pom.xml
```diff
@@ -20,7 +20,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <java.version>1.7</java.version>
+    <java.version>1.8</java.version>
```
sormuras: Is the upgrade to Java 8 intended?

JakeWharton: Yes. Collectors are a Java 8 API.

ronshapiro: I imagine this is a safe change - with JDK 9 out and 10 on the way, most people should be using `javac8` at the very least by this point, right?

sormuras: iirc it was left on 1.7 for android compat reason ... some year(s) ago.

JakeWharton: Android tooling has required Java 8 since AGP 2.2 or 2.3. The biggest holdout for me was the IntelliJ platform but it's on Java 8 as of 2017.1.

sormuras: Perfect. This enables the migration to the JUnit Plaform! ;)

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -102,6 +104,46 @@ public static CodeBlock of(String format, Object... args) {
...
+   * Joins {@code codeBlocks} into a single {@link CodeBlock}, each separated by {@code separator}.
+   * For example, joining {@code String s}, {@code Object o} and {@code int i} using {@code ", "}
+   * would produce {@code String s, Object o, int i}.
+   */
+  public static CodeBlock join(Iterable<CodeBlock> codeBlocks, String separator) {
+    return StreamSupport.stream(codeBlocks.spliterator(), false).collect(joining(separator));
+  }
+
+  /**
+   * A {@link Collector} implementation that joins {@link CodeBlock} instances together into one
```
swankjesse: Great docs.

src/main/java/com/squareup/javapoet/CodeBlock.java
swankjesse: I like this. Avoids escaping weirdness too.

src/main/java/com/squareup/javapoet/CodeBlock.java
ronshapiro: You can thanks @gk5885 for those!

PR #603, opened 2018-01-21
JUnit 5

sormuras:
Proof-of-concept PR showing how-to upgrade to [JUnit 5](http://junit.org/junit5/). Due to the modular architecture of JUnit 5, JUnit Platform executes "legacy" JUnit 4 tests with the Vintage Engine. No code was changed here.

A single test, namely `UtilTest`, was migrated and now uses the new JUnit Jupiter API:

- [assertAll](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html#assertAll-java.lang.String-java.util.stream.Stream-)
- [Dynamic Tests](http://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,8 +22,8 @@
 
     <java.version>1.7</java.version>
     <junit.version>4.12</junit.version>
-    <truth.version>0.28</truth.version>
-    <compile-testing.version>0.9</compile-testing.version>
+    <truth.version>0.39</truth.version>
+    <compile-testing.version>0.15</compile-testing.version>
   </properties>
 
   <scm>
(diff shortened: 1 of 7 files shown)


sormuras:
The number of dependencies will shrink, when JUnit 4 + Vintage is not longer needed.

image


More flexibility, more UI structure, easier to re-run single (dynamic/parameterized) cases.

sormuras:
The _apparent added complexity_ is only due to my desire to have better reporting and structure. Like the screenshot demonstrates.

I could just as easily have upgraded to JUnit 5 without the use of `assertAll` or **dynamic tests**. Only some import statements needed some tweaks then.

swankjesse:
It just seems like I'm typing in a lot more code, using more complicated idioms, and gaining something I never asked for in return.

I don't need dynamic tests or assertAll(). I see the motivation but don't like the cost.

Frameworks like JUnit are best when they're simple. Here they've gone off and lambdafied everything and the result is complexity.

I'm also angry that JUnit 4 is degraded to “legacy”. You don't get to deprecated a framework that everyone uses because you got bored.

swankjesse:
Lets sit out of JUnit 5 for a while. If it has something we actually benefit from, we'll upgrade.

PR #611, opened 2018-02-06
Correct usage of toString() in readme

evonsdesigns:
Should have return type of `String`, not `void`.


JakeWharton:
`void toString()` methods are all the rage.

Issue #614, opened 2018-02-19
Type annotations on array types.


rluble:
Type annotations can be applied to the array type or any of its components. From JLS §[9.7.4](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.4):

> Type annotations can apply to an array type or any component type thereof (§10.1). For example, assuming that `A`, `B`, and `C` are annotation types meta-annotated with `@Target(ElementType.TYPE_USE)`, then given the field declaration:
>
> ```java
>@C int @A [] @B [] f;
>```
> `@A` applies to the array type `int[][]`, `@B` applies to its component type `int[]`, and `@C` applies to the element type int. For more examples, see §[10.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-10.html#jls-10.2).

In its current state javapoet lacks the expressivity regarding Java 8 `TYPE_USE` annotations and  makes the current usage `arraytype.annotated(typeAnnotation)` confusing. The annotation will be emitted preceding the type and hence applies to the component not to the array type.

This issue is related to #431.

ronshapiro:
Yes

rluble:
This issue is also solved by commit d91f8e1374f7ccdb68d92c206c84864cd09c02f6.

PR #615, opened 2018-02-19
Add tests for Java 8 TYPE_USE annotations on arrays (#614).

rluble:


rluble:
Yes I saw the way it applies to varargs. I'll add the tests. I guess they chose the ordering for the brackets on array types that way to match the ordering on array access and array instantiation where the first bracket is the outermost dimension, and that carried over to varargs which is a bit unnatural.

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -137,4 +140,79 @@ private void annotatedEquivalence(TypeName type) {
     String actual = type.toString();
     assertEquals(expected, actual);
   }
+
+  // https://github.com/square/javapoet/issues/614
+  @Ignore @Test public void annotatedArrayType() {
+    String expected =
+        "java.lang.Object @" + TypeUseAnnotation.class.getCanonicalName() + " []";
```
tbroyer: Should this introduce `TUA` and `TYPE_USE_ANNOTATION` constants like the `NN` and `NEVER_NULL` above? That would probably make the tests a bit easier to read. WDYT?

rluble: Good idea.

rluble:
I was looking into fixing the whole type use annotations and it requires a bit of refactoring the design. Currently the way annotations on types are handled is already quite hacky IMHO. Look at the difference between `TypeName.emit` and `TypeName.toString` which blur the line of where the responsibility for emitting annotations lie.

The only concern is that currently type annotations are treated as if they were annotations on the enclosing construct (`ParameterSpec`, etc etc) for most part. Fixing the issue might break questionable uses, i.e. someone using a `ElementType.METHOD` annotation on the return type of a method.

rluble:
I'll have to abandon this pull request and submit a new. It seems that I should be using my @google.com email instead of my personal email for the commits.

rluble:
Let's continue the review at #618

PR #617, opened 2018-02-19
Update the codebase now that we require Java 8

tbroyer:


src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -60,7 +60,7 @@ private ParameterizedTypeName(ParameterizedTypeName enclosingType, ClassName raw
 
   @Override public TypeName withoutAnnotations() {
     return new ParameterizedTypeName(
-        enclosingType, rawType, typeArguments, new ArrayList<AnnotationSpec>());
```
ronshapiro: I thinks can become 1 line now?

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -253,7 +252,7 @@ void emit(CodeWriter codeWriter, String enumName, Set<Modifier> implicitModifier
         Map.Entry<String, TypeSpec> enumConstant = i.next();
         if (!firstMember) codeWriter.emit("\n");
         enumConstant.getValue()
-            .emit(codeWriter, enumConstant.getKey(), Collections.<Modifier>emptySet());
+            .emit(codeWriter, enumConstant.getKey(), Collections.emptySet());
```
ronshapiro: One line here too

JakeWharton:
Needs a rebase now as well, when convenient.

tbroyer:
Rebased, addressed Ron's comments, and also replaced `Util.join` with `String.join`.

tbroyer:
What would you think about passing a `Supplier<String>` to `Util.check{Argument,NotNull,State}` to defer computation of some error message arguments (there's a `String.join()` and two `Arrays.toString()` in here for example, and `String#substring()`). That would move the `String.format()` call to every call point though (unless we only add it as an overload for the few cases pointed out above), and there are quite a few of them.
It'd change:
```java
checkArgument(unused.isEmpty(), "unused argument%s: %s", s, String.join(", ", unused));
```
to
```java
checkArgument(unused.isEmpty(), () -> String.format("unused argument%s: %s", s, String.join(", ", unused)));
```
…and we could then also replace `Util.checkNotNull` with `Objects.requireNonNull`.

JakeWharton:
That change STGM. Definitely in favor of using the JDK helpers.

PR #618, opened 2018-02-21
Add tests for Java 8 TYPE_USE annotations on arrays (#614).

rluble:


src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -114,27 +117,81 @@ private void annotatedEquivalence(TypeName type) {
...
-  // @Target(ElementType.TYPE_USE) requires Java 1.8
+  private final static String TUA = TypeUseAnnotation.class.getCanonicalName();
+  private final AnnotationSpec TYPE_USE_ANNOTATION =
+      AnnotationSpec.builder(TypeUseAnnotation.class).build();
+
+  @Target(ElementType.TYPE_USE)
   public @interface TypeUseAnnotation {}
 
   // https://github.com/square/javapoet/issues/431
   @Ignore @Test public void annotatedNestedType() {
```
swankjesse: remove `@Ignore` ?

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -114,27 +117,81 @@ private void annotatedEquivalence(TypeName type) {
...
-    assertEquals(expected, actual);
+        .annotated(TYPE_USE_ANNOTATION);
+    assertThat(type.toString())
+        .isEqualTo("java.util.Map.@" + TUA + " Entry<java.lang.Byte, java.lang.Byte>");
+  }
+
+  // https://github.com/square/javapoet/issues/614
+  @Ignore @Test public void annotatedArrayType() {
+    TypeName type = ArrayTypeName.of(ClassName.get(Object.class)).annotated(TYPE_USE_ANNOTATION);
+    assertThat(type.toString()).isEqualTo("java.lang.Object @" + TUA + " []");
```
swankjesse: this one seems counterintuitive to me. The annotation is on the outermost type, but it’s printed here?

(not saying it’s incorrect, but wow, this feature is weird)

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
```diff
@@ -114,27 +117,81 @@ private void annotatedEquivalence(TypeName type) {
...
+        .annotated(TYPE_USE_ANNOTATION));
+    MethodSpec varargsMethod = MethodSpec.methodBuilder("m")
+        .addParameter(
+            ParameterSpec.builder(type, "p")
+                .build())
+        .varargs()
+        .build();
+    assertThat(varargsMethod.toString()).isEqualTo(""
+        + "void m(\n"
+        + "    java.lang.Object[] @" + TUA + " ... p) {\n"
```
swankjesse: whoa

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
swankjesse: Nope, this PR is just tests!

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
rluble: I know.  See JLS [§9.7.4](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html)

>Type annotations can apply to an array type or any component type thereof ([§10.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-10.html#jls-10.1)). For example, assuming that `A`, `B`, and `C` are annotation types meta-annotated with `@Target(ElementType.TYPE_USE)`, then given the field declaration:
>
>```java
>@C int @A [] @B [] f;
>```
>`@A` applies to the array type `int[][]`, `@B` applies to its component type `int[]`, and `@C` applies to the element type `int`. For more examples, see [§10.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-10.html#jls-10.2).

My guess is that this was chosen so that is consistent with array creation and access where the first pair of brackets refer to the outermost dimension. Still it does not feel right.

src/test/java/com/squareup/javapoet/AnnotatedTypeNameTest.java
rluble: The other issue is that implementing these semantics requires either distinguishing between `TYPE_USE` annotations when annotating a type or change existing behavior which is a breaking change.

IMHO I like the simplicity of having implicit `TYPE_USE` semantics when annotating a type as the old usage is supported through ParameterSpec, MethodSpec, etc. I plan to make a PR with the implementation. Does this look as an acceptable solution or would you prefer to make the annotations aware whether they are `TYPE_USE` or not and emit them accordingly

swankjesse: Yup, that makes sense. Good idea!

rluble: I am not sure which approach you prefer. I assume that by yes you prefer the second choice, i.e. making annotations aware and emit accordingly, right?

swankjesse: Actually I’d prefer that annotations on TypeName objects are always interpreted as type annotations.

That matches the language model, and it's flexible.

tbroyer:
> it will be definitely simpler to model `ClassName` as a `ParameterizedTypeName` with no type parameters. 

That'd be a breaking change to anyone doing instanceof and expecting non-empty type parameters; but maybe introduce a `ParametrizableTypeName` common supertype?

rluble:
Went over all the tests in `AnnotatedTypeTest` and cleaned them up for consistency. @swankjesse could you review the changes.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -32,131 +32,260 @@
...
-  public static final ClassName OBJECT = ClassName.get(Object.class);
+public abstract class ClassName extends TypeName implements Comparable<ClassName> {
+  final String simpleName;
+  String canonicalName;
 
-  /** From top to bottom. This will be ["java.util", "Map", "Entry"] for {@link Map.Entry}. */
-  final List<String> names;
-  final String canonicalName;
+  /** A fully-qualified class name for top-level classes. */
+  private static final class TopLeveLClassName extends ClassName {
```
swankjesse: nit: LeveL –> Level

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -32,131 +32,260 @@
...
    * it is equivalent to {@code get(packageName(), name)}.
    */
-  public ClassName peerClass(String name) {
-    List<String> result = new ArrayList<>(names);
-    result.set(result.size() - 1, name);
-    return new ClassName(result);
-  }
+  public abstract ClassName peerClass(String name);
+
+  protected abstract ClassName prefixWithAtMostOneAnnotatedClass();
```
swankjesse: package-private? users don’t need to see this do they?

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -32,131 +32,260 @@
...
+
+  protected abstract boolean hasAnnotatedEnclosingClass();
```
swankjesse: same

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -32,131 +32,260 @@
...
+
+  protected abstract CodeWriter emitWithoutPrefix(
```
swankjesse: same

swankjesse:
I’m torn. I think your implementation is pretty much perfect. It is as simple as it could be, given the complexity of the annotations-on-anything feature. And it composes nicely.

But I wonder if it’s worth the complexity of supporting this feature in JavaPoet at all. We’re striving to balance simplicity and completeness. This is a case where completeness potentially harms simplicity? Like, do we really need to expose all the different absurd ways that type annotations can be applied? Is there demand from users to do this?

I think we probably want this. But I wanted to complain about how type annotations were a bad idea for Java and how the cost of the feature dramatically outweighs the benefits.

rluble:
The problem is that as it stands, without proper support for type use annotations, it makes it really difficult to emit code that compiles in Java if you need to use those annotations (there is also the loss of expressivity). In my use case, there is the need to synthesize code with the proper `@Nullable`, `@Nonnull` annotations (the checker framework ones).

src/main/java/com/squareup/javapoet/ClassName.java
rluble: done.

src/main/java/com/squareup/javapoet/ClassName.java
rluble: done

src/main/java/com/squareup/javapoet/ClassName.java
rluble: done

src/main/java/com/squareup/javapoet/ClassName.java
rluble: done

rluble:
> It is as simple as it could be

:) There are a couple of things that could be simplified further. One is the interaction between TypeName and CodeWriter re: lookupName (it'd be nice to remove the remaining string parsing operations there). The second is that having both ClassName and ParameterizedTypeName is pretty much unnecessary since ClassName is just a special case. But I guess that is pretty much the API of javapoet and a breaking change there would be rather painful (even if ClassName static members were preserved to create the 0-parameter instances).

rluble:
Any chance this PR could be approved and merged in the near future?

swankjesse:
Thanks.

Issue #622, opened 2018-02-26
Refactor indent method to accept type and number of indents


DrPeroxide:
The current implementation of the indent method is questionable, accepting any old String and appending this to the beginning of every line with no validation done as to the validity of the indentation provided.

I recommend adding a method that sets the indentation type to either SPACE or TAB, along with another method accepting an integer to describe how many of these should be appended before each line.

JakeWharton:
Would you ever expect multiple tabs? Also is this a problem in practice? We could add a check to ensure that it's filled with whitespace and that seems like it would cover any accidental value.

PR #623, opened 2018-02-27
Refactor configurable indent(#622)

DrPeroxide:
Enhanced indent feature with new methods for setting the indent. Have worked around the original indent method, which setting the String directly, for the sake of backwards compatibility, but I recommend adding the @deprecated annotation to that method.

All development work has followed TDD and comes with JavaDocs.

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -268,6 +268,13 @@ public final class JavaFile {
       return this;
     }
 
+    public Builder indent(int indentCount) {
+      char[] indentChars = new char[indentCount];
+      Arrays.fill(indentChars, ' ');
+      this.indent = new String(indentChars);
+      return this;
+    }
+
     public JavaFile build() {
       return new JavaFile(this);
     }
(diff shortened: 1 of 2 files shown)


swankjesse:
I think this is a ton of code for very marginal new utility. We really want this library to be as small as possible. This feature is overkill. I'd almost prefer to get rid of configurable indents altogether.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -263,13 +264,80 @@ public Builder skipJavaLangImports(boolean skipJavaLangImports) {
...
+    /**
+     * Sets the indentation of the Java file to the given String, filtering out any invalid characters, such as letters or
+     * symbols. Uses the {@link IndentChar} enum to define which characters are allowed.
+     *
+     * @param indent Value of the indent
+    */
     public Builder indent(String indent) {
-      this.indent = indent;
+      String allowedChars = Arrays.stream(IndentChar.values())
+                                  .map(IndentChar::toString).reduce((a, b) -> a+b).orElse(" ");
```
tbroyer: That `reduce` looks a lot like a `.collect(joining())`

PR #625, opened 2018-03-02
Use contextual type parameters for throws types.

NightlyNexus:


src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -177,6 +182,12 @@ private ExecutableElement findFirst(Collection<ExecutableElement> elements, Stri
         + "@java.lang.Override\n"
         + "public int compareTo(" + ExtendsOthers.class.getCanonicalName() + " arg0) {\n"
         + "}\n");
+    exec = findFirst(methods, "fail");
+    method = MethodSpec.overriding(exec, classType, types).build();
+    assertThat(method.toString()).isEqualTo(""
+        + "@java.lang.Override\n"
+        + "public void fail() throws java.lang.IllegalStateException {\n"
```
NightlyNexus: instead of `throws  R`.
i'll try to remember to check KotlinPoet this weekend, too.

NightlyNexus: there's no way to represent
```java
interface Throws<R extends RuntimeException> {
  void fail() throws R;
}
```
in Kotlin, as far as I can tell!

PR #626, opened 2018-03-02
Disable overriding methods in final classes.

NightlyNexus:


src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -188,6 +188,10 @@ public static Builder constructorBuilder() {
   public static Builder overriding(ExecutableElement method) {
     checkNotNull(method, "method == null");
 
+    if (method.getEnclosingElement().getModifiers().contains(Modifier.FINAL)) {
+      throw new IllegalArgumentException("Cannot override method on final class.");
```
JakeWharton: Include the class name?

NightlyNexus: yep.

PR #627, opened 2018-03-02
Add another test for direct generics usage.

NightlyNexus:
and remove Mockito from MethodSpecTest.

### DIFF
--- a/src/test/java/com/squareup/javapoet/MethodSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/MethodSpecTest.java
@@ -15,7 +15,6 @@
  */
 package com.squareup.javapoet;
 
-import com.google.common.collect.ImmutableSet;
 import com.google.testing.compile.CompilationRule;
 import java.io.Closeable;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -151,9 +176,7 @@ private ExecutableElement findFirst(Collection<ExecutableElement> elements, Stri
     TypeElement classElement = getElement(ExtendsIterableWithDefaultMethods.class);
     DeclaredType classType = (DeclaredType) classElement.asType();
     List<ExecutableElement> methods = methodsIn(elements.getAllMembers(classElement));
-    ExecutableElement exec = findFirst(methods, "iterator");
-    assume().that(Util.DEFAULT).isNotNull();
```
JakeWharton: Why is `Util.DEFAULT` still a thing? Shouldn't it be inlined everywhere now as `Modifier.DEFAULT`?

NightlyNexus: #617

JakeWharton: Ah I thought that had landed

PR #628, opened 2018-03-03
Refactor preconditions with a message supplier

tbroyer:
In some cases, this allows deferring computation of some of
the String.format() parameters.
Also, Objects.requireNonNull replaces Util.checkNotNull.


tbroyer:
I had to add a few "effectively final" local variables, and I changed a few `String.format()` into simple string concatenations.
Regarding formatting, I settled on wrapping before the lambda, except when the line would be longer than 100 chars in which case I wrapped after the "arrow".

Let me know if you'd prefer that I revert any part of those changes.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -199,8 +199,8 @@ public JavaFileObject toJavaFileObject() {
   }
 
   public static Builder builder(String packageName, TypeSpec typeSpec) {
-    checkNotNull(packageName, "packageName == null");
-    checkNotNull(typeSpec, "typeSpec == null");
+    Objects.requireNonNull(packageName, "packageName == null");
+    Objects.requireNonNull(typeSpec, "typeSpec == null");
```
JakeWharton: Static import?

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -41,15 +41,15 @@
   private ParameterizedTypeName(ParameterizedTypeName enclosingType, ClassName rawType,
       List<TypeName> typeArguments, List<AnnotationSpec> annotations) {
     super(annotations);
-    this.rawType = checkNotNull(rawType, "rawType == null");
+    this.rawType = Objects.requireNonNull(rawType, "rawType == null");
```
JakeWharton: Import?

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -109,27 +109,27 @@ public boolean hasModifier(Modifier modifier) {
   }
 
   public static Builder classBuilder(String name) {
-    return new Builder(Kind.CLASS, checkNotNull(name, "name == null"), null);
+    return new Builder(Kind.CLASS, Objects.requireNonNull(name, "name == null"), null);
```
JakeWharton: Static import

tbroyer:
> Just import nits

Oops! I'd have sworn I reviewed every file…

Updated.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -203,8 +203,8 @@ public Builder addMember(String name, String format, Object... args) {
     }
 
     public Builder addMember(String name, CodeBlock codeBlock) {
-      checkNotNull(name, "name == null");
-      checkArgument(SourceVersion.isName(name), "not a valid name: %s", name);
+      requireNonNull(name, "name == null");
+      checkArgument(SourceVersion.isName(name), () -> "not a valid name: " + name);
```
swankjesse: I don't think I like this idiom for Java. Should we just use an if() clause?

PR #630, opened 2018-03-06
Improve compatibility with source-based ABI generation

dreiss:


JakeWharton:
Are there tests we can add to not regress these prior problems?

dreiss:
They only show up in the highly unusual compilation environment that Buck creates during source-only ABI generation.  I can create a mock that would be accurate enough to detect this particular issue.

dreiss:
Updated with test.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -205,20 +205,27 @@ public static ClassName get(String packageName, String simpleName, String... sim
   public static ClassName get(TypeElement element) {
     checkNotNull(element, "element == null");
     List<String> names = new ArrayList<>();
-    for (Element e = element; isClassOrInterface(e); e = e.getEnclosingElement()) {
-      checkArgument(element.getNestingKind() == TOP_LEVEL || element.getNestingKind() == MEMBER,
+    TypeElement e = element;
+    // Add nested class names one at a time before adding the package.
```
swankjesse: the comment says what it does. Can you add why?

src/test/java/com/squareup/javapoet/SimpleElementName.java
```diff
@@ -0,0 +1,39 @@
+package com.squareup.javapoet;
```
swankjesse: supernit: copyright header

src/test/java/com/squareup/javapoet/SimpleElementName.java
```diff
@@ -0,0 +1,39 @@
...
+
+import javax.lang.model.element.Name;
+
+/**
+ * Simple implementation of {@link Name} for testing.
+ */
+public class SimpleElementName implements Name {
```
swankjesse: off topic: it’s so weird that this API _is a _ CharSequence rather than _has a_ char sequence.

src/main/java/com/squareup/javapoet/ClassName.java
dreiss: Sure.  I'm not 100% sure why (the old code was already doing this), but I think it's just "`because {@link #names} expects the package first as a single item.`"

dreiss:
Updated comments.

swankjesse:
Last thing:

dreiss:
Facebook's lawyer told me that I'm not supposed to sign that and that we signed a corporate CLA instead.  I'll get the name of the person on Square's side who knows about it.

dreiss:
Romi Akpala on the IP legal team.

PR #629, opened 2018-03-04
Add JDK 11 and load install-jdk.sh on-the-fly

sormuras:
No more updates needed when new EA feature builds are published or paths to existing archives are changed.

PR #642, opened 2018-04-23
Upgrade errorprone to 2.3.1

sormuras:
Use latest `errorprone` version, which seems to be Java 11-ready.


sormuras:
...merging #629 first, will make this PR happy. :)

swankjesse:
thanks!

PR #643, opened 2018-04-28
Tighten up the code that handles emitting type annotations.

swankjesse:
I found the previous code difficult to follow. Hopefully this is less so.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -17,10 +17,9 @@ package com.squareup.javapoet;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.PackageElement;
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -32,232 +31,106 @@
...
+        concatAnnotations(annotations));
   }
 
+  @Override public ClassName withoutAnnotations() {
+    ClassName resultEnclosingClassName = enclosingClassName != null
+        ? enclosingClassName.withoutAnnotations()
+        : null;
+    return isAnnotated()
+        ? new ClassName(packageName, resultEnclosingClassName, simpleName)
+        : this;
```
JakeWharton: I'm not sure this ternary should exist. What does `TypeName.get(Map.class).annotatedWith(..).nestedClass("Entry").withoutAnnotations()` produce?

swankjesse: Just saving allocations when it's a no-op!

swankjesse: Gotta get back in a Java desktop mood. This doesn't need every allocation to be justified.

swankjesse: Ahhh, I see so it's also incorrect! Good catch!

JakeWharton: Or, put another way, this is equivalent to
```java
if (!isAnnotated()) return this;
ClassName resultEnclosingClassName = enclosingClassName != null
    ? enclosingClassName.withoutAnnotations()
    : null;
return new ClassName(packageName, resultEncosingClassName, simpleName)
```
Does `isAnnotated()` recurse on `enclosingClassName` for annotations?

swankjesse: It doesn’t. I’ll fix.

PR #646, opened 2018-05-09
Fix TypeName.get() on top-level error types

ronshapiro:


ronshapiro:
One thing I noticed: if there's a type that can be resolved that has type parameters, javac seems to give a TypeMirror that's simple name is [`"<any>"`](http://hg.openjdk.java.net/jdk9/jdk9/langtools/file/5ec5a14e1627/src/jdk.compiler/share/classes/com/sun/tools/javac/util/Names.java#l241). 

I imagine that that was never supported (I don't see any way to get out of that), but maybe that means we should throw on ErrorTypes? That would be a behavioral change though, and it's not clear to me that that's what we should do.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -206,18 +208,17 @@ public static ClassName get(TypeElement element) {
...
-        return new ClassName(packageName.toString(), null, simpleName);
-
-      case MEMBER:
-        ClassName enclosingClass = get((TypeElement) element.getEnclosingElement());
-        return enclosingClass.nestedClass(simpleName);
-
-      default:
-        throw new IllegalArgumentException("unexpected type nesting");
+    Element enclosingElement = element.getEnclosingElement();
+    if (enclosingElement instanceof PackageElement) {
```
swankjesse: Any way to do this without `instanceof`? My read of the docs is that a single class is allowed to implement multiple types.

src/test/java/com/squareup/javapoet/AbstractTypesTest.java
```diff
@@ -68,10 +77,38 @@ private TypeMirror getMirror(Class<?> clazz) {
         .isEqualTo(ParameterizedTypeName.get(ClassName.get(Set.class), ClassName.OBJECT));
   }
 
-  @Test public void getErrorType() {
-    ErrorType errorType =
-        new DeclaredTypeAsErrorType(getTypes().getDeclaredType(getElement(Set.class)));
-    assertThat(TypeName.get(errorType)).isEqualTo(ClassName.get(Set.class));
+  @Test public void errorTypes() {
```
swankjesse: Great test.

Issue #478, opened 2016-07-13
access to TypeSpec.Builder state (modifiable TypeSpec)


asereda:
Hello,

What is the easiest way to create a modified TypeSpec from existing one ?  It is easy to use ```toBuilder()``` method and add new specs (methods / annotations / fields), however how do one removes / replaces / renames certain elements in a spec ? 

I would like to avoid recreating it from scratch (ie manually copying javadoc / superclass / interfaces / static blocks etc.).

Am I missing something ?

Thanks

asereda:
To give more context suppose one has an existing typespec describing the following class

## Before
```
@Generated
public class MyClass {
    public int identity(int value) {
        return value;
    }
}
```
## After
How to I generate a slightly different class like below
```
@Generated(date="2016-07-02T11:22:00-04:00")
public class MyClass {
    public long identity(long value) {
        return value;
    }
}
```

swankjesse:
We currently lack APIs to remove things on a builder. But we need ’em. Maybe just exposing the raw mutable list fields on the builders is sufficient.

asereda:
Probably it is a good start. Not sure if there is anything better

yevhen13:
> exposing the raw mutable list fields on the builders

This reminds me that I had a similar task once, it was actually related to MethodSpecs though.
And the solution I came up with was [this](https://gist.github.com/yevhen13/d901d5df6591eba6938cc7e0b3f2c03e).

PR #650, opened 2018-06-03
Mutable builder list fields

ZacSweers:
A case that often comes up for us is that if you expose a post processing API in code gen, builders become a pain because any list property in a builder is append-only. 

What this means is that in order to modify one of them, you have to manually make a new copy of the type that builder is constructing and copy over every property but the one you're trying to change. Then for the one you're changing, you have to copy all the other ones and then your modified version.

A classic example is a typespec with a methodspec in it that you want to add an annotation to.

After talking with @swankjesse about it, we decided to take the approach of making builders' collection fields public and mutable with the idea that if you're manipulating those directly, we just assume you know what you're doing and defer to checks in `build()` to enforce things.

### DIFF
--- a/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
@@ -109,11 +109,28 @@ public class ParameterSpecTest {
     modifiers.add(Modifier.PUBLIC);
 
     try {
-      ParameterSpec.builder(int.class, "foo").addModifiers(modifiers);
+      ParameterSpec.builder(int.class, "foo")
+          .addModifiers(modifiers);
       fail();
     } catch (Exception e) {
-      assertThat(e.getMessage())
-          .isEqualTo("unexpected parameter modifier: public");
+      assertThat(e.getMessage()).isEqualTo("unexpected parameter modifier: public");
     }
+  }
+
+  @Test public void modifyAnnotations() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "foo")
+            .addAnnotation(Override.class)
+            .addAnnotation(SuppressWarnings.class);
+
+    builder.annotations.remove(1);
+    assertThat(builder.build().annotations).hasSize(1);
+  }
+
+  @Test public void modifyModifiers() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "foo")
+            .addModifiers(Modifier.PUBLIC, Modifier.STATIC);
+
+    builder.modifiers.remove(1);
+    assertThat(builder.build().modifiers).containsExactly(Modifier.PUBLIC);
   }
 }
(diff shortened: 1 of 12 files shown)


hzsweers:
I started moving checks to `build()`, but wanted to check first - do we want to keep early checks in the main APIs or just move them whole sale to `build()`?

ronshapiro:
I don't fully understand the point of this, and my initial instinct is that it's probably going to be abused more often than it is actually useful. I presume, for one, that we wouldn't want someone to call `builder.someList.add()` and would instead prefer `builder.addSomething()`.

Perhaps you could provide a more real-world example, or point to some code that would benefit from this

sormuras:
Examples: #341 and #478 and other I don't remember anymore.

One use-case is to build up a type spec and enhance it afterwards with information gathered from that same type spec. Without the need to have an extra meta-model from which you would generate the "final type spec" in one go.

In "beethoven" I called the mutators, components that work on the mutable model, "composers": https://github.com/sormuras/beethoven#composers

hzsweers:
Another example in a production library: Thrifty has support for post processing that hands over the `TypeSpec`. If all you're doing is adding to the top level type, it's sort of fine. If you want to add anything to anything within it that supports multiple instances (methods, fields, annotations, etc), then you run into this problem.

https://github.com/Microsoft/thrifty/blob/master/thrifty-example-postprocessor/src/main/java/com/microsoft/thrifty/compiler/SerializableTypeProcessor.java

hzsweers:
Basically - it feels strange to have a `toBuilder()` API that only allows replacement/destructive operations on *some* of its properties.

swankjesse:
This is great. Do you mind doing a quick validation pass in `build()`? For example, we shouldn’t have enum constants in TypeSpec unless it’s an enum type?

hzsweers:
Moved validation to build() where appropriate, which really only TypeSpec and AnnotationSpec had validations to (or that made sense to) move

ZacSweers:
I've rebased this, definitely would like to still see this merged!

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -25,6 +25,7 @@
 import javax.lang.model.util.Elements;
 import org.junit.Before;
 import org.junit.Rule;
+import javax.lang.model.element.Modifier;
```
Egorand: nit: this import looks misplaced

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -17,6 +17,14 @@
 
 import com.google.common.collect.ImmutableMap;
 import com.google.testing.compile.CompilationRule;
+
+import org.junit.Rule;
```
Egorand: nit: misplaced imports?

PR #652, opened 2018-06-28
Parameterized type arguments autoboxed

stechio:
_Current implementation of ParameterizedTypeName constructor rejects primitive and void type parameters_: that's a pretty annoying and useless restriction, as primitive and void are obviously expected to get boxed instead of being rejected. This pull request amends that behaviour.


kenzierocks:
I'm not sure this is a great idea, because it will become breaking behavior to change it once Project Valhalla lands in the JDK, and primitive types can be used directly in parameters. On the other hand, if it's an exception, it's not breaking to start allowing it.

swankjesse:
Yeah, I don't think we want to autocorrect your generated code.

PR #657, opened 2018-07-26
Qualify types masked by type variables

ronshapiro:
Originally reported here: https://github.com/google/dagger/issues/1230

Type variables technically have preference even over types in the same package, so if one exists with the same name, we should ignore any package information.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -57,7 +57,7 @@ final class CodeWriter {
   private final Map<String, ClassName> importedTypes;
   private final Map<String, ClassName> importableTypes = new LinkedHashMap<>();
   private final Set<String> referencedNames = new LinkedHashSet<>();
-  private final Set<String> currentTypeVariables = new LinkedHashSet<>();
+  private final Multiset<String> currentTypeVariables = new Multiset<>();
   private boolean trailingNewline;
 
   /**
@@ -506,5 +506,27 @@ final class CodeWriter {
     Map<String, ClassName> result = new LinkedHashMap<>(importableTypes);
     result.keySet().removeAll(referencedNames);
     return result;
+  }
+
+  // A makeshift multi-set implementation
+  private static final class Multiset<T> {
+    private final Map<T, Integer> map = new LinkedHashMap<>();
+
+    void add(T t) {
+      int count = map.getOrDefault(t, 0);
+      map.put(t, count + 1);
+    }
+
+    void remove(T t) {
+      int count = map.getOrDefault(t, 0);
+      if (count == 0) {
+        throw new IllegalStateException(t + " is not in the multiset");
+      }
+      map.put(t, count - 1);
+    }
+
+    boolean contains(T t) {
+      return map.getOrDefault(t, 0) > 0;
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -203,6 +206,10 @@ public void emitTypeVariables(List<TypeVariableName> typeVariables) throws IOExc
     emit(">");
   }
 
+  public void popTypeVariables(List<TypeVariableName> typeVariables) throws IOException {
+    typeVariables.forEach(typeVariable -> currentTypeVariables.remove(typeVariable.name));
```
tbroyer: Won't this cause issues with type variable shadowing? (or will the compiler reject it anyway?)

Even without shadowing, how about type variables of the same name in inner elements? (i.e. a static method or an inner class)

E.g. when generating
```java
class MyClass<T> {
  static <T> newInstance(T foo) { return new MyClass<T>(foo); }

  private final com.squareup.tacos.T<T> foo;

  private MyClass(com.squareup.tacos.T<T> foo) { this.foo = foo; }

  public com.squareup.tacos.T<T> getFoo() { return foo; }
}
```
won't this remove `T` from the set when exiting the `newInstance` method?

A more realistic example being:
```java
class MyClass<T> {
  static class Inner<T> {}

  private com.squareup.tacos.T<T> foo;
}
```
Will the `T` type variable still be in scope when exiting `Inner`?

src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -372,9 +385,9 @@ String lookupName(ClassName className) {
       return className.canonicalName;
     }
 
-    // If the class is in the same package, we're done.
+    // If the class is in the same package, we're done
```
JakeWharton: 😮 

We might have to fight. I take it you don't subscribe to [comments should be sentences][1]?

 [1]: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

JakeWharton:
Ugh we need to get off of Travis and these flaky scripts that set up the build environment and onto something that lets us use a Docker image.

src/main/java/com/squareup/javapoet/CodeWriter.java
ronshapiro: Forgive me master! 

I had initially appended to that comment because I had changed the conditional, but then backed it out. And I tried to restore the comment and forget the period. Reverted :)

src/main/java/com/squareup/javapoet/CodeWriter.java
ronshapiro: You're right. I don't think there's a way to then reference the "outer" `T`, but we do still need to have something closer to a multiset

ronshapiro: @JakeWharton @swankjesse do you have an alternative approach to a multiset? I know we don't depend on Guava here, so I was about to implement a Map<String, Integer>, but that felt gross, so I was thinking of creating a private Multiset type to wrap it. Thoughts?

ronshapiro: I went with the basic multiset approach, but happy to switch to another alternative.

ronshapiro:
@JakeWharton can you restart the build and see if it deflakes?

ronshapiro:
Do we need to restart the Travis build?

sormuras:
Maybe rebase on current master. The build script was updated lately.

ronshapiro:
Looks like its passing now :)

PR #664, opened 2018-08-03
Change AnnotationSpec.type to be a ClassName

ronshapiro:
Annotation types _must_ be Classes, right? I think that's reflected in the API for the builder that it only accepts `Class<?>` or `ClassName`

Note that this would technically be a binary-incompatible change.

I'd be happy to also make this change on the KotlinPoet side of things too.


ronshapiro:
@swankjesse can you requeue this travis build too?

ronshapiro:
Ping

PR #674, opened 2018-09-07
Allow multiple TypeSpecs to be added to a JavaFile

kevinherron:
This change allows multiple `TypeSpec`s to be added to a single `JavaFile`.

This allows files like this to be generated:

```java
package com.squareup.tacos;

import java.lang.String;
import java.lang.StringBuilder;

public interface Foo {
  String getBar();
}
class FooImpl implements Foo {
  public String getBar() {
    StringBuilder sb = new StringBuilder();
    sb.append("bar");
    return sb.toString();
  }
}
```

A more compelling/complete example would have a static factory method on the interface that built and returned the package-private implementation but this should illustrate the point well enough.

There's room for improvement by validating that the first `TypeSpec` has as either public or package visibility and that any additional specs have at most package visibility.

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -24,12 +24,14 @@ import java.io.Writer;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 import javax.annotation.processing.Filer;
 import javax.lang.model.element.Element;
 import javax.tools.JavaFileObject;
@@ -56,7 +58,7 @@ public final class JavaFile {
 
   public final CodeBlock fileComment;
   public final String packageName;
-  public final TypeSpec typeSpec;
+  public final List<TypeSpec> typeSpecs;
   public final boolean skipJavaLangImports;
   private final Set<String> staticImports;
   private final String indent;
@@ -64,7 +66,7 @@ public final class JavaFile {
   private JavaFile(Builder builder) {
     this.fileComment = builder.fileComment.build();
     this.packageName = builder.packageName;
-    this.typeSpec = builder.typeSpec;
+    this.typeSpecs = builder.typeSpecs;
     this.skipJavaLangImports = builder.skipJavaLangImports;
     this.staticImports = Util.immutableSet(builder.staticImports);
     this.indent = builder.indent;
@@ -93,7 +95,8 @@ public final class JavaFile {
       Files.createDirectories(outputDirectory);
     }
 
-    Path outputPath = outputDirectory.resolve(typeSpec.name + ".java");
+    TypeSpec firstTypeSpec = typeSpecs.get(0);
+    Path outputPath = outputDirectory.resolve(firstTypeSpec.name + ".java");
     try (Writer writer = new OutputStreamWriter(Files.newOutputStream(outputPath), UTF_8)) {
       writeTo(writer);
     }
@@ -106,10 +109,12 @@ public final class JavaFile {
 
   /** Writes this to {@code filer}. */
   public void writeTo(Filer filer) throws IOException {
+    TypeSpec firstTypeSpec = typeSpecs.get(0);
     String fileName = packageName.isEmpty()
-        ? typeSpec.name
-        : packageName + "." + typeSpec.name;
-    List<Element> originatingElements = typeSpec.originatingElements;
+        ? firstTypeSpec.name
+        : packageName + "." + firstTypeSpec.name;
+    List<Element> originatingElements = new ArrayList<>();
+    typeSpecs.forEach(ts -> originatingElements.addAll(ts.originatingElements));
     JavaFileObject filerSourceFile = filer.createSourceFile(fileName,
         originatingElements.toArray(new Element[originatingElements.size()]));
     try (Writer writer = filerSourceFile.openWriter()) {
@@ -153,7 +158,9 @@ public final class JavaFile {
       codeWriter.emit("\n");
     }
 
-    typeSpec.emit(codeWriter, null, Collections.emptySet());
+    for (TypeSpec ts : typeSpecs) {
+      ts.emit(codeWriter, null, Collections.emptySet());
+    }
 
     codeWriter.popPackage();
   }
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -56,15 +58,15 @@
 
   public final CodeBlock fileComment;
   public final String packageName;
-  public final TypeSpec typeSpec;
+  public final List<TypeSpec> typeSpecs;
```
JakeWharton: This is a breaking change

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -93,7 +95,8 @@ public void writeTo(Path directory) throws IOException {
       Files.createDirectories(outputDirectory);
     }
 
-    Path outputPath = outputDirectory.resolve(typeSpec.name + ".java");
+    TypeSpec firstTypeSpec = typeSpecs.get(0);
```
JakeWharton: This might not be the public type

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -106,10 +109,12 @@ public void writeTo(File directory) throws IOException {
 
   /** Writes this to {@code filer}. */
   public void writeTo(Filer filer) throws IOException {
+    TypeSpec firstTypeSpec = typeSpecs.get(0);
```
JakeWharton: Same

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -153,7 +158,9 @@ private void emit(CodeWriter codeWriter) throws IOException {
       codeWriter.emit("\n");
     }
 
-    typeSpec.emit(codeWriter, null, Collections.emptySet());
+    for (TypeSpec ts : typeSpecs) {
```
JakeWharton: `ts` -> `typeSpec`

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -180,9 +187,10 @@ private void emit(CodeWriter codeWriter) throws IOException {
   }
 
   public JavaFileObject toJavaFileObject() {
+    TypeSpec firstTypeSpec = typeSpecs.get(0);
```
JakeWharton: Again, might not be public

src/main/java/com/squareup/javapoet/JavaFile.java
kevinherron: Oops. I'll rework this.

src/main/java/com/squareup/javapoet/JavaFile.java
kevinherron: Could it be that by convention (and documentation) file name will be the first public or package visible TypeSpec added to the JavaFile?

kevinherron:
> We didn't add this feature because we couldn't think of any value gained by doing this vs. multiple files. I'm still not convinced it's needed.

It's legal Java code, a fairly trivial change to implement and maintain, and this *is* a library for generating Java code, after all...

I've got a few hundred structures to auto-generate and was hoping to cut down the number of files. You're right that it's mostly a cosmetic issue, but so are a lot of things.

kevinherron:
Currently uses the first/original/non-additional (whatever) `TypeSpec` as the file name. Not sure what approach would be best.

ronshapiro:
From a readability perspective, generating hundreds of types in the same file sounds like a nightmare. I think unless presented with a good use case + evidence, this seems like something not worth the extra maintenance cost.

kevinherron:
@ronshapiro 2 types per file: the public interface and the protected implementation.

I have no intention of generating hundreds per file :)

kevinherron:
Here's a more detailed use case of the pattern I'm talking about:

```java
import javax.annotation.Nullable;

import com.google.auto.value.AutoValue;

public interface Foo {

    int p1();

    @Nullable
    String p2();

    static Foo create(int p1, String p2) {
        return builder()
            .p1(p1)
            .p2(p2)
            .build();
    }

    static Builder builder() {
        return new AutoValue_FooBase.Builder();
    }

    abstract class Builder {
        abstract Builder p1(int p1);

        abstract Builder p2(String p2);

        abstract Foo build();
    }

}

@AutoValue
abstract class FooBase implements Foo {

    @AutoValue.Builder
    abstract static class BuilderImpl extends Builder {
        abstract BuilderImpl p1(int p1);

        abstract BuilderImpl p2(String p2);

        abstract FooBase build();
    }

}
```

The use of AutoValue is immaterial here. The focus should be on the interface/impl pattern happening.

ronshapiro:
What about `Foo` and `Foo.Base`? If they're in pairs, and the naming is similar...

kevinherron:
A nested `Foo.Base` would have public visibility instead of package private :/

kevinherron:
@ronshapiro's suggestion led me to take another look and I've come up with a passable single-file solution that will work.

If you guys really don't want to merge this feature that's your prerogative. I can always shade my fork into the code generator if I need it at some point in the future.

For the morbidly curious:

```java
public abstract class Bar {

    Bar() {}

    public abstract int p1();

    @Nullable
    public abstract String p2();

    public static Bar create(int p1, String p2) {
        return builder()
            .p1(p1)
            .p2(p2)
            .build();
    }

    public static BarBuilder builder() {
        return null;
        // return new AutoValue_Bar_BarStruct.Builder();
    }

    public abstract static class BarBuilder {
        public abstract BarBuilder p1(int p1);

        public abstract BarBuilder p2(String p2);

        public abstract Bar build();
    }

    @AutoValue
    abstract static class BarStruct extends Bar {

        @AutoValue.Builder
        abstract static class BarStructBuilder extends BarBuilder {
            public abstract BarStructBuilder p1(int p1);

            public abstract BarStructBuilder p2(String p2);

            public abstract BarStruct build();
        }

    }
}
```

```java
public abstract class BarSub extends Bar {

    BarSub() {}

    @Override
    public abstract int p1();

    @Nullable
    @Override
    public abstract String p2();

    public abstract long p3();

    public static BarSub create(int p1, String p2, long p3) {
        return builder()
            .p1(p1)
            .p2(p2)
            .p3(p3)
            .build();
    }

    public static BarSubBuilder builder() {
        return null;
        // return new AutoValue_BarSub_BarSubStruct.Builder();
    }

    public abstract static class BarSubBuilder extends BarBuilder {
        public abstract BarSubBuilder p1(int p1);

        public abstract BarSubBuilder p2(String p2);

        public abstract BarSubBuilder p3(long p3);

        public abstract BarSub build();
    }

    @AutoValue
    abstract static class BarSubStruct extends BarSub {

        @AutoValue.Builder
        abstract static class BarSubStructBuilder extends BarSubBuilder {
            public abstract BarSubStructBuilder p1(int p1);

            public abstract BarSubStructBuilder p2(String p2);

            public abstract BarSubStructBuilder p3(long p3);

            public abstract BarSubStruct build();
        }

    }

}
```

Sub-type relationships preserved, no duplicated/unused fields, instance of `BarSub` assignable to `Bar`, and I can implement `equals`/`hashCode` that aren't broken (or let AutoValue do it).

PR #676, opened 2018-09-11
Add Javadoc to ParameterSpec

### LINKED ISSUE #495: ParameterSpec.addJavadoc may be usefull
```
MethodSpec dismiss = MethodSpec.methodBuilder("dismiss")
    .addJavadoc("Some documentation")
    .addParameter(ParameterSpec.builder(String.class, "message")
        .addJavadoc("the message to send").build())
    .build();
```

would generate :

```
/**
 * Some documentation.
 * @param message the message to send.
 */
void dismiss(String message);
```

Note : the dot automatically added to the end of the @param doc line would be great.

shaishavgandhi05:
Resolves #495 

Wanted to get some feedback. Right now, to ensure that the `@param name description of name` comes to the end of the Javadoc, I've kept a separate field called `parameterJavadoc`. I'm unsure whether that is the best practice since if we think about adding Javadoc to return types and exception, it might be a bit much. 

The alternative is to put the onus on the consumer to then do:
```java
methodSpec
  .addJavadoc("Method description")
  .addParameter(ParameterSpec.builder(String.class, "name")
      .addJavadoc("The name of the user\n")
      .build())
```
and make sure that the method javadoc comes before the parameter javadoc.

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -142,7 +142,7 @@ public final class MethodSpec {
 
   private CodeBlock javadocWithParameters() {
     CodeBlock.Builder builder = javadoc.toBuilder();
-    for (ParameterSpec parameterSpec: parameters) {
+    for (ParameterSpec parameterSpec : parameters) {
       if (!parameterSpec.javadoc.isEmpty()) {
         builder.add("@param $L $L", parameterSpec.name, parameterSpec.javadoc);
       }
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -145,7 +145,8 @@ public final class MethodSpec {
     boolean emitTagNewline = true;
     for (ParameterSpec parameterSpec : parameters) {
       if (!parameterSpec.javadoc.isEmpty()) {
-        if (emitTagNewline) builder.add("\n");
+        // Emit a new line before @param section only if the method javadoc is present.
+        if (emitTagNewline && !javadoc.isEmpty()) builder.add("\n");
         emitTagNewline = false;
         builder.add("@param $L $L", parameterSpec.name, parameterSpec.javadoc);
       }
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -63,6 +63,8 @@ public final class MethodSpec {
         "last parameter of varargs method %s must be an array", builder.name);
 
     this.name = checkNotNull(builder.name, "name == null");
+    // Add parameter Javadoc to the method Javadoc
+    builder.javadoc.add(builder.parameterJavadoc.build());
     this.javadoc = builder.javadoc.build();
     this.annotations = Util.immutableList(builder.annotations);
     this.modifiers = Util.immutableSet(builder.modifiers);
@@ -281,6 +283,7 @@ public final class MethodSpec {
     private final String name;
 
     private final CodeBlock.Builder javadoc = CodeBlock.builder();
+    private final CodeBlock.Builder parameterJavadoc = CodeBlock.builder();
     private final List<AnnotationSpec> annotations = new ArrayList<>();
     private final List<Modifier> modifiers = new ArrayList<>();
     private List<TypeVariableName> typeVariables = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


shaishavgandhi05:
I checked out the KotlinPoet implementation and it seems we don't really care about the order of the `@param`. I can model that here also if someone can confirm that's what we'd like!

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -63,6 +63,8 @@ private MethodSpec(Builder builder) {
         "last parameter of varargs method %s must be an array", builder.name);
 
     this.name = checkNotNull(builder.name, "name == null");
+    // Add parameter Javadoc to the method Javadoc
+    builder.javadoc.add(builder.parameterJavadoc.build());
```
JakeWharton: I don't think this is correct. If you build a method and then call toBuilder(), add javadoc, and add a parameter you'll get a split set of `@param`. We need to track the Javadoc tags separately from the original Javadoc.

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -371,16 +374,24 @@ public Builder returns(Type returnType) {
...
 
     public Builder addParameter(ParameterSpec parameterSpec) {
+      addParameterJavadoc(parameterSpec);
       this.parameters.add(parameterSpec);
       return this;
     }
 
+    private void addParameterJavadoc(ParameterSpec parameterSpec) {
+      if (!parameterSpec.javadoc.isEmpty()) {
+        this.parameterJavadoc.add("@param $L $L", parameterSpec.name, parameterSpec.javadoc);
```
JakeWharton: Is there any reason to do this eagerly? Why not aggregate when we emit and avoid all this code?

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -140,6 +140,16 @@ void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitMod
     codeWriter.popTypeVariables(typeVariables);
   }
 
+  private CodeBlock javadocWithParameters() {
+    CodeBlock.Builder builder = javadoc.toBuilder();
+    for (ParameterSpec parameterSpec : parameters) {
+      if (!parameterSpec.javadoc.isEmpty()) {
+        builder.add("@param $L $L", parameterSpec.name, parameterSpec.javadoc);
```
JakeWharton: I'd really like a newline before these tags but it's incredibly annoying to do in Java and 1 line of code in Kotlin. How about
```java
boolean emitTagNewline = true;
for (ParameterSpec parameter : parameters) {
  if (!parameter.javadoc.isEmpty()) {
    if (emitTagNewline) builder.add("\n");
    emitTagNewline = false;
    builder.add("...");
  }
}
```

tbroyer: How about

```java
return javadoc.toBuilder().add(
    parameters
        .stream()
        .filter(p -> !p.javadoc.isEmpty())
        .map(p -> CodeBlock.of("@param $L $L", p.name, p.javadoc))
        .collect(CodeBlock.joining("", "\n", "")));
```
It will create (and merge) many intermediate `CodeBlock` but that's probably an acceptable tradeoff (we don't check whether any `ParameterSpec` has a non-empty javadoc either before creating the builder, therefore making an unnecessary copy of the javadoc, so :man_shrugging:)

But how about `@return`, `@throws`, `@see`, etc. (that, in addition, [generally](https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#orderoftags) come *after* the `@param`s)? If you put them in the method javadoc, there'd be a blank line between them and `@param`s, unless you remember **not** to end your method javadoc with a `\n` in presence of `ParameterSpec`-level javadoc.
Maybe we should add `returnTypeJavadoc` et al. at the same time as `ParameterSpec.javadoc`?

:can_of_worms:

JakeWharton: That was close to what my Kotlin line was going to be!

I was going to do the other tags in a follow-up PR.

shaishavgandhi05: Ended up going with the uglier approach. The stream approach would add the extra line even when there were no parameter javadocs since the `collect` was still called.

JakeWharton: I find the stream to be the uglier approach!

shaishavgandhi05:
Just noticed that the current code would emit a new line even if there was no method javadoc. This would emit something like this
```java
/**
 *
 * @param name this is a name
 */
```
instead of 
```java
/**
 * @param name this is a name
 */
```

Have added a test case that covers that.

shaishavgandhi05:
Any changes required here?

swankjesse:
This is brilliant. Thanks!

PR #677, opened 2018-09-13
[WIP] JUnit 5

sormuras:
_in progress..._

- [x] Use JUnit Platform at test runtime
- [x] Migrate single test to use JUnit Jupiter API
- [ ] Don't over-use the Jupiter API, tune it down a bit

Related to #603 and https://github.com/junit-team/junit5/issues/1582

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,6 +23,7 @@
     <java.version>1.8</java.version>
     <junit.version>4.12</junit.version>
     <!-- Upgrade to JUnit 5.3.1 (Jupiter and Vintage) blocked by https://github.com/apache/maven-surefire/pull/193 -->
+    <junit.jupiter.version>5.2.0</junit.jupiter.version>
     <junit.vintage.version>5.2.0</junit.vintage.version>
     <maven.surefire.version>2.22.0</maven.surefire.version>
     <truth.version>0.39</truth.version>
@@ -77,6 +78,12 @@
       <artifactId>junit-vintage-engine</artifactId>
       <version>${junit.vintage.version}</version>
       <scope>runtime</scope> <!-- test-runtime -->
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId> <!-- junit-jupiter-api, Surefire is not smart enough, yet. -->
+      <version>${junit.jupiter.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.google.jimfs</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,9 @@
 
     <java.version>1.8</java.version>
     <junit.version>4.12</junit.version>
+    <!-- Upgrade to JUnit 5.3.1 (Jupiter and Vintage) blocked by https://github.com/apache/maven-surefire/pull/193 -->
+    <junit.vintage.version>5.2.0</junit.vintage.version>
+    <maven.surefire.version>2.22.0</maven.surefire.version>
     <truth.version>0.39</truth.version>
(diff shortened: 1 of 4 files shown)


JakeWharton:
JavaPoet is imported as source into Google's monorepo so I'll have to check as to whether JUnit 5 is supported or not. Or maybe @ronshapiro knows immediately off-hand (I'm inclined to think no).

My favorite quote from the linked issue is:

> they are two separate frameworks

JUnit 5 was named incorrectly. This is an entirely different framework that should be vetted on its own merits compared to JUnit 4. It is decidedly **not** an upgrade.

So we should evaluate it against other new frameworks and also against doing nothing.

ronshapiro:
It's not, but we also don't run JavaPoet's tests (for a silly reason, I think at the time none of the tests had `@RunWith` and our runner didn't support detecting them (over JUnit3 tests)).

But please don't make a decision based on Google's eng structure (especially for tests!)

swankjesse:
Yep. Because JUnit 4 and JUnit 5 don’t interoperate nicely I think we’re better off waiting until all of our dependent projects have upgraded. In particular, we should wait for compile-testing and Mockito to support JUnit 5.

src/test/java/com/squareup/javapoet/UtilTest.java
```diff
@@ -15,61 +15,87 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.DynamicContainer.dynamicContainer;
+import static org.junit.jupiter.api.DynamicTest.dynamicTest;
 
-import org.junit.Test;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.DynamicNode;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestFactory;
```
swankjesse: Going from 2 imports to 9 makes me want to back away from Jupiter forever.

pom.xml
```diff
@@ -69,6 +73,18 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <version>${junit.vintage.version}</version>
+      <scope>runtime</scope> <!-- test-runtime -->
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId> <!-- junit-jupiter-api, Surefire is not smart enough, yet. -->
```
swankjesse: Blech. They say one of Jupiter’s benefits is better integration with test runners. But I see that it isn’t integrated with Maven yet. That doesn’t inspire confidence.

sormuras:
Mockito 2.22.0 already provides a `MockitoExtension`: https://static.javadoc.io/org.mockito/mockito-junit-jupiter/2.22.0/org/mockito/junit/jupiter/MockitoExtension.html

src/test/java/com/squareup/javapoet/UtilTest.java
sormuras: +7 more imports to go from

image

to

image

Better reporting needs some imports.

But as I stated in the original description text, I'll rework this commit to only show minimal changes.

pom.xml
sormuras: > Blech.

Hey, I understand that. (-:

Come over and help improving Surefire!

Or try https://github.com/sormuras/junit-platform-maven-plugin which already "is smart enough" to add required runtime dependencies.

sormuras: Surefire 2.22.2 or the upcoming 3.0.0-M4 do help in this configuration ... "Blech". I'll upgrade the PR ... soon.

PR #679, opened 2018-09-14
support add imports

chengweidev:


JakeWharton:
This should never be needed when using $T

sormuras:
Duplicate of #512

tbroyer:
@chengweidev I quickly looked at X2C's code, it looks like you could _quite easily_ replace your `StringBuffer` (sic! why not `StringBuilder`?) with a `CodeBlock.Builder` and remove your imports handling entirely so you wouldn't need that change.

chengweidev:
@tbroyer Thank you for pointing out that using StringBuffer is a mistake, which should be using StringBuilder, but replacing StringBuffer with CodeBlock will complicate the code, for example,
 ```java
 stringBuffer. append (String. format ("% s% s = new% s (ctx); n", mName, obj, mName));
 ```
 requiring codeBlock to nest codeBlock

chengweidev:
@sormuras   Yes, I have the same problem with him, and I think JavaFile provides an import method that makes it easier to use

tbroyer:
```java
builder.addStatement("$T $N = new $T(ctx)", mName, obj, mName);
```
(You could also use `$1T` and avoid passing `mName` twice)

chengweidev:
@tbroyer  Very powerful. Thank you very much.

PR #681, opened 2018-10-07
Add checks to ParameterSpec with VariableElement + copy over annotations

### LINKED ISSUE #665: ParameterSpec miscellany
I noticed a few things today when looking at the `ParameterSpec` source:
  - `ParameterSpec.get(VariableElement)` doesn't check that the element kind is a parameter
  - `ParameterSpec.get(VariableElement)` doesn't copy annotations from the element
  - The only valid modifier on a parameter is `final`, but JavaPoet will accept any. This isn't inconsistent with the rest of the APIs, but I wonder if it's a good place to fail fast instead of generating code that will surely not compile. I noticed this when thinking that `ParameterSpec.getModifiers()` returns a `Set<Modifier>` - it is consistent with the other `*Spec` classes, but may make someone think twice about what types of modifiers might go there.

Happy to send PRs for any of the above. I presume the second is surely a bug; not sure if the first is Working As Intended or not.

ShaishavGandhi:
Resolves #665 

I'd love to add some tests around the `get(variableElement)` method but I'm not sure how to do that with creating a `VariableElement`. If you have any suggestions, let me know!

### DIFF
--- a/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
@@ -15,16 +15,69 @@
  */
 package com.squareup.javapoet;
 
+import com.google.testing.compile.CompilationRule;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import javax.annotation.Nullable;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
+import static javax.lang.model.util.ElementFilter.fieldsIn;
+import static javax.lang.model.util.ElementFilter.methodsIn;
 import static org.junit.Assert.fail;
 
 import javax.lang.model.element.Modifier;
 
 public class ParameterSpecTest {
+  @Rule public final CompilationRule compilation = new CompilationRule();
+
+  private Elements elements;
+
+  @Before public void setUp() {
+    elements = compilation.getElements();
+  }
+
+  private TypeElement getElement(Class<?> clazz) {
+    return elements.getTypeElement(clazz.getCanonicalName());
+  }
+
+  private VariableElement findFirst(Collection<VariableElement> elements, String name) {
+    for (VariableElement variableElement : elements) {
+      if (variableElement.getSimpleName().toString().equals(name)) {
+        return variableElement;
+      }
+    }
+    throw new IllegalArgumentException(name + " not found in " + elements);
+  }
+
+  private ExecutableElement findFirstExecutable(Collection<ExecutableElement> elements, String name) {
+    for (ExecutableElement executableElement : elements) {
+      if (executableElement.getSimpleName().toString().equals(name)) {
+        return executableElement;
+      }
+    }
+    throw new IllegalArgumentException(name + " not found in " + elements);
+  }
+
+  final class VariableElementFieldClass {
+    String name;
+  }
+
+  final class VariableElementParameterClass {
+    public void foo(@Nullable String bar) {
+    }
+  }
+
+
   @Test public void equalsAndHashCode() {
     ParameterSpec a = ParameterSpec.builder(int.class, "foo").build();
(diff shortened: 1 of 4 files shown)


shaishavgandhi05:
Ping!

shaishavgandhi05:
Thanks! I'm gonna try and get to this soon!

shaishavgandhi05:
I've updated with tests! Thanks for the help

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -48,6 +101,28 @@
...
+  }
+
+  @Test public void parameterVariableElement() {
+    TypeElement classElement = getElement(VariableElementParameterClass.class);
+    List<ExecutableElement> methods = methodsIn(elements.getAllMembers(classElement));
+    ExecutableElement element = findFirstExecutable(methods, "foo");
+    VariableElement parameterElement = element.getParameters().get(0);
+
+    assertThat(ParameterSpec.get(parameterElement).toString())
+        .isEqualTo("@javax.annotation.Nullable java.lang.String arg0");
```
shaishavgandhi05: This seems like a bug to me. This is unrelated to my changes but in `ParameterSpec`, we get the name of the element with `element.getSimpleName().toString()`. This yields `argo0` instead of `bar`. I'm not sure if this is a consequence of my test or this is the normal behavior (which seems like a bug)

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -15,16 +15,69 @@
...
+
+  @Before public void setUp() {
+    elements = compilation.getElements();
+  }
+
+  private TypeElement getElement(Class<?> clazz) {
+    return elements.getTypeElement(clazz.getCanonicalName());
+  }
+
+  private VariableElement findFirst(Collection<VariableElement> elements, String name) {
```
Egorand: Out of scope for this PR, but it'd be great to move those helper functions into a Util class

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -15,16 +15,69 @@
...
+  private ExecutableElement findFirstExecutable(Collection<ExecutableElement> elements, String name) {
+    for (ExecutableElement executableElement : elements) {
+      if (executableElement.getSimpleName().toString().equals(name)) {
+        return executableElement;
+      }
+    }
+    throw new IllegalArgumentException(name + " not found in " + elements);
+  }
+
+  final class VariableElementFieldClass {
```
Egorand: nit: Maybe move these declarations closer to the tests that need them

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -48,6 +101,28 @@
     }
   }
 
+  @Test public void fieldVariableElement() {
+    TypeElement classElement = getElement(VariableElementFieldClass.class);
+    List<VariableElement> methods = fieldsIn(elements.getAllMembers(classElement));
+    VariableElement element = findFirst(methods, "name");
+
+    try {
+      ParameterSpec.get(element);
```
Egorand: This tests will pass even if `get()` doesn't throw. This should be followed by `fail()`.

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
Egorand: Can you repro this on master? If yes - definitely worth filing.

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
shaishavgandhi05: I don't mind doing it in this PR!

src/test/java/com/squareup/javapoet/ParameterSpecTest.java
shaishavgandhi05: Can reproduce on master. Will file a bug.

shaishavgandhi05:
Let me know if `TestUtil` sounds good for the util methods. Didn't name it `Util` since there's already one in `src/main`

shaishavgandhi05:
Oops. Missed that. Will change it to make more generic.

shaishavgandhi05:
Anything that I can address here?

Egorand:
Nope, good to merge - will do on Monday.

MatushkaRossiya:
It would be funnier if you actually merged this on Monday. :)

PR #687, opened 2018-10-23
Public getter for canonical name of ClassName

int02h:
Canonical name may be useful when creating `JavaFileObject` for code generation in APT Processor instead of manually concatenating `packageName()` and `simpleName()`

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -138,6 +138,14 @@ public final class ClassName extends TypeName implements Comparable<ClassName> {
     return simpleName;
   }
 
+  /**
+   * Returns the full class name of this class, like {@code "java.util.Map.Entry"}
+   * for {@link Map.Entry}.
+   * */
+  public String canonicalName() {
+    return canonicalName;
+  }
+
   public static ClassName get(Class<?> clazz) {
     checkNotNull(clazz, "clazz == null");
     checkArgument(!clazz.isPrimitive(), "primitive types cannot be represented as a ClassName");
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/ClassNameTest.java
```diff
@@ -193,4 +193,14 @@ public void reflectionName() {
     assertEquals("Foo$Bar$Baz", ClassName.get("", "Foo", "Bar", "Baz").reflectionName());
     assertEquals("a.b.c.Foo$Bar$Baz", ClassName.get("a.b.c", "Foo", "Bar", "Baz").reflectionName());
   }
+
+  @Test
+  public void canonicalName() {
+    assertEquals("java.lang.Object", TypeName.OBJECT.canonicalName());
```
Egorand: Nice tests!

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -138,6 +138,14 @@ public String simpleName() {
     return simpleName;
   }
 
+  /**
+   * Returns the full class name of this class, like {@code "java.util.Map.Entry"}
```
JakeWharton: Can we put a period between the description and the example? This will prevent the example from showing in the short Javadoc in the method table and the IDE unless you click/expand it for more.

int02h: I tried to make this JavaDoc looks like JavaDoc for `simpleName()` where example goes right after the comma.

JakeWharton: We should fix that one too! I can do it separately.

src/test/java/com/squareup/javapoet/ClassNameTest.java
int02h: Is it some kind of approve or necessary to do something?

Issue #698, opened 2018-12-19
$C for Characters?


bjansen:
Hi,

I was looking for something similar to `$S` but for character literals, and could not find any easy way to do it. 

Basically what I'm trying to achive is something like this:

```java
Character someChar = ...;

builder.addStatement("myArray['$L'] = ...", someChar);
```

This doesn't work when `someChar` needs to be escaped (`\`, `'`, etc.) because it produces code like `myArray['''] = ...`.

It looks like JavaPoet has everything needed to implement that (`com.squareup.javapoet.Util#characterLiteralWithoutSingleQuotes`), it just needs to handle `$C` and wrap the value with single quotes.

If you are interested in this feature, I can implement it and submit a PR.

Thanks for this nice library btw :)

galderz:
Hi @bjansen, your proposal makes sense to me as well. I've just sent an initial PR (#700).

It's worth noting though that there's a workaround for this issue. You'd have to create your own copy of `Util#characterLiteralWithoutSingleQuotes` (since that method is not public), and pass your characters through. The output would then be replaced by `$L`.

PR #700, opened 2019-01-28
Initial proposal for $C to enhance character literal usage #698

galderz:
This is an initial proposal to provide an easier way to deal with character literals.

The majority of characters can currently be dealt with using `'$L'`. However, special characters such as `'` or `\` (amongst others), would require special handling by the users.

In essence, a user would need to define a method such as `com.squareup.javapoet.Util.characterLiteralWithoutSingleQuotes()` and pass characters through that method to get it printed as expected.

Rather than putting the burden on the user, @bjansen has proposed in issue #698 to add support for single-quoted character literals via `$C`. I also think this is a good enhancement.

As an initial proposal, this PR should not be considered as the final product. Here's a list of pending tasks: 

- [ ] Documentation
- [ ] Further test cases should be added to verify its robustness. 
- [ ] `singleQuoteCharacterInLiteral` should probably be removed since it's likely covered already by other tests. I just added this test to compare existing alternatives for handling special characters.
- [ ] Convert `AnnotationSpec.Builder.addMemberForValue` to use `$C` instead of ``$L``


src/test/java/com/squareup/javapoet/CodeBlockTest.java
```diff
@@ -339,4 +339,27 @@
...
+    // Util method not public, so user would need to create its own version
+    CodeBlock utilInBlock = CodeBlock.builder().add("'$L'", Util.characterLiteralWithoutSingleQuotes(ch2)).build();
+    assertThat(utilInBlock.toString()).isEqualTo("'\\''");
+  }
+
+  @Test public void singleQuoteCharacterInCharacter() {
+    char c = '\'';
+    CodeBlock block = CodeBlock.builder().add("$C", c).build();
+    assertThat(block.toString()).isEqualTo("'\\''");
+  }
```
kenzierocks: Perhaps add tests for other escaped literals?

src/main/java/com/squareup/javapoet/CodeBlock.java
```diff
@@ -360,6 +363,10 @@ private TypeName argToType(Object o) {
       throw new IllegalArgumentException("expected type but was " + o);
     }
 
+    private Character argToChar(Object o) {
+      return o != null ? ((char) o) : null;
```
kenzierocks: (Apparently GitHub lost this one)

Check `instanceof Character` and throw `IllegalArgumentException` like above?

src/test/java/com/squareup/javapoet/CodeBlockTest.java
galderz: Yeah, as mentioned in the description, I was planning to add more tests :). I've just added some more tests for other escaped character literals and squashed the commits.

src/main/java/com/squareup/javapoet/CodeBlock.java
galderz: Good point, should be fixed now.

Issue #691, opened 2018-11-08
JavaFile.writeTo(Path) should return output Path


koritakoa:
`JavaFile.writeTo(Path)` does some work to calculate the actual Path to which the source is written. There is no way to find what that output Path is without duplicating the logic. `writeTo(Path)` should return the actual output `Path` so we don't have to. `writeTo(File)` should return a `File`.

If you don't want to change the method signature, please consider adding a comment to these methods that the signature should change in 2.x and add a separate method to return the output path taking a Path parameter.

JakeWharton:
Makes sense.

galderz:
Indeed, changing the method signature would break binary compatibility since changing return type has the effect of removing the old method and adding the new one.

I'm not sure though about adding a separate method that returns the output path. First, I couldn't think of a good name for such method and second, it felt like you're trying to find a way to define an subset of an operation that is meant to stay as internal detail.

Instead, I'd like to suggest adding two methods that would keep the API relatively consistent with the downside of slightly bloating the method names:

```
/** Writes this to {@code directory} as UTF-8 using the standard directory structure.
   * Returns the {@link Path} instance to which source is actually written. */
public Path writeToPath(Path directory) throws IOException { ... }
```

```
/** Writes this to {@code directory} as UTF-8 using the standard directory structure.
 * Returns the {@link File} instance to which source is actually written. */
public File writeToFile(File directory) throws IOException { ... }
```

I've created a PR (#701) with this proposal (note: no tests added yet).

PR #701, opened 2019-01-29
Add writeTo methods returning Path/File written to #691

galderz:
Proposal for new `writeToPath` and `writeToFile` methods to be added to JavaFile that return Path/File instances to which the sources are actually written to.

I've not yet added tests since this suggestion has not yet been approved. If approved I'll add tests.

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -83,6 +83,12 @@ public final class JavaFile {
 
   /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
   public void writeTo(Path directory) throws IOException {
+    writeToPath(directory);
+  }
+
+  /** Writes this to {@code directory} as UTF-8 using the standard directory structure.
+   * Returns the {@link Path} instance to which source is actually written. */
+  public Path writeToPath(Path directory) throws IOException {
     checkArgument(Files.notExists(directory) || Files.isDirectory(directory),
         "path %s exists but is not a directory.", directory);
     Path outputDirectory = directory;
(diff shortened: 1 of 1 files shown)


src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -83,6 +83,12 @@ public void writeTo(Appendable out) throws IOException {
 
   /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
   public void writeTo(Path directory) throws IOException {
+    writeToPath(directory);
+  }
+
+  /** Writes this to {@code directory} as UTF-8 using the standard directory structure.
```
Egorand: nit: New line after `/**`

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -83,6 +83,12 @@ public void writeTo(Appendable out) throws IOException {
...
+   * Returns the {@link Path} instance to which source is actually written. */
```
Egorand: nit: New line before the closing `*/`

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -97,13 +103,22 @@ public void writeTo(Path directory) throws IOException {
...
+
+    return outputPath;
   }
 
   /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
   public void writeTo(File directory) throws IOException {
     writeTo(directory.toPath());
   }
 
+  /** Writes this to {@code directory} as UTF-8 using the standard directory structure.
```
Egorand: nit: Javadoc formatting, same as above

Egorand:
Fixed conflicts and merged manually. Thanks @galderz!

PR #702, opened 2019-01-30
Allow setting method name on MethodSpec.Builder

### LINKED ISSUE #699: Consider MethodSpec toBuilder(String name) overload
(This is somewhat related to some other issues, but I see this as a much smaller scope, additionally this is just a "would be nice to have" feature, I can workaround this pretty easily)

If I call toBuilder() on a MethodSpec, the name of the method cannot be changed in the resulted Builder (I did not find any way at least, sorry if I missed. Based on the code it is final anyway at https://github.com/square/javapoet/blob/master/src/main/java/com/squareup/javapoet/MethodSpec.java). 

It would be nice if I could pass the name of the method to the builder as an overload toBuilder(String methodName). I'm aware that only additions are possible through the builder, however the name is a mandatory parameter (since you need it for MethodSpec.methodBuilder(name) always), so it cannot be changed. 

My use-case would be to generate a set of methods, with some being "XXXAsync", which are mostly identical (return type and final statement are different), but the name is different.

Might need some thinking for constructors, however in another usecase, one of my methods would be identical to a (partially constructed) constructor MethodSpec. Therefore it would be sort-of nice if I could transform a method MethodSpec to a constructor MethodSpec builder, and vice versa (the code seems to use `static final String CONSTRUCTOR = "<init>"; ` for the name, could it be possible to expose this constant as public?). And as extra mention, it would be nice to have a single method to construct MethodSpecs, and if passed that <init> it would be a constructor (e.g. let's say MethodSpec.rawBuilder or something).

shaishavgandhi05:
Resolves #699

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -278,7 +278,7 @@ public final class MethodSpec {
   }
 
   public static final class Builder {
-    private final String name;
+    private String name;
 
     private final CodeBlock.Builder javadoc = CodeBlock.builder();
     private final List<AnnotationSpec> annotations = new ArrayList<>();
@@ -292,11 +292,16 @@ public final class MethodSpec {
     private CodeBlock defaultValue;
 
     private Builder(String name) {
+      setName(name);
+    }
+
+    public Builder setName(String name) {
       checkNotNull(name, "name == null");
       checkArgument(name.equals(CONSTRUCTOR) || SourceVersion.isName(name),
-          "not a valid name: %s", name);
+              "not a valid name: %s", name);
       this.name = name;
       this.returnType = name.equals(CONSTRUCTOR) ? null : TypeName.VOID;
+      return this;
     }
 
     public Builder addJavadoc(String format, Object... args) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -292,11 +292,16 @@ public Builder toBuilder() {
...
 
     private Builder(String name) {
+      setName(name);
+    }
+
+    public Builder setName(String name) {
       checkNotNull(name, "name == null");
       checkArgument(name.equals(CONSTRUCTOR) || SourceVersion.isName(name),
-          "not a valid name: %s", name);
+              "not a valid name: %s", name);
```
Egorand: Could you please revert this?

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -302,4 +303,16 @@ static void staticMethod() {
       assertThat(e.getMessage()).isEqualTo("modifiers == null");
     }
   }
+
+  @Test public void modifyMethodName() {
+    MethodSpec methodSpec = MethodSpec.methodBuilder("initialMethod")
+            .build()
```
Egorand: Continuation indents should be 4 spaces

src/main/java/com/squareup/javapoet/MethodSpec.java
shaishavgandhi05: Oops. Fixed fe3f339c168ddca267c2e6ae0aad095b455d8c5f

src/test/java/com/squareup/javapoet/MethodSpecTest.java
shaishavgandhi05: Fixed fe3f339c168ddca267c2e6ae0aad095b455d8c5f

Issue #693, opened 2018-11-26
Update documentation to include nextControlFlow


davidnewcomb:
The [README.md](https://github.com/square/javapoet/blob/master/README.md) has a section on flow control but it completely misses out everything related to exceptions.

The answer in #692 was less than helpful as the `README.md` does not even mention `nextControlFlow`.

Please can you update the documentation to include information about how to `catch` exceptions and while we're at it - let's go totally bonkers crazy and include the `finally` block in the example too ;)

swankjesse:
Pull request please!

davidnewcomb:
Well if I knew how it worked, I'd be more than happy to! But then I probably wouldn't need the documentation as much. Do you see my problem?

tbroyer:
Well, if you go read the [javadoc for `beginControlFlow`](https://square.github.io/javapoet/1.x/javapoet/com/squareup/javapoet/MethodSpec.Builder.html#beginControlFlow-java.lang.String-java.lang.Object...-), `newControlFlow` is just underneath. So:
```java
.beginControlFlow("try")
.addStatement("return (" + name + ") classDef.getConstructor($T.class).newInstance(entry)",
MyObj.class)
.endControlFlow("catch ($T e)", Exception.class)
.addStatement("throw new $T(e)"", RuntimeException.class)
.endControlFlow()
```

(btw, this `+ name +` should probably be replaced with `$T`)

davidnewcomb:
I think you are labelling my point. The documentation for `beginControlFlow` and `newControlFlow` is exactly the same (not very helpful unless they do the same thing, in which case what's the point of one of them).
Your example doesn't use `newControlFlow` so either the other person didn't understand the documentation either or it's a typo. I'm not really sure, if only we had something we could refer to!

Thanks for the tip re $T, I think I have updated that bit already.

The documentation in the README is generally excellent and complete except for any mention of catching exceptions and this new-flow thing. On this occasion Google wasn't even much help because all the search terms I can use bring back so much unrelated stuff, which is why I ended up here.

galderz:
Learning about javapoet, so gave `nextControlFlow` a go and sent a PR to add docu (#703)

Egorand:
Closed by #703

PR #703, opened 2019-01-30
Add documentation on nextControlFlow to README #693

galderz:


galderz:
Updated commit to add generated code.

PR #705, opened 2019-02-05
Remove n^2 algorithm in CodeWriter.resolve() by precomputing all of the nested simple names of a TypeSpec

ronshapiro:
For one large (100K lines) Dagger component implementation, this saved 3.5s/build. The type has ~360 subcomponents, and each of those have a sibling builder type, so ~720 nested types at the worst point. 

I wasn't sure what your style preferences are for whether `nestedTypesSimpleNames` should be immutable or not given that it's internal, so I implemented it as mutable and figured I'd ask your thoughts. I can also call `Util.immutableSet()` or create a single mutable set that never escaped and wrap that in an `unmodifiableSet`.


ronshapiro:
Is this ready? The Kotlinpoet version was already merged

Egorand:
Was waiting for somebody else to merge, gonna merge myself

Issue #564, opened 2017-06-08
Provide a simple way to generate Java source code with a specific encoding


saw303:
Currently `com.squareup.javapoet.JavaFile` provides the following method to write the Java source code to the file system.

- `writeTo(Appendable)`
- `writeTo(Path)`
- `writeTo(File)`
- `writeTo(Filer)`

The methods `writeTo(Path)` and `writeTo(File)` provide a very nice way to write code to a standard directory. The do 

- create all directories (according to the package name)
- and create the Java file itself in the correct directory structure

The downside of this convenient method is that they choose the default encoding of the current Java virtual machine when writing the Java source code into the target file.

In our build we'd like to make sure all our source code is `UTF-8` encoded. Therefore it would be very handy to have convenience methods such as:

- `writeTo(Path, Charset)`
- `writeTo(File, Charset)`

sormuras:
Since [1.8.0](https://github.com/square/javapoet/blob/master/CHANGELOG.md#javapoet-180-2016-11-09) JavaPoet always uses UTF-8:
_Fix: Always write UTF-8. Previously JavaPoet would use the system default charset which was potentially inconsistent across environments._

tbroyer:
Dupe of #473

saw303:
Good to hear that JavaPoet uses UTF-8 in general. 

But this issue asks about more flexibility. I as programmer would like to tell JavaPoet what encoding to use. There are projects that are not using UTF-8 as source code encoding. In these projects I cannot use JavaPoet at the current state.

If you do not like to change the API you could introduce an SPI using Java's [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) to provide a way to change the default encoding.

ronshapiro:
JavaFile.toString() gives you the file contents as a string, then you could
call getBytes yourself with whatever charset you want

saw303:
I agree. That does the job.

PR #712, opened 2019-03-05
easier_way_provide_encoding

fischey:
provide an easier using an encoding while writing a java file to a directory.

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -87,7 +87,10 @@ public final class JavaFile {
     writeTo(directory, UTF_8);
   }
 
-  /** Writes this to {@code directory} with the provided {@code charset} using the standard directory structure. */
+  /**
+   * Writes this to {@code directory} with the provided {@code charset}
+   * using the standard directory structure.
+   */
   public void writeTo(Path directory, Charset charset) throws IOException {
     checkArgument(Files.notExists(directory) || Files.isDirectory(directory),
         "path %s exists but is not a directory.", directory);
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -87,8 +87,8 @@ public final class JavaFile {
     writeTo(directory, UTF_8);
   }
 
-  /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
-  public void writeTo(Path directory, final Charset charset) throws IOException {
+  /** Writes this to {@code directory} with the provided {@code charset} using the standard directory structure. */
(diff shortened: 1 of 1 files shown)


fischey:
Might be a fix for this issue: https://github.com/square/javapoet/issues/564.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -83,6 +84,11 @@ public void writeTo(Appendable out) throws IOException {
 
   /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
   public void writeTo(Path directory) throws IOException {
+    writeTo(directory, UTF_8);
+  }
+
+  /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
```
tbroyer: Javadoc needs fixing as this is no longer utf8 only.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -83,6 +84,11 @@ public void writeTo(Appendable out) throws IOException {
...
+  public void writeTo(Path directory, final Charset charset) throws IOException {
```
JakeWharton:
```suggestion
  public void writeTo(Path directory, Charset charset) throws IOException {
```

src/main/java/com/squareup/javapoet/JavaFile.java
fischey: Thank you for the hint! :-)

src/main/java/com/squareup/javapoet/JavaFile.java
fischey: Thank you for the hint. Its done.

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -83,6 +84,11 @@ public void writeTo(Appendable out) throws IOException {
 
   /** Writes this to {@code directory} as UTF-8 using the standard directory structure. */
   public void writeTo(Path directory) throws IOException {
+    writeTo(directory, UTF_8);
+  }
+
+  /** Writes this to {@code directory} with the provided {@code charset} using the standard directory structure. */
```
Egorand: Looks like checkstyle is complaining that this line is longer than the limit, which is 100 lines. Can you please fix?

fischey: I corrected this. Sorry that I not see this in the first place.

Egorand: Thanks for fixing!

fischey:
Thank you for accepting the change :-)
I signed your CLA. So you can proceed. Best regards.

PR #721, opened 2019-04-17
Create a OSGi-Bundle using nd-maven-plugin

stbischof:


Egorand:
Could you please explain why this is needed? Generally it's better to open an issue describing the problem and then submit a PR that closes the issue.

stbischof:
I would like to use this jar in an OSGi Environment. The Manifest file needs some information which could be generated by the bnd plugin

Egorand:
Thanks for you PR, but I don't think we want to merge this. If your use-case depends on this configuration, I'd suggest forking and self-publishing JavaPoet with necessary POM modifications.

PR #723, opened 2019-04-26
Fix an issue where ClassName could not handle classes in the default (empty) package

agavra:
A product that I'm working on utilizes javapoet to generate code from user-defined-functions. As an example, we have built a simple jar file that contains a class that is in the default java package (i.e. empty package). If we use this class and try to format it using 
```java
CodeBlock.builder()
  .addStatement("$T", ThatClass.class)
  .build()
  .toString()
```
It fails with an NPE in the most recent version of javapoet.

This PR attempts to fix it by adding null checks on any access to the package name.

__NOTE:__ I added the test in the default package instead of in `ClassNameTest` so that I could reference the class without mocking it. If anyone has a better idea I'm all ears!

P.S. Thanks for this library :) it's helped me clean up legacy code in multiple different companies!

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
@@ -32,7 +33,7 @@ import static com.squareup.javapoet.Util.checkNotNull;
 public final class ClassName extends TypeName implements Comparable<ClassName> {
   public static final ClassName OBJECT = ClassName.get(Object.class);
 
-  /** The name representing the default Java package */
+  /** The name representing the default Java package. */
   private static final String NO_PACKAGE = "";
 
   /** The package name of this class, or "" if this is in the default package. */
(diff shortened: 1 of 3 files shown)


kenzierocks:
I think it would be better normalize the package name on construction, rather than check it for `null`, since it seems to already be the standard to have the default as `""`.

JakeWharton:
Agree

agavra:
Thanks for the reviews @kenzierocks @JakeWharton! I agree that's certainly a better suggestion. I updated to code to do this, including updating all call-sites to be extra clear.

src/main/java/com/squareup/javapoet/ClassName.java
```diff
@@ -53,12 +56,12 @@ private ClassName(String packageName, ClassName enclosingClassName, String simpl
   private ClassName(String packageName, ClassName enclosingClassName, String simpleName,
       List<AnnotationSpec> annotations) {
     super(annotations);
-    this.packageName = packageName;
+    this.packageName = (packageName != null) ? packageName : NO_PACKAGE;
```
JakeWharton: Is this still required? Seems like we might want to prefer exploding instead
```suggestion
    this.packageName = Objects.requireNonNull(packageName, "packageName == null")
```

agavra: Technically not required - I just wanted to be defensive for the future (I tracked down all first level call-sites, but it's hard to tell)

agavra:
Thanks again @JakeWharton, anything else that needs to be done before merging?

JakeWharton:
Squashed and merged as 3829f2ca6f03a4b941fea41a1e2f4eead8f37bc1. Thanks!

PR #732, opened 2019-08-07
Ensure trailing newlines in javadocs and method bodies

### LINKED ISSUE #722: Add trailing newline to methods if necessary
When `MethodSpec.addCode()` is used as opposed to `addStatement()`, it's often possible to end up with methods that have no trailing newline and thus look like:

```java
void foo() {
  doThing();
  doThing2();}
```

This can easily happen when the body of a method is composed as a `List<CodeBlock>` and then passed to `CodeBlock.join("\n", statements)` (and happens within Dagger often. It's coming up now that we've turned off formatting by default).

This is easily avoidable by using `CodeBlock.join("", statements)` and ensuring that each statement uses `CodeBlock.builder().addStatement().build()` instead of `CodeBlock.of()`, or by using `CodeBlock.join("", "\n", "\n", statements) to add a suffix newline. But both of these are awkward and easy to forget.

I think it should be relatively easy, with some possible hackery in `LineWrapper` to retain whether the last character is a `;` and to propagate that information so that `MethodSpec.emit()` can emit an extra newline if necessary.

### LINKED ISSUE #731: Should javadocs include a trailing newline if one isn't present?
Right now, if I add a one line doc comment, I get something like this

```java
/**
 * This type + annotation exists for sharing information to the Crumb annotation processor and should not be considered public API. */
final class Lib1ProducerCrumbIndex {
  private Lib1ProducerCrumbIndex() {
  }
}
```

I think it should just always add a trailing newline for style reasons, but want to check before opening a PR if there was a reason for this. Kotlinpoet appears to always add the trailing newline for the equivalent kdoc

```java
/**
 * This type + annotation exists for sharing information to the Crumb annotation processor and should not be considered public API.
 */
final class Lib1ProducerCrumbIndex {
  private Lib1ProducerCrumbIndex() {
  }
}
```

ZacSweers:
This introduces functionality that tracks the last char emitted, and uses it to check if a trailing newline needs to be added after emitting a codeblock for javadocs and method bodies.

Resolves #731
Resolves #722

I tried to keep this as performant and lightweight as possible. If this implementation looks good, I'll do the same for KotlinPoet.

### DIFF
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -2440,4 +2440,19 @@ public final class TypeSpecTest {
         + "class Taco {\n"
         + "}\n");
   }
+
+  @Test public void javadocEnsuresTrailingLine() {
+    TypeSpec spec = TypeSpec.classBuilder("Taco")
+        .addJavadoc("Some doc with a newline")
+        .build();
+
+    assertThat(toString(spec)).isEqualTo(""
+        + "package com.squareup.tacos;\n"
+        + "\n"
+        + "/**\n"
+        + " * Some doc with a newline\n"
+        + " */\n"
+        + "class Taco {\n"
+        + "}\n");
+  }
 }
--- a/src/test/java/com/squareup/javapoet/TypeSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/TypeSpecTest.java
@@ -1848,7 +1848,8 @@ public final class TypeSpecTest {
         + "  }\n"
         + "\n"
         + "  /**\n"
-        + "   * chosen by fair dice roll ;) */\n"
+        + "   * chosen by fair dice roll ;)\n"
+        + "   */\n"
         + "  public int getRandomQuantity() {\n"
         + "    return 4;\n"
         + "  }\n"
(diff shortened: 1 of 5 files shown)


ZacSweers:
The CI failures seem unrelated

```
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

src/test/java/com/squareup/javapoet/TypeSpecTest.java
```diff
@@ -2424,4 +2425,19 @@ private String toString(TypeSpec typeSpec) {
     assertThat(TypeSpec.enumBuilder(className).addEnumConstant("A").build().name).isEqualTo("Example");
     assertThat(TypeSpec.annotationBuilder(className).build().name).isEqualTo("Example");
   }
+
+  @Test public void javadocWithTrailingLineDoesNotAddAnother() {
```
JakeWharton: What about a dedicated test for adding the newline?

ZacSweers: edaccf4

ZacSweers:
CI failure appears to be https://travis-ci.community/t/install-of-oracle-jdk-8-failing/3038. Either requires `dist: trusty` or using openjdk instead

JakeWharton:
Why do we have 400 JDKs on CI?

Egorand:
Whoops, did I break something when fixing conflicts?

ZacSweers:
@Egorand fixed

PR #733, opened 2019-08-08
Switch to openjdk8 for CI

ZacSweers:
This should hopefully resolve CI issues on PRs since oraclejdk8 isn't available without the trusty dist anymore

PR #734, opened 2019-08-09
Add alwaysQualify() API to avoid collisions with known colliding types

### LINKED ISSUE #77: Imports sometimes incorrect when inheriting nested type of same name

ZacSweers:
Implementation based on https://github.com/square/javapoet/issues/77#issuecomment-507387399

Resolves #77

Note that I'm not sure what to do about nested types like `java.util.Map.Entry`, open to suggestions!

CC @eamonnmcmanus

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -34,6 +34,8 @@ import java.util.Set;
 import java.util.TreeSet;
 import javax.annotation.processing.Filer;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.util.ElementFilter;
 import javax.tools.JavaFileObject;
 import javax.tools.JavaFileObject.Kind;
 import javax.tools.SimpleJavaFileObject;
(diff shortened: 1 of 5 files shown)


ZacSweers:
CI failure needs #733, but otherwise ran all tests pass locally on JDK 8

image

src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -258,6 +275,83 @@ public Builder addStaticImport(ClassName className, String... names) {
...
+     *
+     * <p>
+     * Then this would add {@code "NestedTypeA"} and {@code "NestedTypeB"} as names that should
+     * always be qualified via {@link #alwaysQualify(String...)}. This way they would avoid
+     * possible import conflicts when this JavaFile is written.
+     *
+     * @param typeElement the {@link TypeElement} with nested types to avoid clashes with.
+     * @return this builder instance.
+     */
+    public Builder avoidClashesWithNestedClasses(TypeElement typeElement) {
```
JakeWharton: Feels like maybe this should be an overload of declaring a superclass or interface. Otherwise when would you use these functions?

ZacSweers: That could work, how would that be communicated to the JavaFile when it's figuring out imports though?

JakeWharton: The list would be maintained in each class and then aggregated into a final list when pulling them together into a file.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -582,6 +590,83 @@ public Builder addOriginatingElement(Element originatingElement) {
...
+      return this;
+    }
+
+    /**
+     * Call this to always fully qualify any types that would conflict with possibly nested types of
+     * this {@code typeElement}. For example - if the following type was passed in as the
+     * typeElement:
+     *
+     * <pre><code>
+     *   class Foo {
```
Egorand: I think a better example is when a nested type clashes with a well-known Java type. I like the example presented in #77.

src/main/java/com/squareup/javapoet/JavaFile.java
ZacSweers: I've updated this to move the tracking of always qualified types to TypeSpecs here: 67d5094. Now the JavaFile just parses them out of all the typespecs in it. As for making them overloads of superclass/superinterface, there's no APIs for those that accept elements currently. I added them to compliment this

src/main/java/com/squareup/javapoet/TypeSpec.java
ZacSweers: Done in my added commits 👍

ZacSweers:
Updated with an overload that adds this behavior to superclass/superinterface methods, with the existing ones calling through to them with the default behavior being to avoid collisions (i.e. the new behavior). Let me know if we should keep this off by default, or if we should maybe make this another builder method like `avoidTypeCollisions(boolean enabled)`

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -483,7 +495,32 @@ public Builder superclass(TypeName superclass) {
     }
 
     public Builder superclass(Type superclass) {
-      return superclass(TypeName.get(superclass));
+      return superclass(superclass, true);
+    }
+
+    public Builder superclass(Type superclass, boolean avoidNestedTypeNameClashes) {
```
Egorand: Would we ever want to _not_ avoid nested type name clashes? Feels like this should just be the default behavior and we don't need the overloads.

ZacSweers: I think it's invasive enough of a change that it's worth an off switch. Otherwise I think it merits an RC first so some major consumers could give it a spin

Egorand: Makes sense. Then I think the default should false to ensure that the behavior of `superclass(Type)` is unchanged.

eamonnmcmanus: FWIW I'm upgrading Google's internal copy of JavaPoet at the moment and running into a number of problems because of this. I think the default _should_ have been false, but I guess it's too late now.

PR #748, opened 2019-12-22
Remove non-LTS JDK versions from Travis config

Egorand:


.travis.yml
```diff
@@ -3,13 +3,9 @@ language: java
 matrix:
   include:
     - jdk: openjdk8
-    - jdk: openjdk10
     - jdk: openjdk11
-    - jdk: openjdk-ea
   allow_failures:
-    # ErrorProne/javac is not yet working on JDK 11 nor 12 (current -ea)
```
tbroyer: Error Prone works ok with openjdk11, so the entire allow_failures could be removed (possibly after updating Error Prone)

sormuras: Hm? The CI run for this PR failed on JDK 11: https://travis-ci.org/square/javapoet/jobs/628377940#L1079

tbroyer: `plexus-compiler-javac-errorprone` does not work with JDK 10+, but ErrorProne itself works ok (through `-Xplugin: ErrorProne`)

Egorand: Thanks, I can look into addressing this in a followup.

PR #750, opened 2020-01-01
Copy originating elements in toBuilder() as well

### LINKED ISSUE #749: Originating elements are not copied in `TypeSpec#toBuilder()`

ZacSweers:
Fixes #749


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -168,6 +168,7 @@ public Builder toBuilder() {
     builder.typeSpecs.addAll(typeSpecs);
     builder.initializerBlock.add(initializerBlock);
     builder.staticBlock.add(staticBlock);
+    builder.originatingElements.addAll(originatingElements);
```
Egorand: Add a test?

ZacSweers: oops, done originatingElement

PR #755, opened 2020-01-13
Remove parameter annotations in MethodSpec.overriding

Egorand:
- Change to not copy parameter annotations was first introduced in
9505ad0e027a1f125b5352ac722ea141831fbf1c.
- Change to properly copy mirror annotations in ParameterSpec.get
was introduced in a0eadbbf0e7b70f0fbbc66043536e4328c3808fd,
breaking the behavior of MethodSpec.overriding.
- This change preserves the correct behavior of ParameterSpec.get
while also removing annotations in MethodSpec.overriding.


Egorand:
CC @ronshapiro @ZacSweers

Issue #718, opened 2019-03-25
addComment can't handle newlines in comments


JakeWharton:
```
addComment("Hey\nFriends\nI\nBreak\nCode!")
```
Can't just split on newlines either since that would break formatting.

lazyleopard5202:
public Builder addComment(String format, Object... args) {
      String[] array = format.split("\n");
      StringBuffer r = new StringBuffer();
      for (int i = 0; i < array.length; i++) {
        r.append("//");
        r.append(array[i]);
        r.append("\n");
      }
      code.add(r.toString(), args);
      return this;
    }

jdyjjj:
I want to solve this problem by using "//"

PR #759, opened 2020-01-24
Added support for multiline comments

theaveragedude:
Modified `addComment` to add multiline comments. Possible fix for #718 .

### DIFF
--- a/src/test/java/com/squareup/javapoet/MethodSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/MethodSpecTest.java
@@ -456,6 +456,44 @@ public final class MethodSpecTest {
         "}\n");
   }
 
+  @Test public void multilineComments() {
+    MethodSpec main = MethodSpec.methodBuilder("main")
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .returns(void.class)
+        .addParameter(String[].class, "args")
+        .addComment("Hello\nmultiline\ncomments!")
+        .addStatement("$T.out.println($S)", System.class, "Hello, JavaPoet!")
+        .build();
+
+    assertThat(main.toString()).isEqualTo("" +
+        "public static void main(java.lang.String[] args) {\n" +
+        "  /*\n" +
+        "  * Hello\n" +
+        "  * multiline\n" +
+        "  * comments!\n" +
+        "  **/\n" +
+        "  java.lang.System.out.println(\"Hello, JavaPoet!\");\n" +
+        "}\n");
+  }
+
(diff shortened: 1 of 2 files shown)


src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -456,6 +456,44 @@ static void staticMethod() {
...
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .returns(void.class)
+        .addParameter(String[].class, "args")
+        .addComment("Hello\nmultiline\ncomments!")
+        .addStatement("$T.out.println($S)", System.class, "Hello, JavaPoet!")
+        .build();
+
+    assertThat(main.toString()).isEqualTo("" +
+        "public static void main(java.lang.String[] args) {\n" +
+        "  /*\n" +
```
Egorand: I suggest using the following syntax for inline comments:
```java
public static void main(String[] args) {
  // Hello
  // multiline
  // comments!
}
```
The `/** */` is most frequently used for Javadoc, so generating it inside a method body feels awkward.

octylFractal: While `/**` is for javadoc, `/*` is actually the intended multi-line comment syntax: https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141999.html

Egorand: > The last revision to this document was made on April 20, 1999

octylFractal: It's fine if you would prefer not to have it, but that syntax is valid multi-line comment syntax. You can look at other sources and they all distinguish multi-line comments and JavaDoc comments by the extra asterisk. It's not a valid argument to dismiss this by claiming it as JavaDoc.

theaveragedude: The multiline comment syntax is valid Java as @octylFractal pointed out. It's common to see single line comments completely replaced by block comments throughout source code. The one that's used for JavaDocs begin with slash followed by two asterisks. /* */ and  /** */ are two entirely different things. See [here](https://stackoverflow.com/questions/29815636/and-in-java-comments/29815698) for more information.

Egorand: Yeah, I totally confused the two. My preference is still for `//` - that's less lines and feels more common to me. I haven't seen the `/* */` form used in method bodies in a long while.

PR #770, opened 2020-04-19
Don't copy parameter annotations when creating a ParameterSpec.

danysantiago:
This further preserves the behaviour discussed in https://github.com/square/javapoet/issues/482.
Unifying it for both MethodSpec.overriding and ParameterSpec.get.

### DIFF
--- a/src/test/java/com/squareup/javapoet/MethodSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/MethodSpecTest.java
@@ -15,6 +15,7 @@
  */
 package com.squareup.javapoet;
 
+import com.google.testing.compile.Compilation;
 import com.google.testing.compile.CompilationRule;
 import java.io.Closeable;
 import java.io.IOException;
@@ -30,14 +31,20 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.util.ElementFilter;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
+import javax.tools.JavaFileObject;
+
+import com.google.testing.compile.CompilationSubject;
+import com.google.testing.compile.JavaFileObjects;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.testing.compile.Compiler.javac;
 import static com.squareup.javapoet.MethodSpec.CONSTRUCTOR;
 import static com.squareup.javapoet.TestUtil.findFirst;
 import static javax.lang.model.util.ElementFilter.methodsIn;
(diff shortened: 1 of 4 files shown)


src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -247,6 +254,26 @@ static void staticMethod() {
...
+    TypeElement abstractTypeElement = getElement(AbstractClassWithPrivateAnnotation.class);
+    ExecutableElement fooElement = ElementFilter.methodsIn(abstractTypeElement.getEnclosedElements()).get(0);
+    ClassName implClassName = ClassName.get("com.squareup.javapoet", "Impl");
+    TypeSpec type = TypeSpec.classBuilder(implClassName)
+            .superclass(abstractTypeElement.asType())
+            .addMethod(MethodSpec.overriding(fooElement).build())
+            .build();
+    JavaFileObject jfo = JavaFile.builder(implClassName.packageName, type).build().toJavaFileObject();
+    Compilation compilation = javac().compile(jfo);
+    CompilationSubject.assertThat(compilation).succeeded();
```
ronshapiro, line 274: Static import `assertThat()`?

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -30,14 +31,20 @@
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.util.ElementFilter;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
+import javax.tools.JavaFileObject;
+
+import com.google.testing.compile.CompilationSubject;
+import com.google.testing.compile.JavaFileObjects;
```
ronshapiro, line 40: I think this is unused?

src/test/java/com/squareup/javapoet/MethodSpecTest.java
```diff
@@ -247,6 +254,26 @@ static void staticMethod() {
...
   }
 
+  abstract static class AbstractClassWithPrivateAnnotation {
+
+    private @interface PrivateAnnotation{ }
+
+    abstract void foo(@PrivateAnnotation final String bar);
+  }
+
+  @Test public void overrideNotCopyParameterAnnotations() {
```
Egorand, line 264:
```suggestion
  @Test public void overrideDoesNotCopyParameterAnnotations() {
```

src/test/java/com/squareup/javapoet/MethodSpecTest.java
danysantiago: There is already a static import for `com.google.common.truth.Truth.assertThat` that is way more used than this one, so keeping that one as import and explicit usage of this here.

src/test/java/com/squareup/javapoet/MethodSpecTest.java
danysantiago: It is, removed, thx

src/test/java/com/squareup/javapoet/MethodSpecTest.java
danysantiago: Applied, thx

src/test/java/com/squareup/javapoet/MethodSpecTest.java
JakeWharton: You can overload static imports: https://github.com/square/javapoet/pull/772

Issue #746, opened 2019-12-21
Complete the documentation of formatting placeholders


h908714124:
Please add a section in the README that briefly covers every formatting placeholder. Right now the README only covers `$L`, `$S`, `$T` and `$N`, so `$Z` is missing for example.

XvTianYuan:
I'm trying to write sections for `$<` `$$` `$>` and other place holder signs, but I found it hard to understand the usage of $Z and $W

tbroyer:
From the code, `$Z` will "Emit a newline character if the line will exceed it's limit, otherwise do nothing"; I'd say you would use it for example after an opening parenthesis in a method call, or before a `.` and method call or field access.

`$W` will "Emit either a space or a newline character"; you would use it after a comma in a method call, or a semicolon in a `for` loop, or before/after an operator.

They're both used to conditionally wrap, but behave differently in the case where it doesn't wrap.

h908714124:
How do you escape `[` and `]` for array access?

PR #773, opened 2020-05-01
Update README.md

XvTianYuan:
issue #746
Add description of Space Holder and $Z,$W into README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ void main() {
     System.out.println("Ok, time still moving forward");
   }
 }
-``` 
+```
 
 Catching exceptions using `try/catch` is also a use case for `nextControlFlow()`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -488,6 +488,147 @@ map.put("count", 3);
...
+    .build()
+```
+
+##### `$[` `$]` begin/end a statement. 
+
+`$[` and `$]` are used to start or end a statement. What should be mentioned is that every line after the first line will be double-indented if there is multiple statements.
+
+```java
+CodeBlock codeBlock =CodeBlock.builder().add("class Student")
+                .add("{$[\n")
```
tbroyer, line 535: This is a very bad example: these are not statements!

Switch the examples for `$>` and `$[`.

README.md
```diff
@@ -488,6 +488,147 @@ map.put("count", 3);
...
+}
+```
+
+
+
+#### $Z and $W
+
+##### `$Z` emits a zero-space or newline
+
+`$Z` will emit a newline character if the line will exceed it's limit. If not, `$Z` will emit a zero-width space, which means it will do nothing. Thus, you can use it in method call, such as after an opening parenthesis , or before a dot operator and other similar conditions like field access.
```
tbroyer, line 557: If the line is not too long, `$Z` won't emit anything, this is different from a zero-width space (which is a real Unicode character)

XvTianYuan:
I have modified my update, please have a check.

README.md
```diff
@@ -510,41 +510,40 @@ The price of burger is 10$
...
+CodeBlock codeBlock =CodeBlock.builder().add("void method")
                 .add("{$[\n")
-                .add("String name;\n")
-                .add("int score;\n")
-                .add("public Student(){}")
-                .add("$]\n}")
+                .add("int a = 1;\n")
+                .add("int b = 2;\n")
+                .add("int c = a+b;\n")
+                .add("$]}")
```
tbroyer: Those are 3 statement within a (bogus) method declaration. Basically statement ends with `;`  (this is far from being that simple, but it's a good approximation). Your initial indent example **was** a single statement on multiple lines, and is exactly what those placeholders are meant to make easier (compared to using the indent placeholders).

```java
CodeBlock.builder().
        .add("$[$1T foo = $1T.builder()\n", SomeType.class)
        .add(".foo($S)\n", "bar")
        .add(".baz(42)\n")
        .add(".build();\n$]");
```

See examples in https://github.com/square/javapoet/blob/304176ef98fa80f35cb49ed130620c7b0ecdf5a6/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L2359-L2408 and https://github.com/square/javapoet/blob/304176ef98fa80f35cb49ed130620c7b0ecdf5a6/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L1632-L1719 (because `addStatement` will use `$[` and `$]`)

XvTianYuan: do you mean that I shouldn't put a newline after $[, I am confused

PR #775, opened 2020-05-06
Support lambda with a trick

NoodlesPi:
I have a naive but helpful idea to sovle this issue. I notice that when the parameter in MethodSpec.methodBuilder is a empty string "", it will throw a java.lang.NullPointerException. Actually, we can use this feature to build our lambda because we can stipulate that a method with empty string here is the lambda. To realize it, we just need some slightly revise which do not affect previous code.
Example:
```
MethodSpec lambda = MethodSpec.methodBuilder("")
                        .addParameter(String.class, "a")
                        .addParameter(String.class, "b")
                        .addStatement("return $N.length() - $N.length()", "a", "b")
                        .build();

        TypeSpec helloWorld = TypeSpec.classBuilder("HelloWorld")
                .addMethod(MethodSpec.methodBuilder("sortByLength")
                        .addParameter(ParameterizedTypeName.get(List.class, String.class), "strings")
                        .addStatement("$T.sort($N, $L)", Collections.class, "strings", lambda)
                        .build())
                .build();
```
Output with my modification:
```
class HelloWorld {
  void sortByLength(List<String> strings) {
    Collections.sort(strings, (a, b) -> {
          return a.length() - b.length();
        });
  }
}
```

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -24,6 +24,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
@@ -54,6 +55,7 @@ public final class MethodSpec {
   public final List<TypeName> exceptions;
   public final CodeBlock code;
   public final CodeBlock defaultValue;
+  public final boolean isLambda;
 
   private MethodSpec(Builder builder) {
     CodeBlock code = builder.code.build();
@@ -73,6 +75,7 @@ public final class MethodSpec {
     this.exceptions = Util.immutableList(builder.exceptions);
     this.defaultValue = builder.defaultValue;
     this.code = code;
+    this.isLambda = builder.isLambda;
   }
 
   private boolean lastParameterIsArray(List<ParameterSpec> parameters) {
@@ -82,9 +85,11 @@ public final class MethodSpec {
 
   void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitModifiers)
       throws IOException {
-    codeWriter.emitJavadoc(javadocWithParameters());
-    codeWriter.emitAnnotations(annotations, false);
-    codeWriter.emitModifiers(modifiers, implicitModifiers);
+    if(!isLambda) {
+      codeWriter.emitJavadoc(javadocWithParameters());
+      codeWriter.emitAnnotations(annotations, false);
+      codeWriter.emitModifiers(modifiers, implicitModifiers);
+    }
 
     if (!typeVariables.isEmpty()) {
       codeWriter.emitTypeVariables(typeVariables);
@@ -94,7 +99,10 @@ public final class MethodSpec {
     if (isConstructor()) {
       codeWriter.emit("$L($Z", enclosingName);
     } else {
-      codeWriter.emit("$T $L($Z", returnType, name);
+      if(isLambda)
+        codeWriter.emit("(");
+      else
+        codeWriter.emit("$T $L($Z", returnType, name);
     }
 
     boolean firstParameter = true;
@@ -106,6 +114,8 @@ public final class MethodSpec {
     }
 
     codeWriter.emit(")");
+    if(isLambda)
+      codeWriter.emit(" ->");
 
     if (defaultValue != null && !defaultValue.isEmpty()) {
       codeWriter.emit(" default ");
@@ -134,8 +144,10 @@ public final class MethodSpec {
       codeWriter.indent();
       codeWriter.emit(code, true);
       codeWriter.unindent();
-
-      codeWriter.emit("}\n");
+      if(isLambda)
+        codeWriter.emit("}");
+      else
+        codeWriter.emit("}\n");
     }
     codeWriter.popTypeVariables(typeVariables);
   }
@@ -188,6 +200,7 @@ public final class MethodSpec {
     return new Builder(name);
   }
 
+
   public static Builder constructorBuilder() {
     return new Builder(CONSTRUCTOR);
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -151,6 +177,7 @@ public Builder addJavadoc(CodeBlock block) {
     public Builder addAnnotations(Iterable<AnnotationSpec> annotationSpecs) {
       checkArgument(annotationSpecs != null, "annotationSpecs == null");
       for (AnnotationSpec annotationSpec : annotationSpecs) {
+        System.out.println(annotationSpec.type);
```
JakeWharton, line 180: revert

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -37,28 +38,40 @@
...
+    }
     if (varargs) {
       TypeName.asArray(type).emit(codeWriter, true);
     } else {
-      type.emit(codeWriter);
+      if (type != null)
+        type.emit(codeWriter);
     }
-    codeWriter.emit(" $L", name);
+    if (type != null)
```
JakeWharton, line 71: curly braces

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -37,28 +38,40 @@
...
-    codeWriter.emitModifiers(modifiers);
+    if (type != null) {
+      codeWriter.emitAnnotations(annotations, true);
+      codeWriter.emitModifiers(modifiers);
+    }
     if (varargs) {
       TypeName.asArray(type).emit(codeWriter, true);
     } else {
-      type.emit(codeWriter);
+      if (type != null)
```
JakeWharton, line 68: curly braces

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -37,28 +38,40 @@
...
   public final CodeBlock javadoc;
+  public final boolean isLambda;
 
   private ParameterSpec(Builder builder) {
     this.name = checkNotNull(builder.name, "name == null");
     this.annotations = Util.immutableList(builder.annotations);
     this.modifiers = Util.immutableSet(builder.modifiers);
-    this.type = checkNotNull(builder.type, "type == null");
+    this.isLambda = builder.isLambda;
+    if (!isLambda)
```
JakeWharton, line 48: curly braces

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -402,6 +427,8 @@ public Builder addParameter(ParameterSpec parameterSpec) {
     }
 
     public Builder addParameter(TypeName type, String name, Modifier... modifiers) {
+      if (isLambda)
```
JakeWharton, line 430: curly braces

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -389,7 +413,8 @@ public Builder returns(Type returnType) {
     }
 
     public Builder addParameters(Iterable<ParameterSpec> parameterSpecs) {
-      checkArgument(parameterSpecs != null, "parameterSpecs == null");
+      if (!isLambda)
```
JakeWharton, line 416: curlies

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -188,6 +200,7 @@ public static Builder methodBuilder(String name) {
     return new Builder(name);
   }
 
+
```
JakeWharton, line 203: revert

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -134,8 +144,10 @@ void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitMod
       codeWriter.indent();
       codeWriter.emit(code, true);
       codeWriter.unindent();
-
-      codeWriter.emit("}\n");
+      if (isLambda)
```
JakeWharton, line 147: curly braces

src/main/java/com/squareup/javapoet/MethodSpec.java
```diff
@@ -94,7 +99,10 @@ void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitMod
     if (isConstructor()) {
       codeWriter.emit("$L($Z", enclosingName);
     } else {
-      codeWriter.emit("$T $L($Z", returnType, name);
+      if (isLambda)
```
JakeWharton, line 102: curly braces

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -37,28 +38,40 @@
...
+      this.type = checkNotNull(builder.type, "type == null");
+    else
+      this.type = null;
     this.javadoc = builder.javadoc.build();
+
```
JakeWharton, line 53: remove

src/main/java/com/squareup/javapoet/ParameterSpec.java
```diff
@@ -86,10 +99,14 @@ void emit(CodeWriter codeWriter, boolean varargs) throws IOException {
   public static ParameterSpec get(VariableElement element) {
     checkArgument(element.getKind().equals(ElementKind.PARAMETER), "element is not a parameter");
 
+    // Copy over any annotations from element.
+    List<AnnotationSpec> annotations = element.getAnnotationMirrors()
+        .stream()
+        .map((mirror) -> AnnotationSpec.get(mirror))
```
JakeWharton, line 105: nit;
```suggestion
        .map(AnnotationSpec::get)
```

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -38,6 +38,7 @@
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.SimpleTypeVisitor8;
 
+
```
JakeWharton, line 41: revert

NoodlesPi:
> I'm tempted to instead force the use of an anonymous `ClassSpec` which could be converted to a lambda given a source level. This change lacks a ton of tests. I would need to see coverage of edge cases like inline lambdas `e -> e.foo()`, block lambdas `e -> { return e.foo(); }`, typed argument-lambdas like `String e -> e.foo()`, annotated argument lambdas like `@NonEmpty var e -> e.foo()`, and annotated typed lambdas like `@NonEmpty String e -> e.foo()` at least.

Yes, sure. But one issue with anonymous classes is that if the implementation of your anonymous class is very simple, such as an interface that contains only one method, then the syntax of anonymous classes may seem unwieldy and unclear. In these cases, you're usually trying to pass functionality as an argument to another method. This is the intention of the lambda. And from the user's perspective, they just care the functionality when use lambda. Therefore, it's may more reasonable to implements lambda using 'MethodSpec'. As for anonymous 'ClassSpec', it's also can be converted to a lambda but it seems not straightforward.
I just present my ideas here and do a simple implementation without enough tests. If you think this idea is OK, i will complete my work.

PR #778, opened 2020-05-30
add sort Method (Issues 683)

LuzzyQWQ:
Add a sort method for sort `methodSpec` in dictionary order in `TypeSpec.class`. 
I write some JUnit test to show the result.
By the way, maybe `FieldSpec`, `TypeVariableName`, e.t.c. can also be sortable?


JakeWharton:
The list of method specs is public and mutable so that you can do things like this without needing to support it in the library. We don't want to build it in since, in general, this is not a very good way to sort your members (see https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents).

Issue #666, opened 2018-08-07
Add a way to generate package-info.java files


acabarbaye:
Would it be possible to add a way to generate package-info.java files?

If so, such generator should let the possibility to annotate the package as follows:
```java
/**
 * Some comments
 */
@AnAnnotation
package a.package;

import package.AnAnnotation;
```

swankjesse:
Yeah we need like a PackageSpec and maybe also a ModuleSpec.

JakeWharton:
Isn't the first one a JavaFile without any types?

kenzierocks:
No, there's no support for annotations on the package.

JakeWharton:
But we can add that.

jdyjjj:
Maybe I want to try to sovle this problem.

PR #786, opened 2020-06-01
fix issue #666

UwillNEVERKNW:


JakeWharton:
In its current form, no. It adds all kinds of public API that shouldn't be exposed and creates mutable state in an otherwise stateless class.

See also #866

PR #788, opened 2020-06-04
Add iterable overloads for parameterized parameters

### LINKED ISSUE #727: ParameterizedTypeName ergnomic improvements
`get` doesn't accept an `Iterable<? extends TypeName>` which makes it annoying to build incrementally.

Alternatively or additionally, a builder would be nice.

JakeWharton:
Closes #727. Closes #781.

This, admittedly, isn't the greatest implementation. But it's the easiest.


src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -114,11 +114,31 @@ public static ParameterizedTypeName get(ClassName rawType, TypeName... typeArgum
...
   }
 
+  /** Returns a parameterized type, applying {@code typeArguments} to {@code rawType}. */
+  public static ParameterizedTypeName get(
+      ClassName rawType, Iterable<? extends TypeName> typeArguments) {
+    ArrayList<TypeName> typeArgumentsList = new ArrayList<>();
+    for (TypeName typeArgument : typeArguments) {
+      typeArgumentsList.add(typeArgument);
+    }
+    return get(rawType, typeArgumentsList.toArray(new TypeName[0]));
```
ronshapiro, line 124: Why are you calling toArray instead of calling the constructor directly?

JakeWharton: Ah this is copy/pasta. This overload used to do that, but I couldn't do that with the other one since it needs to pass through the `list(..)` helper which only accepts an array and changing that to accept a `List` is far more invasive.

JakeWharton: Fixed

src/main/java/com/squareup/javapoet/ParameterizedTypeName.java
```diff
@@ -114,11 +114,31 @@ public static ParameterizedTypeName get(ClassName rawType, TypeName... typeArgum
...
   }
 
+  /** Returns a parameterized type, applying {@code typeArguments} to {@code rawType}. */
+  public static ParameterizedTypeName get(
+      Class<?> rawType, Iterable<? extends Type> typeArguments) {
+    ArrayList<Type> typeArgumentsList = new ArrayList<>();
+    for (Type typeArgument : typeArguments) {
+      typeArgumentsList.add(typeArgument);
+    }
+    return get(rawType, typeArgumentsList.toArray(new Type[0]));
```
octylFractal, line 139: Doesn't https://github.com/square/javapoet/pull/788#discussion_r435436043 apply here too? Seems like the same code.

octylFractal: Ah, NVM -- I see the translation that happens in `list()`.

JakeWharton: Yeah fixing that quickly spirals outward into more of the codebase. I'm happy to pursue such a change, but we can do so separately. Plus, this isn't particularly performance sensitive code so I figured it wasn't worth the effort.

PR #789, opened 2020-06-04
Add support for explicit receiver parameters

### LINKED ISSUE #787: Explicit receiver parameters
Explicit receiver parameters (annotated `this` parameters for instance methods and non-static inner class constructors) do not seem to be supported in `MethodSpec`.

```java
class SomeClass
{
  public void SomeMethod(@SomeAnnotation SomeClass this, int param) {}
}
```

https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.1

The simplest way to implement this is probably allowing `this` for parameter names in `MethodSpec.Builder#addParameter(ParameterSpec, String)`, which is currently only checked against `javax.lang.model.SourceVersion#isName(CharSequence)`.

Alternatively:

- `MethodSpec.Builder#receives(TypeName)` - order no longer matters.
- `MethodSpec.Builder#addReceiverAnnotation(AnnotationSpec)` - type is implicitly set (though maybe less resilient to future Java changes).

jasonobrien:
Annotated "this" arguments for instance methods and non-static nested
class constructors were added in Java 8. This special cases "this" and
"OuterClass.this" so that they are allowed as parameter names.

Closes #787

### DIFF
--- a/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
+++ b/src/test/java/com/squareup/javapoet/ParameterSpecTest.java
@@ -60,6 +60,25 @@ public class ParameterSpecTest {
     assertThat(a.toString()).isEqualTo(b.toString());
   }
 
+  @Test public void receiverParameterInstanceMethod() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "this");
+    assertThat(builder.build().name).isEqualTo("this");
+  }
+
+  @Test public void receiverParameterNestedClass() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "Foo.this");
+    assertThat(builder.build().name).isEqualTo("Foo.this");
+  }
+
+  @Test public void keywordName() {
+    try {
+      ParameterSpec.builder(int.class, "super");
+      fail();
+    } catch (Exception e) {
+      assertThat(e.getMessage()).isEqualTo("not a valid name: super");
+    }
+  }
+
   @Test public void nullAnnotationsAddition() {
     try {
       ParameterSpec.builder(int.class, "foo").addAnnotations(null);
--- a/src/main/java/com/squareup/javapoet/ParameterSpec.java
+++ b/src/main/java/com/squareup/javapoet/ParameterSpec.java
@@ -103,9 +103,18 @@ public final class ParameterSpec {
     return result;
   }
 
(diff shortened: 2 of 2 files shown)


src/test/java/com/squareup/javapoet/ParameterSpecTest.java
```diff
@@ -60,6 +60,25 @@ private TypeElement getElement(Class<?> clazz) {
     assertThat(a.toString()).isEqualTo(b.toString());
   }
 
+  @Test public void receiverParameterInstanceMethod() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "this");
+    assertThat(builder.build().name).isEqualTo("this");
+  }
+
+  @Test public void receiverParameterNestedClass() {
+    ParameterSpec.Builder builder = ParameterSpec.builder(int.class, "Foo.this");
```
JakeWharton, line 69: Oh crap this is scary. Does this work with `Foo.Bar.this`? Or `com.example.Foo.this`? If so, that means it probably needs to participate in the automatic import system to ensure you're referring to what you actually mean.

jasonobrien: Luckily `Foo.this` can only be used from the constructor of a direct nested type of `Foo`, so no import required.

In fact, the grammar only allows a single unqualified identifier, so even if you managed to make it ambiguous (another inner nested type with the outer types name?), there's nothing JavaPoet could do to make it compile I think.

```
ReceiverParameter:
{Annotation} UnannType [Identifier .] this
```

That said, this is the simplest solution but probably not the most elegant - if the grammar was more complicated, I might suggest `MethodSpec#receives(TypeName)` with the identifier derived automatically during emission instead.

JakeWharton: I'm not worried about it then. Thanks! Frankly I don't expect many to ever need this feature as-is but it's so easy to support we might as well add it.

jasonobrien: I agree - relevant receiver annotations seem pretty rare (I only know the one), and needing one in a generated class even more so.

If support for Java 8 annotations was otherwise lacking I'd probably have just hacked around it, but it seems very thorough. Thanks for looking!

Issue #795, opened 2020-07-13
MethodSpec.beginControlFlow inserts unneccessary space when argument is empty


DanielGronau:
I want to write code like this:

```
public void test() {
   {
     int value = 42;
     someCode(value, 5, value);
   }
   {
     int value = 23;
     someCode(value, 7, value);
   }
}
```
To generate the code blocks, I use

```
...
.beginControlFlow("");
.addStatement("int value = 42")
.addStatement("someCode(value, 5, value)")
.endControlFlow();
...
```
This works _almost_ as expected, but the curly braces don't line up. Would it be possible to remove the extra space when beginControlFlow gets an empty argument?

Egorand:
Sounds like a legit use case. I'd actually suggest adding a no-arg `beginControlFlow()` overload in addition to fixing the behavior for the empty string case. Wanna submit a PR?

JakeWharton:
I'm not sure this is control flow though. You could build the block contents in a `CodeBlock` and then add it to the enclosing method with `addCode("{$>\n$L\n$<}\n", block)`.

DanielGronau:
@JakeWharton I think this would be a quite theoretical distinction. Is this a control flow then?

```
if(true) {
    System.out.println("hey");
}
```

JakeWharton:
As far as JavaPoet is concerned yes. The library does not care what the semantic meaning your code has, we merely care about its syntactical form.

hariohmprasath:
Hi, I have created a PR for this issue, please review and let me know if it looks good for merge:
[https://github.com/square/javapoet/pull/796](url)

jdyjjj:
I added a no-arg beginControlFlow() overload in addition to fixing the behavior for the empty string case.

PR #796, opened 2020-07-17
Fix for #795 - MethodSpec.beginControlFlow inserts unneccessary space when argument is empty

hariohmprasath:
Code changes made to handle spaces for both empty and non-empty "beginControlFlow"


JakeWharton:
I remain opposed to this change.

PR #804, opened 2020-09-08
#803 Use Set for TypeSpec.Builder.superinterfaces

### LINKED ISSUE #803: Adding a superinterface twice to a TypeSpec causes invalid Java
We have a case where we are building up a Type and we add interfaces to it based on other content.

Since `TypeSpec.Builder.superinterfaces` is a `List` instead of a `Set`, it allows adding the same interface twice, meaning we have to check if it contains before adding to it.

bentatham:
Fixes #803 
In TypeSpec.Builder, by simply switching to a `Set` instead of a `List`, this issue is solved. Use `LinkedHashSet` to preserve ordering. The `TypeSpec` itself can continue to use a `List` since it is immutable and cheaper to preserve ordering that way.

Added test case to demonstrate that this works.


JakeWharton:
This is unfortunately a binary-incompatible change, though.

bentatham:
I changed it to be binary compatible, as I understand it (even if I don't understand that requirement -- is there a reference to an explanation as to why you require that in a library that is ostensibly always used at build time?)

JakeWharton:
This library is used by others which depend on it in binary form in places like the annotation processor classpath. If I depend on library A and library B which both depend on JavaPoet, library B could upgrade the dependency past what library A built against. We don't want to crash with NoSuchMethodError / NoSuchFieldError in this case.

ronshapiro:
I think it may be better to throw here, which is line with other invalid combinations of JavaPoet builder methods (like modifiers).

But either way, if we do this, we should do it consistently for all of the builder methods in the library. Otherwise it will be confusing why some behave differently.

bentatham:
@ronshapiro I don't disagree - I'm happy to add similar checks for the ones that make sense (annotations, modifiers). But the others I would expect to fail if you add the same thing twice (typeVariables, fieldSpecs, methodSpecs, etc). I understand that there might be some disagreement there.  

Due to differing opinions, perhaps best left to the user of the library to decide to handle this or not (which is what I've done in the meantime anyhow).

Let me know if you want to add more of the same to other variables, change to an error, or discard this PR altogether.

eamonnmcmanus:
Yeah, I think it isn't right to ignore duplicate interfaces, any more than it would be right to ignore duplicate field definitions. If you say you want to implement interfaces `A, B, C, D` then you expect to see `implements A, B, C, D` in the generated code, and if you've accidentally made `D` the same as `B` then you should expect that to be an error. If `B` and `D` are both `List<Integer>` would you drop `D`? What if `B` were `List<Integer>` and `D` were `List<String>`? Then you can neither drop `D` nor keep it, since a class can't implement two incompatible instantiations of the same generic interface.

PR #806, opened 2020-09-28
Allow private Interface methods

### LINKED ISSUE #805: Java 9 allows private methods on Interfaces. Javapoet's MethodSpec doesn't allow this
```java
        MethodSpec methodSpec = MethodSpec.methodBuilder(name)
                .addModifiers(Modifier.PRIVATE)
                .addCode(codeBlock)
                .build();
```

The above will fail. Yet, on Java 9+ this is a legal combination of modifiers.

Randgalt:
Closes #805 

Since Java 9 interface methods can be private

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -796,9 +796,16 @@ public final class TypeSpec {
 
       for (MethodSpec methodSpec : methodSpecs) {
         if (kind == Kind.INTERFACE) {
-          requireExactlyOneOf(methodSpec.modifiers, Modifier.ABSTRACT, Modifier.STATIC,
-              Modifier.DEFAULT);
           requireExactlyOneOf(methodSpec.modifiers, Modifier.PUBLIC, Modifier.PRIVATE);
+          if (methodSpec.modifiers.contains(Modifier.PRIVATE)) {
+            checkState(!methodSpec.hasModifier(Modifier.DEFAULT),
+                "%s %s.%s cannot be private and default", kind, name, methodSpec.name);
+            checkState(!methodSpec.hasModifier(Modifier.ABSTRACT),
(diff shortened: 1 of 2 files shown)


src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -796,9 +796,14 @@ public TypeSpec build() {
 
       for (MethodSpec methodSpec : methodSpecs) {
         if (kind == Kind.INTERFACE) {
-          requireExactlyOneOf(methodSpec.modifiers, Modifier.ABSTRACT, Modifier.STATIC,
-              Modifier.DEFAULT);
           requireExactlyOneOf(methodSpec.modifiers, Modifier.PUBLIC, Modifier.PRIVATE);
+          if (methodSpec.modifiers.contains(Modifier.PRIVATE)) {
+            checkState(!methodSpec.hasModifier(Modifier.DEFAULT), "%s %s.%s cannot be private and default",
```
octylFractal, line 801: Potentially also check that it's not `ABSTRACT`? I don't think that's valid either.

Though perhaps a check like that belongs higher up, since it applies to _all_ `private` methods, not just ones in interfaces.

octylFractal (changes requested):

JakeWharton:
Can you add a test? There should be tests in `TypeSpecTest` for things like static and default methods in an interface from when Java 8 was released. You can copy/paste/modify those for this.

Randgalt:
Ping

liach (approved):

JakeWharton:
Will look tomorrow

JakeWharton:
@Randgalt can you sign the Square CLA at https://squ.re/sign-the-cla?

Randgalt:
@JakeWharton done

JakeWharton:
Thanks!

PR #811, opened 2020-10-28
GitHub actions

HanSolo:
Added a github action to run mvn test on each push using the latest Ubuntu and JDK 8.


JakeWharton:
Unless we're going to migrate fully, I don't see much reason to have both CI systems running.

PR #812, opened 2020-11-01
return path when writing to path

arnzel:
return Path when writing to Path


JakeWharton:
This is a binary incompatible change. The reason we added methods with new names was to retain binary compatibility.

PR #816, opened 2020-12-11
 Add a className property to TypeSpec

fkleedorfer:
* changes the TypeSpec.(class|interface|..)Builder(ClassName) methods to pass the ClassName into the builder
* The builder passes the ClassName instance into the TypeSpec it generates
* The TypeSpec has a getter for the ClassName that throws a helpful exception if the TypeSpec was not instantiated with a ClassName but only a string
* A JavaFile.builder method is added that extracts the package name from the specified TypeSpec (which will throw the helpful exception if no ClassName is set)

The basic idea is explained in the change to README.md:

---
Sometimes, you'll generate a `TypeSpec` and require its `ClassName` later, for example, in a 
method declaration. If that is the case, create the `TypeSpec` instance using a `ClassName` object, and you'll 
be able to retrieve that `ClassName` from the instance:

```
// We'll need the ClassName later, so we provide it when obtaining the builder:
TypeSpec typeA = TypeSpec.classBuilder(ClassName.get("com.foo", "A"))
        .addModifiers(PUBLIC).build())
    .build();
                
// Now, we can use typeA.getClassName() to retrieve it 
// (see the line that starts with '.returns' )
TypeSpec typeB = TypeSpec.classBuilder(ClassName.get("com.foo", "B"))
    .addMethod(MethodSpec.methodBuilder("createA")
        .addModifiers(STATIC, PUBLIC)
        .returns(typeA.getClassName())
        .addStatement("return new $T", typeA.getClassName())
        .build())
    .build();
```
---
Additional Note: I had to deactivate the checkstyle plugin to be able to compile the project, some style errors seem unfixable. The current source on master does not seem to pass the test. Please advise if you want the PR to pass checkstyle before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,28 @@ MethodSpec beyond = MethodSpec.methodBuilder("beyond")
     .build();
 ```
 
+Sometimes, you'll generate a `TypeSpec` and require its `ClassName` later, for example, in a 
+method declaration. If that is the case, create the `TypeSpec` instance using a `ClassName` object, and you'll 
+be able to retrieve that `ClassName` from the instance:
+
+```
+// We'll need the ClassName later, so we provide it when obtaining the builder:
+TypeSpec typeA = TypeSpec.classBuilder(ClassName.get("com.foo", "A"))
+    .addMethod(MethodSpec.constructorBuilder()
+        .addModifiers(PUBLIC).build())
+    .build();
+                
+// Now, we can use typeA.getClassName() to retrieve it 
+// (see the line that starts with '.returns' )
+TypeSpec typeB = TypeSpec.classBuilder(ClassName.get("com.foo", "B"))
+    .addMethod(MethodSpec.methodBuilder("createA")
+        .addModifiers(STATIC, PUBLIC)
+        .returns(typeA.getClassName())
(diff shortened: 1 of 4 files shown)


JakeWharton:
This is unreviewable in its current form. Please revert all formatting changes. You can find our code style at https://github.com/square/java-code-styles/.

Going from the PR description, however, I'm skeptical this is an approach we want to take. I don't like properties that are conditionally available with no recourse for the consumer when unavailable.

fkleedorfer:
Thanks for the link to the code style, that makes it easier. That said, it still does not pass the checkstyle test, I always get the `NewLineAtEndOfFile ` error. Not sure what that is about.

fkleedorfer:
As to the criticism, I agree with this not being ideal, as it is still possible to make `TypeSpec`s without a `ClassName`. However,
* the user is responsible for instantiating `TypeSpecs`, so if they find they want to access the `ClassName` later but don't provide it when creating the builder, they can change the way they create the builder.
* As suggested in #197 I would actually go further, e.g., deprecating the ways of instantiating `TypeSpec` without a `ClassName`. If you can imagine going in that direction, I could provide another PR with more changes.

PR #821, opened 2021-01-04
fix #637: Prevent class loading deadlock between TypeName and ClassName

### LINKED ISSUE #637: Deadlock in Java 8
We are seeing deadlocks when running several different instances of java poet in parallel threads

```
"Task worker for ':' Thread 2" #147 prio=5 os_prio=0 tid=0x00007fcdb870a000 nid=0x6894 in Object.wait() [0x00007fcd0b8f8000]
   java.lang.Thread.State: RUNNABLE
        at com.squareup.javapoet.TypeSpec$Builder.<init>(TypeSpec.java:403)
        at com.squareup.javapoet.TypeSpec$Builder.<init>(TypeSpec.java:394)
        at com.squareup.javapoet.TypeSpec.interfaceBuilder(TypeSpec.java:121)
        ...

"Task worker for ':' Thread 3" #148 prio=5 os_prio=0 tid=0x00007fcdb870b000 nid=0x6895 in Object.wait() [0x00007fcd0b7f7000]
   java.lang.Thread.State: RUNNABLE
        at com.squareup.javapoet.TypeSpec$Builder.<init>(TypeSpec.java:403)
        at com.squareup.javapoet.TypeSpec$Builder.<init>(TypeSpec.java:394)
        at com.squareup.javapoet.TypeSpec.interfaceBuilder(TypeSpec.java:121)
        ...
```

The current hypothesis is that it is related to https://bugs.openjdk.java.net/browse/JDK-8037567, where [ClassName](https://github.com/square/javapoet/blob/d91f8e1374f7ccdb68d92c206c84864cd09c02f6/src/main/java/com/squareup/javapoet/ClassName.java) is a subclass of TypeName, but TypeName has a static variable of type ClassName [here](https://github.com/square/javapoet/blob/d91f8e1374f7ccdb68d92c206c84864cd09c02f6/src/main/java/com/squareup/javapoet/TypeName.java#L78).

carterkozak:
Previously TypeName included static fields initialized with
values of the ClassName subtype, resulting in a deadlock on
class initialization when separate threads attempt to load
ClassName and TypeName in parallel.

The static fields of type ClassName have been migrated from
TypeName to ClassName to prevent this deadlock.

NOTE: This includes a public API break. TypeName.OBJECT has been
removed, consumers should prefer ClassName.OBJECT instead.

### DIFF
--- a/src/main/java/com/squareup/javapoet/ClassName.java
+++ b/src/main/java/com/squareup/javapoet/ClassName.java
@@ -32,6 +32,15 @@ import static com.squareup.javapoet.Util.checkNotNull;
 /** A fully-qualified class name for top-level and member classes. */
 public final class ClassName extends TypeName implements Comparable<ClassName> {
   public static final ClassName OBJECT = ClassName.get(Object.class);
+  static final ClassName BOXED_VOID = ClassName.get("java.lang", "Void");
+  static final ClassName BOXED_BOOLEAN = ClassName.get("java.lang", "Boolean");
+  static final ClassName BOXED_BYTE = ClassName.get("java.lang", "Byte");
+  static final ClassName BOXED_SHORT = ClassName.get("java.lang", "Short");
+  static final ClassName BOXED_INT = ClassName.get("java.lang", "Integer");
+  static final ClassName BOXED_LONG = ClassName.get("java.lang", "Long");
+  static final ClassName BOXED_CHAR = ClassName.get("java.lang", "Character");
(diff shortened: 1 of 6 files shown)


liach (approved):

ignesious:
@liach  We are having a production problem on the same deadlock scenario. I dont see this is released yet. The latest is 1.13.0 on June 18 2020. 

Please release this fix.

PR #830, opened 2021-03-18
Use GitHub Actions for CI

JakeWharton:


.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version:
+          - 1.8
+          - 9
+          - 10
+          # TODO:
```
JakeWharton, line 16: maven-compiler-plugin chokes on these. Probably best to switch to Gradle first anyway...

Egorand (approved):

PR #834, opened 2021-04-23
Split up TypeSpec.emit(...) to avoid method length limit

nhwalker:
I was looking into adding support for the Java 16 features records and sealed classes.  But quickly ran into a checkstyle error stating that the TypeSpec.emit() method is too many lines.  So I used my IDE's "extractMethod" capability to split TypeSpec.emit() into bite sized pieces.  No functionality of the TypeSpec.emit() method was changed, just moved around.


JakeWharton:
We should get rid of checkstyle and use google-java-format.

PR #840, opened 2021-05-13
Record support

### LINKED ISSUE #829: Add support for Java 16 records
Java 16 has added support for records under [JEP 395](https://openjdk.java.net/jeps/395), of the form. 

```
public record Point(int x, int y) {
}
```

It would be great to support these in JavaPoet, as among other things, they significantly reduce boilerplate.

liach:
Fixes #829

A short summary:
Added record support to javapoet as it has become a stable feature in modern java. Turns out the support is quite easy given the robust existing framework in javapoet.

1. Shared some code from method spec for emitting parameter and their javadocs for record components.
2. Moved the emission of type headers to a separate method since otherwise the type emission method would be too long (per checkstyle)
3. Added a few tests ensuring the functionalities for basic record creation.

`mvn clean verify` ends with 17 file-ending errors (presumably because of crlf/lf problems), which I cannot fix locally; all tests pass. CLA signed as well.

It's my pleasure to have you reviewing my code.

P.S. for more information about java records, feel free to check out [the related java language specification section](https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.10) and look at [a jdk example](https://github.com/openjdk/jdk/blob/6dd06add9d6c4a4198541118f4bea209ba995609/src/jdk.net/share/classes/jdk/net/UnixDomainPrincipal.java) and [its javadocs](https://docs.oracle.com/en/java/javase/16/docs/api/jdk.net/jdk/net/UnixDomainPrincipal.html)

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -80,7 +80,8 @@ public final class MethodSpec {
         && TypeName.asArray((parameters.get(parameters.size() - 1).type)) != null;
   }
 
-  void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitModifiers)
+  void emit(CodeWriter codeWriter, String enclosingName, Set<Modifier> implicitModifiers,
+      boolean compactConstructor)
       throws IOException {
     codeWriter.emitJavadoc(javadocWithParameters());
     codeWriter.emitAnnotations(annotations, false);
@@ -91,21 +92,17 @@ public final class MethodSpec {
       codeWriter.emit(" ");
     }
 
-    if (isConstructor()) {
-      codeWriter.emit("$L($Z", enclosingName);
+    if (compactConstructor) {
+      codeWriter.emit("$L", enclosingName);
     } else {
-      codeWriter.emit("$T $L($Z", returnType, name);
-    }
+      if (isConstructor()) {
+        codeWriter.emit("$L", enclosingName);
+      } else {
+        codeWriter.emit("$T $L", returnType, name);
+      }
 
-    boolean firstParameter = true;
-    for (Iterator<ParameterSpec> i = parameters.iterator(); i.hasNext(); ) {
-      ParameterSpec parameter = i.next();
-      if (!firstParameter) codeWriter.emit(",").emitWrappingSpace();
-      parameter.emit(codeWriter, !i.hasNext() && varargs);
-      firstParameter = false;
+      emitParameters(codeWriter, parameters, varargs);
     }
-
-    codeWriter.emit(")");
 
     if (defaultValue != null && !defaultValue.isEmpty()) {
       codeWriter.emit(" default ");
@@ -141,6 +138,11 @@ public final class MethodSpec {
   }
 
   private CodeBlock javadocWithParameters() {
+    return makeJavadocWithParameters(javadoc, parameters);
+  }
+
+  static CodeBlock makeJavadocWithParameters(CodeBlock javadoc,
+      Iterable<ParameterSpec> parameters) {
     CodeBlock.Builder builder = javadoc.toBuilder();
     boolean emitTagNewline = true;
     for (ParameterSpec parameterSpec : parameters) {
@@ -152,6 +154,22 @@ public final class MethodSpec {
       }
     }
     return builder.build();
+  }
+
+  static void emitParameters(CodeWriter codeWriter, Iterable<ParameterSpec> parameters,
+      boolean varargs) throws IOException {
+    codeWriter.emit(CodeBlock.of("($Z"));
+
+    boolean firstParameter = true;
+    for (Iterator<ParameterSpec> i = parameters.iterator(); i.hasNext(); ) {
+      ParameterSpec parameter = i.next();
+      if (!firstParameter)
+        codeWriter.emit(",").emitWrappingSpace();
+      parameter.emit(codeWriter, !i.hasNext() && varargs);
+      firstParameter = false;
+    }
+
+    codeWriter.emit(")");
   }
 
   public boolean hasModifier(Modifier modifier) {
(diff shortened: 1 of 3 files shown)


Randgalt:
Hi @JakeWharton - can you possibly let the community know the plans for JavaPoet and support for post Java 11 features? Will this PR or something similar be committed to the codebase soon? Thank you.

JakeWharton:
I no longer participate in the day-to-day of this repository or have the bandwidth to review changes with such large API surface. Hopefully @Egorand or someone will pick it up.

Randgalt:
If Square isn't going to continue with Javapoet maybe we can fork it and continue supporting it. I'd be happy to host/republish it. Let's see what @Egorand says.

JakeWharton:
You are always free to do so, but don't mistake my lack of involvement which has been consistent for the last 5 years as any larger indicator. I was mentioned personally therefore I commented on my personal involvement.

Randgalt:
Thanks @JakeWharton

tanadeau:
Any updates on this? This is one of the exciting features of Java 17 since it's now been released.

liach:
@Egorand Mind review this, or is javapoet of low priority right now?

Egorand:
Apologies for not replying sooner and thanks for your contribution! Unfortunately I lost context on this project and Java in general a while ago and can't review this efficiently at the moment. I'll try to find someone who can, or hopefully find some time in the near term to catch up.

arouel:
I reviewed the code and I think it is good to go. It doesn't break the existing API nor logic and introduces Record support.

I'm also in for forking this project, if there is no interest or time from Square anymore to maintain this project.

fcurts:
Are there any good alternatives to JavaPoet that support Java 17?

Randgalt:
If we do fork, what needs to be done? I have a Maven Group ID prefix we can use and I can already publish to Maven Central with it. I think we'd have to change the project's group ID at minimum. Do we have to change the license? Copyright? Package names? Anyone know?

gzsombor:
You should not change the license or copyright, that would be illegal.
 When I fork a project, mostly for personal usage, I just use jitpack.io, which can build any Maven/Gradle projects and distribute the built jars in a coherent way, for example you can use javapoet from there with:
 https://jitpack.io/#square/javapoet/javawriter-2.5.1

Randgalt:
The difference here is that I would like to re-publish this so others can use/contribute. That's allow under AL2 right?

fcurts:
Changing the license isn't illegal but shouldn't be necessary. I wonder if the package name should be changed though.

Publishing to Maven Central is more trustworthy than jitpack (which is denylisted by many organizations).

The main question is whether someone would be willing/able to maintain a fork in the long run.

gzsombor:
You can only modify the license, if you were the original author / copyright holder. Probably in javapoet case, it would mean if all the contributor agree on a license change. But I don't think license change is needed, AL2 is a really permissive license, you can freely modify, publish,etc.
 Sure, publishing to Maven Central makes the code more trustworthy, and makes it more easy to use.

liach:
Fyi I have a fork including this content, available over github packages https://github.com/liachmodded/javapoet

JakeWharton:
Can you add a test with a secondary constructor?

Something like
```java
record Person(String name) {
    Person(int number) {
        this.name = Integer.toString(number);
    }
}
```

liach:
yeah, will do later today

JakeWharton:
Looks good otherwise. I'll have another read of the JLS just to make sure we're covering our bases.

liach:
Done. CI results available at https://github.com/liachmodded/javapoet/runs/4255899160 (The java 8 one fails because setup java action cannot find java 8 by `1.8`)

JakeWharton:
I thought about this over the weekend more.

I'm not sure it makes sense to put the record parameters directly on the type. What do you think about instead changing `compactConstructor` to `recordConstructor` and pulling the parameter specs from that? It also means that we do not need to place the awkward `varargs` property on a type.

liach:
Such a setup may be feasible, but compared to the current impl, there are a few behaviors to be defined:
1. Currently, two sets of parameter specs actually differ for the javadoc they respectively contribute to the compact constructor/class. Need to check in javac on what the behavior of compact constructor parameter docs are like.
2. Think a compact constructor with an empty body is legal as well, and we might want to distinguish that from the absence of a compact constructor.

fcurts:
How can we get this unblocked? Do the maintainers intend to continue this project?

marceloverdijk:
I was looking for this as well...
Would be nice if Square would let the community know what they intend to do this with the library; last commit is from 9 months ago so seems to have at least stalled...

swankjesse:
# Update from Square

Apologies for radio silence on this. We're busy!

I'm talking to our internal copyright & IP legal team to hand off the project. There's logistics here to figure out how & who. But the plan is to hand-off JavaPoet to folks that can give it the attention it deserves.

marceloverdijk:
@swankjesse Thx for the clarification.

fcurts:
Another month has passed. Are they any news on handing off the project?

JakeWharton:
Handing off the project is orthogonal to this PR. And this is not a good place to discuss it. There's a discussions tab. Please use that from now on for project-wide discussion topics.

As to this PR, I expressed concern with the existing shape of the API design so it will not be merged as-is. Either the changes need made to this PR or a new PR needs opened.

We are not working on this project so we no one from our side will be making the changes required to land this support. That is unfortunate, but placing this project into the open is not a promise that we would work on it forever. It exists in open source for you to use, or not use, free of charge. If you are blocked and need this API, forking the project and adding support through this PR's diff or your own design is not that hard to do. You will have to maintain that fork going forward, but doing so is orders of magnitude more trivial than having to write an entire Java code generation API in the first place. There are surely other libraries or template engines that you could use as well.

Please keep future discussions on this PR about the design of the API in the diff.

tanadeau:
@JakeWharton: There is not a discussions tab turned on for this repo.

JakeWharton:
Ah, thanks for the heads up! The tab _was_ enabled but apparently I had to do a one-time setup thing (which I do not recall doing on other repos).

fcurts:
DIscussion: #866

pawellabaj:
Is there anyone who can tell what's the state of this PR?
Is there a chance to merge it?

PR #867, opened 2022-02-27
upgrade to GitHub Actions setup-java v3

sullis:


JakeWharton:
Seems like we're missing dependabot config.

PR #872, opened 2022-04-17
Add value constant in AnnotationSpec

dsinghvi:
I have "value" throughout my code in AnnotationSpecBuilders and it would be nice to replace them with a AnnotationSpec.Value


JakeWharton:
I'm not sure how valuable this is. "value" is mentioned in the JLS, however, so I don't feel strongly and would be comfortable merging.

PR #879, opened 2022-04-21
Fix #860: Nested Type name clashes within same package

### LINKED ISSUE #860: Nested Type name clashes within same package
When we have
```
interface Component {
  interface NestedType;
}
```
If we want to generate a ComponentImpl that reference a type that has the same simple name as a nested type in Component, then there will be a type name clash.
```
package test;
import test.sub.NestedType;

ComponentImpl implements Component {
  NestedType fun() {}
}
```
There is a `avoidClashesWithNestedClasses ` can help with workaround the issue, it will keep the package prefix of NestedType in ComponentImpl. Then the generated code will be like
```
package test;

ComponentImpl implements Component {
  test.sub.NestedType fun() {}
}
```

But if the external Nested type is within the same package.
```
ComponentImpl implements Component {
  NestedType
}
```
The package prefix will gets stripped anyways. So even I added `avoidClashesWithNestedClasses`, it will still clash.

I think it is because in https://github.com/square/javapoet/blob/742af18b143dab63e757d9952c7773f60c1af841/src/main/java/com/squareup/javapoet/CodeWriter.java#L402, Javapoet missed a check to keep the package prefix.

Can someone file a fix for it? Thanks!

BigJIU:
As mentioned in #860 , when we are using a class with name both appearing in the inner declaration of current super class, and a individual class in current package, the later one's name would be the simplest form (e.g. `Nest`). However, it should be `com.squareup.javapoet.Nest` and the `Nest` should be the name of inner class.

Therefore, an judgement is added in the trim step in `lookupname`. With that, the name of external class in same package would not be trim any longer and will be write as its full name.

By the way, in order to test this situation, two other class is added in `test`. They don't look good, but that's the only way I know to get the bug's situation.

### DIFF
--- a/src/main/java/com/squareup/javapoet/CodeWriter.java
+++ b/src/main/java/com/squareup/javapoet/CodeWriter.java
@@ -401,8 +401,16 @@ final class CodeWriter {
 
     // If the class is in the same package, we're done.
     if (Objects.equals(packageName, className.packageName())) {
-      referencedNames.add(topLevelSimpleName);
-      return join(".", className.simpleNames());
+      String conflictClassName = packageName+"."+className.simpleName;//From same package (Nest)
+      //For nested conflict situation, the simpleName should be in QualifyList
+      //But its full name dont contains its super name
+      if(alwaysQualify.contains(className.simpleName)
+              &&Objects.equals(conflictClassName,className.canonicalName)){
+
+      }else{//If not, do regular trim
+        referencedNames.add(topLevelSimpleName);
+        return join(".", className.simpleNames());
+      }
     }
 
     // We'll have to use the fully-qualified name. Mark the type as importable for a future pass.
--- a/src/test/java/com/squareup/javapoet/FooInterface.java
+++ b/src/test/java/com/squareup/javapoet/FooInterface.java
@@ -1,0 +1,5 @@
+package com.squareup.javapoet;
+
+public interface FooInterface {
+    public class Nest{};
+}
--- a/src/test/java/com/squareup/javapoet/JavaFileTest.java
+++ b/src/test/java/com/squareup/javapoet/JavaFileTest.java
@@ -757,6 +757,50 @@ public final class JavaFileTest {
         + "}\n");
   }
 
+  @Test public void avoidClashesWithNestedClasses_SamePackageTest() {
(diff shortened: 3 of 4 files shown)


src/main/java/com/squareup/javapoet/CodeWriter.java
```diff
@@ -401,8 +401,16 @@ String lookupName(ClassName className) {
 
     // If the class is in the same package, we're done.
     if (Objects.equals(packageName, className.packageName())) {
-      referencedNames.add(topLevelSimpleName);
-      return join(".", className.simpleNames());
+      String conflictClassName = packageName+"."+className.simpleName;//From same package (Nest)
```
wanyingd1996, line 404: Thanks for the fix! but what do you need this line for? It is already checked that `packageName ` equal to `className.packageName()`, so `packageName+"."+className.simpleName` is equivalent to `className.packageName() +"."+className.simpleName`, right? Then you don't need the comparison to canonicalName below, because it is always true

BigJIU: Well, actually the `conflictClassName `is not always equal to the `className.canonicalName`, because of the existence of super class. The ignorance of trim only occur when the class is `alwaysQualify `and `canonicalName `equals, that is ,the class share a same name with a subclass but canonical name doesn't act like that. For more detail about the issue described situation, you can go to the `avoidClashesWithNestedClasses_SamePackageTest` in `JavaFileTest`

BigJIU: > Thanks for the fix! but what do you need this line for? It is already checked that `packageName ` equal to `className.packageName()`, so `packageName+"."+className.simpleName` is equivalent to `className.packageName() +"."+className.simpleName`, right? Then you don't need the comparison to canonicalName below, because it is always true

Well, actually the `conflictClassName `is not always equal to the `className.canonicalName`, because of the existence of super class. The ignorance of trim only occur when the class is `alwaysQualify `and `canonicalName `equals, that is ,the class share a same name with a subclass but canonical name doesn't act like that. For more detail about the issue described situation, you can go to the `avoidClashesWithNestedClasses_SamePackageTest` in `JavaFileTest`

PR #880, opened 2022-04-22
Add Record Class for java16

Hongstr:
I have noticed that the new version of java has provided the "record" class, so I tried to modify the original code and complete the feature of record class. 
To use the record class, I added some new method.
First, to create a record class, we can use `TypeSpec.recordBuilder(name of record)` and use `.addRecord(Type, name)` to add parameter of the record.
Second, record has a special constructor called compact constructor, we can use `MethodSpec.CompactConstructorBuilder()` to create one.
Here is a demo to create a record class:
```
TypeSpec record = TypeSpec.recordBuilder("Employee")
        .addField(int.class, "empToken", Modifier.STATIC)
        .addRecord(int.class, "id")
        .addRecord(String.class, "firstName")
        .addRecord(String.class, "lastName")
        .addMethod(MethodSpec.CompactConstructorBuilder()
                .addCode("if (id < 100) {\n" +
                        "    throw new IllegalArgumentException(" +
                        "    \"Employee Id cannot be below 100.\");\n" +
                        "}\n" +
                        "if (firstName.length() < 2) {\n" +
                        "    throw new IllegalArgumentException(\n" +
                        "    \"First name must be 2 characters or more.\");\n" +
                        "}")
                .addModifiers(Modifier.PUBLIC).build())
                .addParameter(int.class, "id")
                .addParameter(String.class, "lastName")
                .addStatement("this(id, firstName, null)").build())
```
Here is what the result looks like:
```
public record Employee(int id, String firstName, String lastName) {
  static int empToken;

  Employee(int id, String lastName) {
    this(id, firstName, null);
  }

  public Employee {
    if (id < 100) {
        throw new IllegalArgumentException(    "Employee Id cannot be below 100.");
    }
    if (firstName.length() < 2) {
        throw new IllegalArgumentException(
        "First name must be 2 characters or more.");
    }
  }
```

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -42,7 +42,7 @@ import static com.squareup.javapoet.Util.checkState;
 /** A generated constructor or method declaration. */
 public final class MethodSpec {
   static final String CONSTRUCTOR = "<init>";
-
+  static final String COMPACT_CONSTRUCTOR = "compact";
   public final String name;
   public final CodeBlock javadoc;
   public final List<AnnotationSpec> annotations;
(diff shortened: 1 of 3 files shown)


Hongstr:
#829

PR #887, opened 2022-04-29
Add ability to insert comment inside class scope

Scottmitch:
Motivation:
When generating code it maybe useful to insert
comments as anchors for post processing and further
extension by other code generation tools. However
this is not easy to achieve with the existing API.

Modifications:
- Add a `TypeSpec.addRawInitializerBlock` method which
  is treated the same as `addInitializerBlock` without
  type restrictions and additional scope block.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeSpec.java
+++ b/src/main/java/com/squareup/javapoet/TypeSpec.java
@@ -626,6 +626,11 @@ public final class TypeSpec {
       return this;
     }
 
+    public Builder addRawInitializerBlock(CodeBlock block) {
+      initializerBlock.add(block).add("\n");
+      return this;
+    }
+
     public Builder addMethods(Iterable<MethodSpec> methodSpecs) {
       checkArgument(methodSpecs != null, "methodSpecs == null");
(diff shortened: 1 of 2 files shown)


Scottmitch:
https://github.com/apple/servicetalk/pull/2160 -> to work around not being able to insert comments outside pre-defined scopes we have to do regex on the generated code. please advise if there is a more preferred way todo this, and if not this PR provides a way. There is somewhat of an escape hatch but the users of javapoet can already generate invalid code in other scenarios so I think something similar would be reasonable.

src/main/java/com/squareup/javapoet/TypeSpec.java
```diff
@@ -626,6 +626,11 @@ public Builder addInitializerBlock(CodeBlock block) {
       return this;
     }
 
+    public Builder addRawInitializerBlock(CodeBlock block) {
```
Scottmitch, line 629: happy to change the name if maintainers prefer (e.g. `addRawCodeBlock` as suggested by @bondolo)

Scottmitch:
closing in favor of https://github.com/square/javapoet/pull/967 (more constrained approach)

PR #888, opened 2022-04-30
enable Dependabot

sullis:


JakeWharton:
Can't believe this project still builds with Maven

Issue #600, opened 2018-01-16
Line endings are forced to be UNIX endings \n


thecheeselover:
There should be a way to switch between LF and CR/LF endings but the \n endings are hardcoded within the JavaFile class.

topalavlad:
The hardcoded new line endings `\n` seem to be used in ~9 files.
I think even though it's not the prettiest thing, it would be easiest for the parameter to be static for example in the `Util` class or somewhere similar. Any opinions on this?

thecheeselover:
That or if the writer could have a default object that returns a line ending, so that we could also supply another object that returns a desired line ending.

the-end-sea-of-the-world:
Hello, I would like to work on this.

JakeWharton:
Can you describe your motivation for wanting different line endings? Any modern editor will be able to handle newline-delimited lines automatically and `javac` doesn't care. 

Changing the line endings based on platforms makes testing dramatically harder as you can no longer compare golden files/strings directly to the output. Even worse, build systems like Bazel and Gradle which hash inputs will get cache misses when developers are working on different operating systems preventing distributed caching from working and forcing needless recompilations.

If we were to add this (and that's a huge _if_, as I currently see no justification for it) then it would probably have to be a configurable value on each `FileSpec`. Definitely not some static value and definitely cannot be inferred from the platform.

jbeckers:
This is probably a weak argument, but... :-)
We use javapoet to generate code that ends up in an IntelliJ project that has Windows line endings.
When viewing the generated code in IJ, it complains that the file's line endings are wrong.
Not a big deal, but annoying none the less.

Inego:
JavaFile has `writeToPath` that accepts a charset.
If charsets are supported then line endings also should be.

lyahim:
Hi all,

Please think about to implement this feature because on Windows environment line endings CRLF.
Java is a platform independent language and the application should be same. If a java source generator don't support System.lineSeparator() the applications which use that can't be platform independent.

JakeWharton:
LF-delimited files compile fine on Windows, and have for the entirety of this library's existence. Moreover, an exact, byte-for-byte output is essential for distributed build caches to work correctly with tools like Gradle. At this point, we aren't changing the library behavior.

lyahim:
You are right, generated files are correctly compiled on Windows too but Git on Windows by default try to force CRLF which cause differences. Of course it could handle with some effort.
[...]

PR #905, opened 2022-05-24
Add the ability to adjust the line separator based on the system.

the-end-sea-of-the-world:
Deal with issue [#600](https://github.com/square/javapoet/issues/600). Check the system and use the corresponding line separator.

### DIFF
--- a/src/main/java/com/squareup/javapoet/AnnotationSpec.java
+++ b/src/main/java/com/squareup/javapoet/AnnotationSpec.java
@@ -53,8 +53,8 @@ public final class AnnotationSpec {
   }
 
   void emit(CodeWriter codeWriter, boolean inline) throws IOException {
-    String whitespace = inline ? "" : "\n";
-    String memberSeparator = inline ? ", " : ",\n";
+    String whitespace = inline ? "" : JavaFile.addLineSeparator("");
+    String memberSeparator = inline ? ", " : JavaFile.addLineSeparator(",");
     if (members.isEmpty()) {
       // @Singleton
       codeWriter.emit("@$T", type);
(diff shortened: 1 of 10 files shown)


src/main/java/com/squareup/javapoet/JavaFile.java
```diff
@@ -333,4 +327,31 @@ public JavaFile build() {
...
+      LS = "\r\n";
+    }else if (osName.contains("Linux")){
+      LS = "\n";
+    }else if (osName.contains("Mac OS X")){
+      LS = "\n";
+    }else if (osName.contains("Mac OS")){
+      LS = "\r";
+    }else{
+      LS = "\n";
+    }
```
ppkarwasz, line 345: I would use `System.lineSeparator()` instead.

src/main/java/com/squareup/javapoet/AnnotationSpec.java
```diff
@@ -53,8 +53,8 @@ private AnnotationSpec(Builder builder) {
   }
 
   void emit(CodeWriter codeWriter, boolean inline) throws IOException {
-    String whitespace = inline ? "" : "\n";
-    String memberSeparator = inline ? ", " : ",\n";
+    String whitespace = inline ? "" : JavaFile.addLineSeparator("");
+    String memberSeparator = inline ? ", " : JavaFile.addLineSeparator(",");
```
ppkarwasz, line 57: What about a `JavaFile.addLineSeparator(String src, boolean inline)` helper function instead?

ppkarwasz (commented):
Looks good to me, any chance seeing this in the next release?

JakeWharton:
This would break many things. For one, you could no longer easily test against golden strings (something this library itself relies on). But even worse is that you break distributed caching of tasks in libraries like Bazel and Gradle which will compute different hashes due to line ending changes despite the contents otherwise being the same.

Issue #656, opened 2018-07-19
MethodSpec.Builder.overriding(ExecutableElement, ExecutableType)


Leland-Takamine:
The following API exists to override a method as a member of the `DeclaredType`:

```
MethodSpec.Builder.overriding(ExecutableElement, DeclaredType, Types)
```

This is inconvenient in certain cases because it requires a `Types` instance. For instance, if you code generation layer only interacts with model objects and no longer has access to the `ProcessingEnvironment` the following API would be more convenient:

```
MethodSpec.Builder.overriding(ExecutableElement, ExecutableType)
```

I understand that this opens up the API to a runtime error in the case where the `ExecutableElement` doesn't match the `ExecutableType` but it sounds like a reasonable tradeoff to expect the user to handle that correctly.

Leland-Takamine:
Thoughts on adding the following API?

```
MethodSpec.Builder.overriding(ExecutableElement, ExecutableType)
```

Hongstr:
I would like to work on it.

PR #906, opened 2022-05-26
Add Method Spec API

Hongstr:
The following API exists to override a method as a member of the DeclaredType:

```
MethodSpec.Builder.overriding(ExecutableElement, DeclaredType, Types)
```
This is inconvenient in certain cases because it requires a Types instance. For instance, if your code generation layer only interacts with model objects and no longer has access to the ProcessingEnvironment the following API would be more convenient:

```
MethodSpec.Builder.overriding(ExecutableElement, ExecutableType)
```

### DIFF
--- a/src/main/java/com/squareup/javapoet/MethodSpec.java
+++ b/src/main/java/com/squareup/javapoet/MethodSpec.java
@@ -276,6 +276,34 @@ public final class MethodSpec {
     return builder;
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


Hongstr:
https://github.com/square/javapoet/issues/656#issuecomment-406424554

liujianpc:
您好，本人已收到您的邮件！谢谢！      祝您工作顺利，生活愉快！

Issue #488, opened 2016-08-06
Floats are not synthesized correctly


nanodeath:
Floats are synthesized as doubles instead of floats.

If I have code like this:

```java
FieldSpec.builder(SpriteDefn::class.java, identifier, Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL)
  .initializer("new \$T(\$S, \$S, new float[] { \$L, \$L })", SpriteDefn::class.java, ship.name,
    ship.graphics.image,
    ship.graphics.origin[0], ship.graphics.origin[1]).build()
```

where `ship.graphics.origin` is a `float[]`, I end up with

```java
public static final SpriteDefn Gun = new SpriteDefn("Gun", "gun.png", new float[] { 0.0, 0.0 });
```

which is invalid Java -- it should be `0.0f` instead of `0.0`.

nanodeath:
As a workaround, I can specify the initializer string as `"new \$T(\$S, \$S, new float[] { \$Lf, \$Lf })"`, but that'll break when this is fixed...

swankjesse:
$L does toString on its input. This is working as intended.

nanodeath:
That seems like an implementation detail. If I say "insert a literal here" (using `$L`) and pass a float, shouldn't it emit a float literal?

JakeWharton:
I think the term "literal" is just overloaded here as `$L` currently doesn't mean Java language literal. Floats aren't anything unique here, it also doesn't wrap strings in quotes, chars in single quotes, append longs with `L`, etc.

nanodeath:
I guess. It would be more reassuring if there were tests or documentation around this behavior.

For strings, it's fine -- I just use `$S` any time I want a string. But there's no equivalent for the other types -- I just have to remember that I need to add additional content to the format string when using certain primitives.

ETCHKILI:
Although the behavior is intended, we could add a feature that check the primitive type and give the corresponding literals. I am working on it.

nanodeath:
It might be interesting if there was a placeholder that turned primitive inputs into primitive outputs, e.g. if you give it a float or string, it emits a float or string.

PR #913, opened 2022-05-29
Add Java literal support for $L literal for issue #488

ETCHKILI:
emitLiteral() is modified.
And consider it add support for float literal, the addMemberForValue() method in AnnotationSpec can delete the case for float type.
Some test added.


liujianpc:
您好，本人已收到您的邮件！谢谢！      祝您工作顺利，生活愉快！

PR #917, opened 2022-06-16
Added maven badge for quick installation

Osiris-Team:


JakeWharton:
This should not be first in the README. We can add a link to jump to the download section, though, which already contains the current version and how to add it to your build.

Issue #536, opened 2017-01-06
Add API for comments on Type level


Zubnix:
Currently it seems TypeSpec.Builder does not allow to add line and/or block comments (or arbitrary code). I would like to add a copyright header to my generated types. Currently the only way to do this, is inside the Javadoc which is not really desirable.

tbroyer:
How about [`JavaFile.Builder.addFileComment`](https://square.github.io/javapoet/1.x/javapoet/com/squareup/javapoet/JavaFile.Builder.html#addFileComment-java.lang.String-java.lang.Object...-)?

Zubnix:
oh look at that :)

snturk:
Is there a way to add comments **inside** an interface/class that created by `TypeSpec.Builder`?

Example:

```
public interface MyInterface {
    // Comment Line
}
```

PR #945, opened 2023-01-27
added JavaFile type-level commenting feature

snturk:
* Comments added to TypeSpec
* Tests updated with comment

### DIFF
--- a/src/main/java/com/squareup/javapoet/JavaFile.java
+++ b/src/main/java/com/squareup/javapoet/JavaFile.java
@@ -57,6 +57,7 @@ public final class JavaFile {
   };
 
   public final CodeBlock fileComment;
+  public final CodeBlock typeComment;
   public final String packageName;
   public final TypeSpec typeSpec;
   public final boolean skipJavaLangImports;
@@ -66,6 +67,7 @@ public final class JavaFile {
 
   private JavaFile(Builder builder) {
     this.fileComment = builder.fileComment.build();
+    this.typeComment = builder.typeComment.build();
     this.packageName = builder.packageName;
     this.typeSpec = builder.typeSpec;
     this.skipJavaLangImports = builder.skipJavaLangImports;
@@ -210,6 +212,10 @@ public final class JavaFile {
       importedTypesCount++;
     }
 
+    if (!typeComment.isEmpty()) {
+        codeWriter.emitComment(typeComment);
+    }
+
     if (importedTypesCount > 0) {
       codeWriter.emit("\n");
     }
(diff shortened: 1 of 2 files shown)


JakeWharton:
What is the motivation for this style of comments on types?

I suspect it will be surprising that you can call
```java
  .addComment(..)
  .addMethod(..)
  .addComment(..)
  .addMethod(..)
```
and the comments will be merged rather than interleaved among the methods.

snturk:
The purpose of this PR is to give javapoet the ability to add comments to TypeSpec.

Thus, comments can be added only for the method, as well as for the interface or class.

An example output would be as follows:


```
package com.squareup.tacos;
    
    class Meat {
      // This is a code block comment 
    }
```

Another option is above-class comment. I can change this branch into accept just a single comment to upper line of defined class like:

```
package com.squareup.tacos;

      // This is a code block comment    
    class Meat {

    }
```

JakeWharton:
But what purpose does it serve in comparison to Javadoc?

snturk:
It is not always desirable to provide explanatory comments as Javadoc. In this case, an extra function is offered to the javapoet user.

For example, just adding a TODO/FIXME does not require adding Javadoc.

JakeWharton:
Sure but you would put TODOs and FIXMEs inside the code performing the code generation. Not inside the code that is being generated.

snturk:
Example usage:

```
TypeSpec myType = TypeSpec.classBuilder...
.addTypeComment("TODO: This type is autogenerated and incomplete!")
...
.build()
```

Then it produces that result:

```
package com.squareup.tacos;

      // TODO: This type is autogenerated and incomplete!
    class Meat {
         ...
    }
```

I think it is ok to provide that kind of feature to javapoet. Which part of that is unsuitable?

JakeWharton:
It's not clear that there's high demand for such a feature.

You could emit such a statement in an HTML comment inside of Javadoc, for example:
```java
package com.squareup.tacos;

    /**
     * <!-- TODO: This type is autogenerated and incomplete! -->
     */
    class Meat {
         ...
    }
```

snturk:
Of course, there is no single way to add a comment above the Type. However, all methods other than directly adding single-line comments can be counted as workarounds.


It is possible to see similar requests in issues and open PRs.

#887 #536

JakeWharton:
We do support file-level comments which is what #536 wanted.

As to #887 that is about non-Javadoc comments on instance initializers (or you could generalize the feature request to being about members and types). But neither this PR nor theirs supports the feature in a way that we're comfortable shipping with.

snturk:
Thanks for the feedback 👍 

I forked and applied my changes. I think these commenting features are not suitable for the project goal yet.

PR #967, opened 2023-05-07
TypeSpec initializer comment support

Scottmitch:
Motivation:
When generating code it maybe useful to insert
comments as anchors for post processing and further extension by other code generation tools. However
this is not easy to achieve with the existing API.

Modifications:
- Add a `TypeSpec.addInitializerComment` method which is emitted only as comments above the `addInitializerBlock`.


Scottmitch:
This is a more constrained approach to https://github.com/square/javapoet/pull/887. @JakeWharton in reference to https://github.com/square/javapoet/pull/945#issuecomment-1406956640, is this approach better suited to merge? Let me know if adjustment are required.

PR #981, opened 2023-08-30
Update to support generate record from java 14

zzBBc:
Motivation:
- Solves issue #829 to support record from java 14

Modifications:
- From commit f10b510 still support and work completely in java 8, don't need to update java version if only purpose is generate record class
- I also update some lib and fix typo in unit test to support java 17, you can separate and create another version to a long term support in commit dc02cf7

Can you please review and merge the PR @JakeWharton


Duzhinsky:
I think it is the most desirable feature for javapoet, but unfortunately, the tool maintenance consists only of small fixes. PR #840 does the same, but still not merged. The future of javapoet is discussed [here](https://github.com/square/javapoet/discussions/866). You could use the [FabricMC/javapoet](https://github.com/FabricMC/javapoet) fork If you want to use records

pom.xml
```diff
@@ -20,7 +20,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <java.version>1.8</java.version>
+    <java.version>17</java.version>
```
JakeWharton, line 23: We cannot require users to use 17

pom.xml
```diff
@@ -51,6 +51,12 @@
   </organization>
 
   <dependencies>
+    <!-- https://mvnrepository.com/artifact/com.google.guava/guava -->
+    <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>32.1.2-jre</version>
+    </dependency>
```
JakeWharton, line 59: Please do not update unrelated dependencies with new features.

pom.xml
zzBBc: Yeah, Like I already comment on this PR, I suggest this commit as a new version to separate support long term for Java 11, 17,... 
- The previous commit f10b510 also support completely to generate record file in Java 8. But we cannot use record in Java 8, so I want something like "If it is created, it must be usable".

zzBBc: I will revert this commit, if my PR can be merged, we can discuss this ideal in another PR

pom.xml
zzBBc: Yeah, this update is only for Java 17 support

pom.xml
zzBBc: I had already reverted this commit, please review this PR again @JakeWharton

lenis0012:
+1 on this.

In the mean time for those who dont want to use a fork, there is another workaround.

I am simply doing a string replace after saving the fie:
```java
return MethodSpec.methodBuilder("Record__%sEvent".formatted(toTypeName(name)))
    // ....
```
This creates a `public void Record__MyEvent(String a, int b, boolean c) {}`
Which I then replace:
```java
try(
    OutputStreamWriter outputStream = new OutputStreamWriter(sourceFile.openOutputStream());
    StringWriter writer = new StringWriter()
) {
    JavaFile.builder(element.getQualifiedName().toString(), result)
        .indent("    ")
        .build()
        .writeTo(writer);
    outputStream.write(writer.toString().replace("void Record__", "record "));
}
```

Not the most elegant solution but it works.

Jonpez2:
Is this likely to land at any point?  Records are extremely nice, just like javapoet!

Randgalt:
I'm planning on publishing this as a separate repo. See here: https://github.com/Randgalt/java-composer. Unless @JakeWharton/Square objects.

Randgalt:
FYI - I've released v 1.0 of Java Composer with this change in it as well as support for sealed classes: https://github.com/Randgalt/java-composer

PR #986, opened 2023-10-05
Fix boxing and unboxing of annotated primitives

invliD:
TypeName#box currently relies on identity comparisons between `this` and one of the primitive TypeName constants. However, it is possible to have an instance of a primitive that isn't any of those constants by calling TypeName#annotated or TypeName#withoutAnnotations. The returned TypeName instance can no longer be boxed.

TypeName#unbox and TypeName#isBoxedPrimitive have a similar issue. `this` is compared using `equals` instead of identity, but annotations are part of the compared string, so a boxed primitive with annotations is not found to be any of the existing constants. Annotated boxed TypeName instances can no longer be unboxed.

This PR changes behavior of all 3 methods to do a more appropriate comparison when trying to find the (un)boxed version of `this`. After (un)boxing, the annotations from `this` are copied onto the returned value. That way, boxing and unboxing works on both unannotated and annotated TypeNames, preserving any annotations. When no annotations are present, identity is preserved.
I added unit tests to test both boxing and unboxing. Those tests fail without the code changes in TypeName.

### DIFF
--- a/src/main/java/com/squareup/javapoet/TypeName.java
+++ b/src/main/java/com/squareup/javapoet/TypeName.java
@@ -189,15 +189,15 @@ public class TypeName {
     TypeName thisWithoutAnnotations = withoutAnnotations();
     TypeName unboxed = null;
     if (thisWithoutAnnotations.equals(BOXED_VOID)) unboxed = VOID;
-    if (thisWithoutAnnotations.equals(BOXED_BOOLEAN)) unboxed = BOOLEAN;
-    if (thisWithoutAnnotations.equals(BOXED_BYTE)) unboxed = BYTE;
-    if (thisWithoutAnnotations.equals(BOXED_SHORT)) unboxed = SHORT;
-    if (thisWithoutAnnotations.equals(BOXED_INT)) unboxed = INT;
-    if (thisWithoutAnnotations.equals(BOXED_LONG)) unboxed = LONG;
-    if (thisWithoutAnnotations.equals(BOXED_CHAR)) unboxed = CHAR;
-    if (thisWithoutAnnotations.equals(BOXED_FLOAT)) unboxed = FLOAT;
-    if (thisWithoutAnnotations.equals(BOXED_DOUBLE)) unboxed = DOUBLE;
-    if (unboxed == null) throw new UnsupportedOperationException("cannot unbox " + this);
+    else if (thisWithoutAnnotations.equals(BOXED_BOOLEAN)) unboxed = BOOLEAN;
+    else if (thisWithoutAnnotations.equals(BOXED_BYTE)) unboxed = BYTE;
+    else if (thisWithoutAnnotations.equals(BOXED_SHORT)) unboxed = SHORT;
+    else if (thisWithoutAnnotations.equals(BOXED_INT)) unboxed = INT;
+    else if (thisWithoutAnnotations.equals(BOXED_LONG)) unboxed = LONG;
+    else if (thisWithoutAnnotations.equals(BOXED_CHAR)) unboxed = CHAR;
+    else if (thisWithoutAnnotations.equals(BOXED_FLOAT)) unboxed = FLOAT;
+    else if (thisWithoutAnnotations.equals(BOXED_DOUBLE)) unboxed = DOUBLE;
+    else throw new UnsupportedOperationException("cannot unbox " + this);
     return annotations.isEmpty() ? unboxed : unboxed.annotated(annotations);
   }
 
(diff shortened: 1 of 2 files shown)


invliD:
@JakeWharton Would you be able to take a look at this?

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -160,16 +164,18 @@ public boolean isBoxedPrimitive() {
...
+    if (keyword.equals(VOID.keyword)) boxed = BOXED_VOID;
+    if (keyword.equals(BOOLEAN.keyword)) boxed = BOXED_BOOLEAN;
+    if (keyword.equals(BYTE.keyword)) boxed = BOXED_BYTE;
+    if (keyword.equals(SHORT.keyword)) boxed = BOXED_SHORT;
+    if (keyword.equals(INT.keyword)) boxed = BOXED_INT;
+    if (keyword.equals(LONG.keyword)) boxed = BOXED_LONG;
+    if (keyword.equals(CHAR.keyword)) boxed = BOXED_CHAR;
+    if (keyword.equals(FLOAT.keyword)) boxed = BOXED_FLOAT;
+    if (keyword.equals(DOUBLE.keyword)) boxed = BOXED_DOUBLE;
+    if (boxed == null) throw new AssertionError("cannot box " + keyword);
```
JakeWharton, line 177: If we match we should short-circuit rather than needlessly testing conditionals we know will fail.
```suggestion
    if (keyword.equals(VOID.keyword)) boxed = BOXED_VOID;
    else if (keyword.equals(BOOLEAN.keyword)) boxed = BOXED_BOOLEAN;
    else if (keyword.equals(BYTE.keyword)) boxed = BOXED_BYTE;
    else if (keyword.equals(SHORT.keyword)) boxed = BOXED_SHORT;
    else if (keyword.equals(INT.keyword)) boxed = BOXED_INT;
    else if (keyword.equals(LONG.keyword)) boxed = BOXED_LONG;
    else if (keyword.equals(CHAR.keyword)) boxed = BOXED_CHAR;
    else if (keyword.equals(FLOAT.keyword)) boxed = BOXED_FLOAT;
    else if (keyword.equals(DOUBLE.keyword)) boxed = BOXED_DOUBLE;
    else throw new AssertionError(keyword);
```

src/main/java/com/squareup/javapoet/TypeName.java
```diff
@@ -180,16 +186,19 @@ public TypeName box() {
...
+    if (thisWithoutAnnotations.equals(BOXED_VOID)) unboxed = VOID;
+    if (thisWithoutAnnotations.equals(BOXED_BOOLEAN)) unboxed = BOOLEAN;
+    if (thisWithoutAnnotations.equals(BOXED_BYTE)) unboxed = BYTE;
+    if (thisWithoutAnnotations.equals(BOXED_SHORT)) unboxed = SHORT;
+    if (thisWithoutAnnotations.equals(BOXED_INT)) unboxed = INT;
+    if (thisWithoutAnnotations.equals(BOXED_LONG)) unboxed = LONG;
+    if (thisWithoutAnnotations.equals(BOXED_CHAR)) unboxed = CHAR;
+    if (thisWithoutAnnotations.equals(BOXED_FLOAT)) unboxed = FLOAT;
+    if (thisWithoutAnnotations.equals(BOXED_DOUBLE)) unboxed = DOUBLE;
+    if (unboxed == null) throw new UnsupportedOperationException("cannot unbox " + this);
```
JakeWharton, line 200:
```suggestion
    if (thisWithoutAnnotations.equals(BOXED_VOID)) unboxed = VOID;
    else if (thisWithoutAnnotations.equals(BOXED_BOOLEAN)) unboxed = BOOLEAN;
    else if (thisWithoutAnnotations.equals(BOXED_BYTE)) unboxed = BYTE;
    else if (thisWithoutAnnotations.equals(BOXED_SHORT)) unboxed = SHORT;
    else if (thisWithoutAnnotations.equals(BOXED_INT)) unboxed = INT;
    else if (thisWithoutAnnotations.equals(BOXED_LONG)) unboxed = LONG;
    else if (thisWithoutAnnotations.equals(BOXED_CHAR)) unboxed = CHAR;
    else if (thisWithoutAnnotations.equals(BOXED_FLOAT)) unboxed = FLOAT;
    else if (thisWithoutAnnotations.equals(BOXED_DOUBLE)) unboxed = DOUBLE;
    else throw new UnsupportedOperationException("cannot unbox " + this);
```

JakeWharton (approved):

PR #1004, opened 2024-04-02
Refactoring and Removing Smells

ywalia01:


JakeWharton:
Please see https://github.com/square/retrofit/pull/3905#issuecomment-1566159367

